PR #854, opened 2025-06-20
[PECOBLR-530] Fix leak in preparedStatement + fix interpolation in metadata

samikshya-db:
## Description

This PR does 2 things : 
1. Closes the describe query statement in preparedStatement (fix the leak)
2. Fixes the errors when sql does not have parameters defined (Otherwise a parsing error is thrown from the backend). For example `Describe select ?` fails whereas `Describe select '?'` passes

## Testing
- The  leakage in our driver is fixed after this code change. Verified by using a serverless warehouse I created on dogfood `samikshya-test-small-warehouse`. The warehouse closes within 5 minutes (which is the timeout I have set too)

Code used :

```
Connection connection = DriverManager.getConnection(JDBC_URL, USERNAME, PASSWORD);
PreparedStatement preparedStatement = connection.prepareStatement("select 1");
ResultSetMetaData rs= preparedStatement.getMetaData();
```

Screenshot 2025-06-03 at 9 19 32 PM

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 - Added support for DoD (.mil) domains
-- Support to fetch metadata in PreparedStatement for SELECT queries before executing the query.
+- Enables fetching of metadata for SELECT queries using PreparedStatement prior to setting parameters or executing the query.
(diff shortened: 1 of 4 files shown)


samikshya-db:
Merging this as we had done the review as part of https://github.com/databricks/databricks-jdbc/pull/847 (This PR got closed as a result of OSS processes.)

PR #856, opened 2025-06-23
Decouple decimal conversion from arrow metadata

vikrantpuppala:
## Description

Decouple decimal conversion from arrow metadata

## Testing

Unit tests


src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
```diff
@@ -262,25 +264,17 @@ private static byte[] convertToByteArray(Object object) {
...
-                  .trim());
-    } catch (Exception e) {
-      LOGGER.error(
-          e, "Failed to parse scale from arrow metadata: {}. Defaulting to scale 0", arrowMetadata);
-      scale = 0;
-    }
-    if (object instanceof Number) {
-      return new BigDecimal(object.toString()).setScale(scale, RoundingMode.HALF_UP);
+    if (object instanceof Text || object instanceof Number) {
+      BigDecimal bigDecimal = new BigDecimal(object.toString());
```
gopalldb, line 270: I had a comment on this on original PR. Instead of doing toString, for many of number types, we can parse the value from respective number types

PR #864, opened 2025-06-25
fix for registration conflicts with BouncyCastleProvider

sreekanth-db:
## Description

The core problem is from the static registration of the BouncyCastle Security Provider in Java's Security API. When multiple BouncyCastle library versions are on the classpath, each trying to register its "BC" provider, conflicts occur. This can arise from different versions being introduced by drivers, the application, or its dependencies. The first registered provider wins, leading to unpredictable behavior if an older or incompatible version is registered before the expected one. This causes difficult-to-diagnose runtime problems for customers, highlighting the need to isolate or control BouncyCastle registration.

Solution: We will no longer globally register BouncyCastleProvider in Java's security API. Instead, our driver will create a dedicated local BouncyCastleProvider instance for it usage.

## Testing
Tested jwt flow in local

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,7 +9,7 @@
 - 
 
 ### Fixed
-- Fixed bouncy castle registration issue by using local instance instead of global registration with java security API.
+- Fixed Bouncy Castle registration conflicts by using local provider instance instead of global security registration.
 
 ---
 *Note: When making changes, please add your change under the appropriate section with a brief description.* 


NEXT_CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 - 
 
 ### Fixed
--
+- Fixed bouncy castle registration issue by using local instance instead of global registration with java security API.
```
samikshya-db, line 12:
```suggestion
- Fixed Bouncy Castle registration conflicts by using local provider instance instead of global security registration.
```

sreekanth-db:
There is some issue with the coverage comparison workflow, will fix it in a subsequent pull request. Coverage percentage from local jacoco report for this branch is 86.31%

PR #878, opened 2025-07-02
Remove usage of pull_request_target everywhere

samikshya-db:
## Description
- Given that GITHUB TOKEN  is also unsafe to use in forked PRs (even though it is rotated automatically), I will remove these too
- Modified the tests accordingly. 

 NO_CHANGELOG=true


jprakash-db (approved):
LGTM

PR #880, opened 2025-07-06
Add time based flush + add force telemetry flag

samikshya-db:
## Description
- We should be sending out telemetry in 3 cases : 
    -  When batch size reaches a threshold (Already implemented ✅ )
    -  When connection/statement is closed  (Already implemented ✅ )
    - When time since last telemetry export reaches threshold  (In this PR 🚧 )
- Add a force telemetry flag : this will turn on telemetry even when server flag say otherwise. This will be useful when : 
  - Internal use-cases need to be laced with telemetry (eg: benchmarking)
  - Force telemetry for customers facing issues : for us to be able to debug better. 

## Testing
- Added unit test to ensure they are working

NO_CHANGELOG=TRUE

### DIFF
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
@@ -8,6 +8,8 @@ import com.databricks.sdk.core.DatabricksConfig;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 public class TelemetryClient implements ITelemetryClient {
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
```diff
@@ -38,6 +44,16 @@ public TelemetryClient(
...
     this.executorService = executorService;
+    this.scheduledExecutorService =
+        java.util.concurrent.Executors.newSingleThreadScheduledExecutor();
+    schedulePeriodicFlush();
+  }
+
+  private void schedulePeriodicFlush() {
+    // Ensure minimum 1 second interval to avoid over-calling flush
+    int intervalMillis = Math.max(1000, context.getTelemetryFlushIntervalInMilliseconds());
+    scheduledExecutorService.scheduleAtFixedRate(
```
gopalldb, line 55: let's reset the last flushedTime on flush, so that timer also gets reset

samikshya-db: Makes sense, addressed ✅

gopalldb (approved):

PR #884, opened 2025-07-07
Remove secrets from the FakeServiceTests

jayantsing-db:
## Description

FakeService tests running in replay mode in CI do not need secrets. When generating stubs locally, replace with actual PAT or secrets.

## Testing
CI


src/test/java/com/databricks/jdbc/integration/fakeservice/tests/UCVolumeIntegrationTests.java
```diff
@@ -354,8 +353,8 @@ private Connection getConnection() throws SQLException {
     String jdbcUrl = String.format(jdbcUrlTemplate, getFakeServiceHost(), HTTP_PATH);
 
     Properties connProps = new Properties();
-    connProps.put(DatabricksJdbcUrlParams.USER.getParamName(), getDatabricksUser());
-    connProps.put(DatabricksJdbcUrlParams.PASSWORD.getParamName(), getDatabricksToken());
```
samikshya-db, line 358: Thanks for this PR. One comment though : can we have `token` hardcoded in `getDatabricksToken ()` so that while generating stubs, we don't have to replace it at multiple places?

PR #887, opened 2025-07-09
Removing EPL and GPL License library

jprakash-db:
## Description
Removing the libraries that have licensing issues when included in the OSS project

- Removed jakarta servlet from the final jar as it is required only by wiremock

`
NO_CHANGELOG=true
`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -484,10 +484,6 @@
                   <shadedPattern>com.databricks.internal.io</shadedPattern>
                 </relocation>
                 <relocation>
-                  <pattern>jakarta.servlet</pattern>
-                  <shadedPattern>com.databricks.internal.jakarta.servlet</shadedPattern>
-                </relocation>
-                <relocation>
                   <pattern>net.jpountz</pattern>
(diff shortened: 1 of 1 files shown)


gopalldb (approved):

jayantsing-db:
iirc in the ticket there were a lot more annotations/libraries with restricted licenses? are they already excluded from the jar?

jayantsing-db:
understood. it's weird that it is needed by wiremock (i don't think so). the library originates from thrift lib.
```
[INFO] +- org.apache.thrift:libthrift:jar:0.19.0:compile
[INFO] |  +- (org.slf4j:slf4j-api:jar:1.7.36:runtime - omitted for conflict with 2.0.13)
[INFO] |  +- (org.apache.httpcomponents.client5:httpclient5:jar:5.2.1:runtime - omitted for conflict with 5.3.1)
[INFO] |  +- (org.apache.httpcomponents.core5:httpcore5:jar:5.2:runtime - omitted for conflict with 5.3.1)
[INFO] |  +- jakarta.servlet:jakarta.servlet-api:jar:5.0.0:runtime
```

PR #904, opened 2025-07-18
Telemetry : fit and finish :)

samikshya-db:
## Description
- Catch the result format in telemetry
- Catch the chunk id/latency accurately
- Change the telemetry log level to trace, we shouldn't be noisy on telemetry. Also, remove un-necessary logging from telemetry
- Add more logs in execution format
- Unify the statement ID in telemetry (we should use the `.toSQLExecStatementId` at all places). This is because we need to join with the http logs table

## Testing
- Tested manually across CUJs : [Link](https://docs.google.com/document/d/1oAu1pTCM9VFrPDCt0I_aWsSIoNGOapmGc_2Zob6QUiA/edit?tab=t.0#heading=h.n1p26dmnjyu)
- Also added unit tests for coverage

- Currently keeping telemetry OFF by default (will make it ON after benchmarking issues are addressed.)
[Link to benchmarking ](https://docs.google.com/document/d/1oAu1pTCM9VFrPDCt0I_aWsSIoNGOapmGc_2Zob6QUiA/edit?tab=t.0#heading=h.9ht5j0colwim)


src/main/java/com/databricks/client/jdbc/Driver.java
```diff
@@ -65,8 +65,6 @@ public Connection connect(String url, Properties info) throws DatabricksSQLExcep
       connection.open();
       isConnectionOpen = true;
       DriverUtil.resolveMetadataClient(connection);
-      exportInitialTelemetryLog(
```
gopalldb, line 68: do we intend this to add this later once we enable safe flag to 100%?

src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
```diff
@@ -94,12 +94,6 @@ public void close() {
     scheduledExecutorService.shutdown();
   }
 
-  @Override
-  public void closeStatement(String statementId) {
-    TelemetryCollector.getInstance().exportTelemetryDetailsAndClear(statementId);
```
gopalldb, line 99: why is this removed?

src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
```diff
@@ -70,26 +71,6 @@ public static boolean isTelemetryAllowedForConnection(IDatabricksConnectionConte
             .isFeatureEnabled(TELEMETRY_FEATURE_FLAG_NAME);
   }
 
-  public static void exportInitialTelemetryLog(IDatabricksConnectionContext connectionContext) {
```
gopalldb, line 73: we still need to export them after a connection is created, no?

src/main/java/com/databricks/jdbc/api/impl/arrow/ArrowStreamResult.java
```diff
@@ -59,10 +59,14 @@ public ArrowStreamResult(
     this.session = session;
     // Check if the result data contains the arrow data inline
     boolean isInlineArrow = resultData.getAttachment() != null;
-
     if (isInlineArrow) {
+      LOGGER.debug(
+          "Creating ArrowStreamResult with inline attachment {}",
```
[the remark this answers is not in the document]
samikshya-db: Ok done ✅

src/main/java/com/databricks/client/jdbc/Driver.java
samikshya-db: Yes that is correct, we can do this once we roll out completely and start using the `un-auth` endpoint

gopalldb: Let's add a TODO then

gopalldb (approved):

PR #905, opened 2025-07-21
Add DOWNLOAD_SUCCEEDED to transitions of DOWNLOAD_RETRY

jayantsing-db:
## Description

During retry of chunk download, status can go
- from DOWNLOAD_RETRY to DOWNLOAD_SUCCEEDED
- from DOWNLOAD_RETRY to DOWNLOAD_FAILED

## Testing

Added a test to verify the state updates/transitions. These were flagged in the warning logs of state machine.

### DIFF
--- a/src/test/java/com/databricks/jdbc/api/impl/arrow/ChunkDownloadTaskTest.java
+++ b/src/test/java/com/databricks/jdbc/api/impl/arrow/ChunkDownloadTaskTest.java
@@ -188,8 +188,8 @@ public class ChunkDownloadTaskTest {
     assertEquals(6, statusHistory.size());
     assertEquals(ChunkStatus.DOWNLOAD_FAILED, statusHistory.get(0));
     assertEquals(ChunkStatus.DOWNLOAD_RETRY, statusHistory.get(1)); // First retry
-    assertEquals(ChunkStatus.DOWNLOAD_FAILED, statusHistory.get(0));
-    assertEquals(ChunkStatus.DOWNLOAD_RETRY, statusHistory.get(1)); // Second retry
+    assertEquals(ChunkStatus.DOWNLOAD_FAILED, statusHistory.get(2));
+    assertEquals(ChunkStatus.DOWNLOAD_RETRY, statusHistory.get(3)); // Second retry
     assertEquals(ChunkStatus.DOWNLOAD_SUCCEEDED, statusHistory.get(4)); // Download success
 
     // The chunk should eventually reach PROCESSING_SUCCEEDED after parsing the Arrow data


src/main/java/com/databricks/jdbc/api/impl/arrow/ChunkStatus.java
```diff
@@ -87,7 +87,8 @@ public enum ChunkStatus {
     VALID_TRANSITIONS.put(PROCESSING_FAILED, Set.of(CHUNK_RELEASED));
     VALID_TRANSITIONS.put(CANCELLED, Set.of(CHUNK_RELEASED));
     VALID_TRANSITIONS.put(CHUNK_RELEASED, Collections.emptySet());
-    VALID_TRANSITIONS.put(DOWNLOAD_RETRY, Set.of(URL_FETCHED, CHUNK_RELEASED));
+    VALID_TRANSITIONS.put(
```
samikshya-db, line 90: When chunk is released, it is possible that it is being downloaded. Do we also want to add : 

` VALID_TRANSITIONS.put(CHUNK_RELEASED, Set.of(DOWNLOAD_FAILED, DOWNLOAD_SUCCEEDED));` ?  Or, we could also skip the hard check if chunk is released already.

samikshya-db: The CUJ I am talking about : I read part of resultSet and close the connection. 

This should not be throwing me a train of warnings for all the chunks which download is scheduled. 

`com.databricks.jdbc.api.impl.arrow.AbstractArrowResultChunk#setStatus - Failed to transition to state [DOWNLOAD_SUCCEEDED] from state [CHUNK_RELEASED] for chunk [13] and statement [01f066a3-2486-13f0-832e-a48771351e0b]. Stack trace: com.databricks.jdbc.exception.DatabricksParsingException: Invalid state transition for chunk [13] and statement [01f066a3-2486-13f0-832e-a48771351e0b]: CHUNK_RELEASED -> DOWNLOAD_SUCCEEDED. Valid transitions from CHUNK_RELEASED are: []`

jayantsing-db: these warnings are indeed expected as i was explaining in the prev comment. the main thread closed result set which released all chunks and rendered them useless. However, there are daemon download threads that tried to initialise/download data even after the chunk was released but before final interrupt. The semantics of chunk download considers release as the terminal step.

a better question then to ask that stem from these warnings: why are threads not immediately interrupting? that's because we use executor's `shutdownNow` which is best-effort shutdown:
```
here are no guarantees beyond best-effort attempts to stop
     * processing actively executing tasks.  For example, typical
     * implementations will cancel via {@link Thread#interrupt}, so any
     * task that fails to respond to interrupts may never terminate.
```
we should probably think more about fixing above than masking the above warnings. I hope this clarifies.

gopalldb (approved):

PR #907, opened 2025-07-22
[wip] parameter counting

madhav-db:


gopalldb (commented):
Good coverage of tests!

gopalldb (approved):
Let's also run this through comparator as discussed offline

PR #909, opened 2025-07-24
Add UID validation to always be token

vikrantpuppala:
## Description

Add UID validation to always be token

## Testing

Unit + manual tests


src/main/java/com/databricks/jdbc/common/DatabricksJdbcUrlParams.java
```diff
@@ -8,7 +8,7 @@ public enum DatabricksJdbcUrlParams {
   LOG_PATH("logpath", "Path to the log file"),
   LOG_FILE_SIZE("LogFileSize", "Maximum size of the log file", "10"), // 10 MB
   LOG_FILE_COUNT("LogFileCount", "Number of log files to retain", "10"),
-  USER("user", "Username for authentication"),
+  UID("uid", "UID for authentication"),
```
jprakash-db, line 11: Won't this break for users using `user` currently ?

jprakash-db (approved):
LGTM. Thanks for making the changes

PR #912, opened 2025-07-25
Fixed unimplemented methods in DatabricksConnection for methods which has default input by users

msrathore-db:
## Description
Added a check in the parameterized createStatement function to check if the parameters match the default value before throwing the error.

## Testing

- Tested Locally

Related tickets:
[PECOBLR-665](https://databricks.atlassian.net/browse/PECOBLR-665)

NO_CHANGELOG=true

[PECOBLR-665]: https://databricks.atlassian.net/browse/PECOBLR-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
@@ -324,9 +324,13 @@ public class DatabricksConnection implements IDatabricksConnection, IDatabricksC
   @Override
   public Statement createStatement(
       int resultSetType, int resultSetConcurrency, int resultSetHoldability) throws SQLException {
-
-    throw new DatabricksSQLFeatureNotImplementedException(
-        "Not implemented in DatabricksConnection - createStatement(int resultSetType, int resultSetConcurrency, int resultSetHoldability)");
+    if (resultSetType != ResultSet.TYPE_FORWARD_ONLY
+        || resultSetConcurrency != ResultSet.CONCUR_READ_ONLY
+        || resultSetHoldability != ResultSet.CLOSE_CURSORS_AT_COMMIT) {
+      throw new DatabricksSQLFeatureNotImplementedException(
+          "Not implemented in DatabricksConnection - createStatement(int resultSetType, int resultSetConcurrency, int resultSetHoldability)");
+    }
+    return createStatement();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
```diff
@@ -324,9 +324,13 @@ public void releaseSavepoint(Savepoint savepoint) throws SQLException {
   @Override
   public Statement createStatement(
       int resultSetType, int resultSetConcurrency, int resultSetHoldability) throws SQLException {
-
-    throw new DatabricksSQLFeatureNotImplementedException(
-        "Not implemented in DatabricksConnection - createStatement(int resultSetType, int resultSetConcurrency, int resultSetHoldability)");
+    if (resultSetType != ResultSet.TYPE_FORWARD_ONLY
+        || resultSetConcurrency != ResultSet.CONCUR_READ_ONLY
+        || resultSetHoldability != ResultSet.CLOSE_CURSORS_AT_COMMIT) {
```
shivam2680, line 329: add a test case as well in `DatabricksConnectionTest`. https://github.com/databricks/databricks-jdbc/blob/main/src/test/java/com/databricks/jdbc/api/impl/DatabricksConnectionTest.java#L356

shivam2680 (approved):

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
```diff
@@ -324,9 +324,13 @@ public void releaseSavepoint(Savepoint savepoint) throws SQLException {
...
+      throw new DatabricksSQLFeatureNotImplementedException(
```
jprakash-db, line 330: Update the error message accordingly., something like `Databricks OSS JDBC only supports resultSetType as ResultSet.TYPE_FORWARD_ONLY, resultSetConcurrency as ResultSet.CONCUR_READ_ONLY and resultSetHoldability as ResultSet.CLOSE_CURSORS_AT_COMMIT`. Check other error messages for reference

jprakash-db (approved):
LGTM. Thanks for making the changes

gopalldb (approved):

madhav-db (commented):
We can add a changelog entry for this so that users whose workflow this affects are aware of this change

PR #913, opened 2025-07-28
added logs for cloud fetch download speed

shivam2680:
## Description

Add logs for cloud fetch download speed
relevant ticket: https://databricks.atlassian.net/browse/XTA-11037

## Testing

verified logs are printed when log level is set to INFO


src/main/java/com/databricks/jdbc/api/impl/arrow/ChunkDownloadTask.java
```diff
@@ -60,7 +60,10 @@ public Void call() throws DatabricksSQLException, ExecutionException, Interrupte
             chunk.setChunkLink(link);
           }
 
-          chunk.downloadData(httpClient, chunkDownloader.getCompressionCodec());
+          chunk.downloadData(
+              httpClient,
+              chunkDownloader.getCompressionCodec(),
+              connectionContext != null ? connectionContext.getCloudFetchSpeedThreshold() : 0.1);
```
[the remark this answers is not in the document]
shivam2680: Just past trauma.

src/main/java/com/databricks/jdbc/api/impl/arrow/ArrowResultChunk.java
```diff
@@ -133,6 +138,26 @@ private void addHeaders(HttpGet getRequest, Map<String, String> headers) {
     }
   }
 
+  private void logDownloadMetrics(
+      long downloadTimeMs, long contentLength, String url, double speedThreshold) {
+    if (downloadTimeMs > 0 && contentLength > 0) {
+      double speedMBps = (contentLength / 1024.0 / 1024.0) / (downloadTimeMs / 1000.0);
+      String baseUrl = url.split("\\?")[0];
+
+      LOGGER.info(
```
jayantsing-db, line 147: are you sure this qualifies as an info log? more like a debug/warn log. And then we have telemetry releasing soon.

jayantsing-db: based on the ticket https://databricks.atlassian.net/browse/XTA-11037 i think info/warn is flexible as long as there is some way to see the speed. info would just pollute the logs with this not-so-relevant info to devs/apps unless debugging. hence a warn/debug

PR #926, opened 2025-08-05
[PECOBLR-741] Retry fix for invalid retryInterval

nishkarsh-db:
## Description
Whenever a request fails and the server does not gives a retryAfter header, we set retryInterval = -1. In this case, now we just return false in retryRequest instead of throwing an exception.

## Testing
Added a test for checking the same.

Incase the server does not pass retryAfter it automatically gets initialised to -1.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,6 @@
 ### Updated
 
 ### Fixed
-- If sever doesn't send retryAfter header, then the request won't be retried. It wouldn't throw an exception.
+- Fixed retry behaviour to not throw an exception when there is no retry-after header for 503 and 429 status codes.
 ---
 *Note: When making changes, please add your change under the appropriate section with a brief description.* 


src/test/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpRetryHandlerTest.java
```diff
@@ -243,6 +243,26 @@ void testIsRequestMethodRetryable() {
         isRequestMethodRetryable("PATCH"), "PATCH requests should not be allowed for retry");
   }
 
+  @Test
+  void testRetryRequestWithInvalidRetryInterval() {
+    // Set up context with -1 retry interval (invalid)
+    httpContext.setAttribute(RETRY_INTERVAL_KEY, -1);
```
shivam2680, line 249: add a comment mentioning this case arises if server doesn't send RETRY_AFTER header

shivam2680 (approved):
lgtm, with very minor comment. 
In future, please add ticket reference PR title

samikshya-db (dismissed):

vikrantpuppala (approved):

PR #929, opened 2025-08-07
parameterised flush to check batch size for force flush calls

nikhilsuri-db:
## Description
Current implementation of telemetry is not 100% thread safe. For a complex thread switch scenario it can flush the batch without having full size in it. 

Detail - 
Thread A adds to the batch making size N - 1 ( assuming eventsBatchSize === N )
[Context Switch] Thread B adds to the batch making size N
Both Thread A & Thread B verifies that they need to flush and try to flush the events
Thread B flushes the batch
[Context Switch] Thread C adds new event to the Batch
[Context Switch] Thread A enters the flush block and checks eventBatch is not empty and ends up flushing just 1 event. 

## Testing
Existing Test Unit. 

Fix: For schedule & close connection call we still flush everything from the batch if it is not empty but for non force calls we only flush if the batch has breached the threshold. 

NO_CHANGELOG=true

### DIFF
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
@@ -95,9 +95,8 @@ public class TelemetryClient implements ITelemetryClient {
   }
 
   /**
-   *
-   * @param forceFlush - Flushes the eventsBatch for all size variations if forceFlush,
-   *                   otherwise only flushes if  eventsBatch size has breached
+   * @param forceFlush - Flushes the eventsBatch for all size variations if forceFlush, otherwise
+   *     only flushes if eventsBatch size has breached
    */
   private void flush(boolean forceFlush) {
     synchronized (this) {
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
@@ -68,7 +68,7 @@ public class TelemetryClient implements ITelemetryClient {
   private void periodicFlush() {
     long now = System.currentTimeMillis();
     if (now - lastFlushedTime >= flushIntervalMillis) {
-      flush();
+      flush(true);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
```diff
@@ -78,25 +78,25 @@ public void exportEvent(TelemetryFrontendLog event) {
...
       flushTask.cancel(false);
     }
     scheduledExecutorService.shutdown();
   }
 
-  private void flush() {
+  private void flush(boolean forceFlush) {
     synchronized (this) {
-      if (!eventsBatch.isEmpty()) {
+      if (!forceFlush ? isBatchFull() : !eventsBatch.isEmpty()) {
```
nikhilsuri-db, line 99: Flushes for all batch size variations if it is a force flush otherwise only flushes when eventBatch threshold is breached

src/test/java/com/databricks/jdbc/telemetry/TelemetryClientTest.java
```diff
@@ -227,11 +230,12 @@ public void testTimerResetOnBatchSizeFlush() throws Exception {
       // Verify it's still in the batch (shouldn't have been flushed yet since timer was reset)
       assertEquals(1, client.getCurrentSize());
 
-      // Wait another 2 seconds (still less than full interval from last flush)
-      Thread.sleep(2000);
+      // Wait another 8 seconds (still less than full interval from last flush)
+      //NOTE: Ideally flush should happen just after 3 seconds but considering the thread switch timings adding a buffer of 5 more seconds
```
nikhilsuri-db, line 234: Not related to this JIRA ticket.

jayantsing-db:
>We also need to enforce disabling the branch creation on OSS JDBC (cc : @shivam2680 who is working on this enforcement)

Restricted branch pushes in admin settings. Can be overridden for a specific use case by an admin based on discussions in PR.

PR #933, opened 2025-08-11
parameterised flush to check batch size for force flush calls

nikhilsuri-db:
## Description
Current implementation of telemetry is not 100% thread safe. For a complex thread switch scenario it can flush the batch without having full size in it.

Detail -
Thread A adds to the batch making size N - 1 ( assuming eventsBatchSize === N )
[Context Switch] Thread B adds to the batch making size N
Both Thread A & Thread B verifies that they need to flush and try to flush the events
Thread B flushes the batch
[Context Switch] Thread C adds new event to the Batch
[Context Switch] Thread A enters the flush block and checks eventBatch is not empty and ends up flushing just 1 event.

## Testing
Existing Test Unit.

NA
NO_CHANGELOG=true

### DIFF
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
@@ -68,7 +68,7 @@ public class TelemetryClient implements ITelemetryClient {
   private void periodicFlush() {
     long now = System.currentTimeMillis();
     if (now - lastFlushedTime >= flushIntervalMillis) {
-      flush();
+      flush(true);
     }
   }
 
@@ -78,8 +78,8 @@ public class TelemetryClient implements ITelemetryClient {
       eventsBatch.add(event);
     }
 
-    if (eventsBatch.size() == eventsBatchSize) {
-      flush();
+    if (isBatchFull()) {
+      flush(false);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/databricks/jdbc/telemetry/TelemetryClient.java
```diff
@@ -78,25 +78,29 @@ public void exportEvent(TelemetryFrontendLog event) {
...
     if (flushTask != null) {
       flushTask.cancel(false);
     }
     scheduledExecutorService.shutdown();
   }
 
-  private void flush() {
+  /**
+   * @param forceFlush - Flushes the eventsBatch for all size variations if forceFlush, otherwise
+   *     only flushes if eventsBatch size has breached
```
nikhilsuri-db, line 99: Flushes for all batch size variations if it is a force flush otherwise only flushes when eventBatch threshold is breached

src/test/java/com/databricks/jdbc/telemetry/TelemetryClientTest.java
```diff
@@ -227,11 +230,13 @@ public void testTimerResetOnBatchSizeFlush() throws Exception {
       // Verify it's still in the batch (shouldn't have been flushed yet since timer was reset)
       assertEquals(1, client.getCurrentSize());
 
-      // Wait another 2 seconds (still less than full interval from last flush)
-      Thread.sleep(2000);
+      // Wait another 8 seconds (still less than full interval from last flush)
+      // NOTE: Ideally flush should happen just after 3 seconds but considering the thread switch
```
nikhilsuri-db, line 234: Not related to this JIRA ticket.

nikhilsuri-db:
Duplicate from : https://github.com/databricks/databricks-jdbc/pull/929

src/test/java/com/databricks/jdbc/telemetry/TelemetryClientTest.java
samikshya-db: Are you going to implement the `delayedExecutor /alternate` you mentioned in https://github.com/databricks/databricks-jdbc/pull/929#discussion_r2260400723

samikshya-db: interesting! thanks for trying it out!

samikshya-db (approved):
LGTM (given that we reduce the 8 second sleep)

gopalldb (approved):

PR #940, opened 2025-08-14
Fix passing of encoded userAgent string in BITool+ Add system configuration in every telemetry request

samikshya-db:
## Description
- The intention of appName in telemetry is to understand BI tools using our driver. And we do not want that to break our UserAgent. This causes DBeaver to hang as it has a app name as `DBeaver%2F25.1.4.202508031529` (and encoded strings are not allowed in our userAgent check). This PR fixes that.
- To create dashboards, we need system configuration in every telemetry row. This PR also adds that.

## Testing
- Added unit tests
- Also tested manually using DBeaver Client

- The userAgent is sent for all customers, and it may not be appropriate for us to send info like processName and appName as part of this. Moreover, UserAgent is supposed to be fixed for a JVM. The `unknown` string issue on thrift table will be addressed by telemetry now. (i.e., we will do a join between two tables to identify unknowns)

- BI tool Error that is fixed in this PR : 

Screenshot 2025-08-14 at 4 42 33 PM

### DIFF
--- a/src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
+++ b/src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
@@ -1,11 +1,14 @@
 package com.databricks.jdbc.common.util;
 
 import com.databricks.jdbc.api.internal.IDatabricksConnectionContext;
+import com.databricks.jdbc.log.JdbcLogger;
+import com.databricks.jdbc.log.JdbcLoggerFactory;
 import com.databricks.sdk.core.UserAgent;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 
 public class UserAgentManager {
+  private static final JdbcLogger LOGGER = JdbcLoggerFactory.getLogger(UserAgentManager.class);
   private static final String SDK_USER_AGENT = "databricks-sdk-java";
   private static final String JDBC_HTTP_USER_AGENT = "databricks-jdbc-http";
   private static final String DEFAULT_USER_AGENT = "DatabricksJDBCDriverOSS";
@@ -23,7 +26,10 @@ public class UserAgentManager {
     // Set the base product and client info
     UserAgent.withProduct(DEFAULT_USER_AGENT, DriverUtil.getDriverVersion());
     UserAgent.withOtherInfo(CLIENT_USER_AGENT_PREFIX, connectionContext.getClientUserAgent());
-    if (connectionContext.getCustomerUserAgent() != null) {
+    if (connectionContext.getCustomerUserAgent() == null) {
+      return;
+    }
+    try {
       String decodedUA =
           URLDecoder.decode(
               connectionContext.getCustomerUserAgent(),
@@ -32,6 +38,11 @@ public class UserAgentManager {
       String customerName = (i < 0) ? decodedUA : decodedUA.substring(0, i);
       String customerVersion = (i < 0) ? VERSION_FILLER : decodedUA.substring(i + 1);
       UserAgent.withOtherInfo(customerName, UserAgent.sanitize(customerVersion));
+    } catch (Exception e) {
+      LOGGER.debug(
+          "Failed to set user agent for customer userAgent entry {}, Error {}",
+          connectionContext.getCustomerUserAgent(),
+          e);
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
```diff
@@ -25,61 +26,23 @@ public static void setUserAgent(IDatabricksConnectionContext connectionContext)
...
-    if (!isNullOrEmpty(appName)) {
-      // Update telemetry
-      TelemetryHelper.updateClientAppName(appName);
-
-      // Update HTTP user agent
-      int i = appName.indexOf('/');
-      String customerName = (i < 0) ? appName : appName.substring(0, i);
-      String customerVersion = (i < 0) ? VERSION_FILLER : appName.substring(i + 1);
+    try {
+      String decodedUA =
```
gopalldb, line 33: Can you give an example of encoded user agent string?

src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
```diff
@@ -308,4 +304,50 @@ public static OperationType mapMethodToOperationType(String methodName) {
...
+      return appName;
+    }
+
+    // Then check client info property
+    if (!isNullOrEmpty(clientInfoAppName)) {
+      return clientInfoAppName;
+    }
+
+    // Finally check system property
+    return System.getProperty(APP_NAME_SYSTEM_PROPERTY);
```
gopalldb, line 351: this will return null or empty in case not set?

src/test/java/com/databricks/jdbc/common/util/UserAgentManagerTest.java
```diff
@@ -42,88 +37,34 @@ public void tearDown() {
...
-    telemetryHelperMock.verify(() -> TelemetryHelper.updateClientAppName("MyApp/1.2.3"));
+  @Test
+  public void testIncorrectUserAgentDoesNotThrowException() {
+    // Test that user agent is updated
+    when(connectionContext.getCustomerUserAgent()).thenReturn("DBeaver    25.1.4.202508031529");
+    when(connectionContext.getClientUserAgent()).thenReturn(USER_AGENT_SEA_CLIENT);
+    UserAgentManager.setUserAgent(connectionContext);
     String userAgent = getUserAgentString();
-    assertTrue(userAgent.contains("MyApp/1.2.3"));
+    assertFalse(userAgent.contains("DBeaver/25.1.4.202508031529"));
```
gopalldb, line 67: assert that doesn't contain DBeaver

src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
samikshya-db: Sure, I have already added a UT for it. Eg: `DBeaver%2F25.1.4.202508031529`

nikhilsuri-db (approved):
LGTM

src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
```diff
@@ -25,61 +26,23 @@ public static void setUserAgent(IDatabricksConnectionContext connectionContext)
...
-      IDatabricksConnectionContext connectionContext, String clientInfoAppName) {
-    String appName = determineApplicationName(connectionContext, clientInfoAppName);
-    if (!isNullOrEmpty(appName)) {
-      // Update telemetry
-      TelemetryHelper.updateClientAppName(appName);
-
-      // Update HTTP user agent
-      int i = appName.indexOf('/');
-      String customerName = (i < 0) ? appName : appName.substring(0, i);
-      String customerVersion = (i < 0) ? VERSION_FILLER : appName.substring(i + 1);
```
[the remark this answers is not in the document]
shivam2680: But then how will the biTool field get populated in thrift statements table? I thought the point of this task was to ensure that BiTool column has less unknown entries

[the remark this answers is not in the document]
samikshya-db: > But then how will the biTool field get populated in thrift statements table? I thought the point of this task was to ensure that BiTool column has less unknown entries

I should have mentioned the context on this PR too, my bad. I will add to the description too 

Based on discussion with @deeksha-db  and @gopalldb : 
The userAgent is sent for all customers, and it may not be appropriate for us to send info like processName and appName as part of this. Moreover, UserAgent is supposed to be fixed for a JVM. The `unknown` issue will be addressed by telemetry now.

samikshya-db:
Added DBeaver screenshot on the PR as well.

PR #945, opened 2025-08-18
Skip file path validation for delete operaation from DatabricksUCVolumeClient

nikhilsuri-db:
## Description
Issue is that in DatabricksUCVolumeClient#deleteObject, we are not setting the allowInputStreamForVolumeOperation as true which we we are  setting in other get/put operations for input stream.
This flag is being send as true even in DBFSVolumeClient#deleteObject or get/put for input streams.
------
Without this flag being set as true the validateLocalFilePath throws error in case allowed volume paths are not present. 
By sending this flag the processor skips validation of file path. 

## Testing
Explicitly removed this flag from connection in driver example to show that delete record will work without this flag being set. 

I will be adding another PR for actual feature gate which restricts using volume operations directly from jdbc url connection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Version Changelog
 
+## [v1.0.9-oss] - 2025-08-19
+### Added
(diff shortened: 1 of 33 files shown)


samikshya-db:
Nit : can you make the title and description more readable?

PR #966, opened 2025-08-28
Fetch columnar results incrementally/lazily

jayantsing-db:
## Description
when JDBC is used with Arrow disabled and operating in Thrift mode (the default mode or when useThriftClient=1), the Hive ThriftServer returns results in a columnar binary format (non-Arrow). Currently, before constructing the result set, JDBC fetches all the required data at once and buffers the entire result set in memory.

This PR modifies this behavior for the disabled Arrow and Thrift mode case, so that the result set maintains only a limited number of rows in memory at a time, controlled by the connection parameter RowsFetchedPerBlock. As the client reads rows, the result set will issue additional fetch requests incrementally.

Implications: The JVM no longer buffers the entire result set upfront, which previously caused a sharp spike in heap memory usage before garbage collection could reclaim space. This spike often led to OutOfMemory (OOM) errors. With this change, peak heap memory usage is reduced, increasing more gradually, which allows the JVM’s memory management to work more effectively.

Following improvements were observed when executing a SQL query on 1 million rows and iterating through the result set without printing the rows.

Existing behaviour. Sudden spike with higher peak:
image

Gradual heap increase within memory management bounds with less peak:
image

## Testing

- e2e tests
- unit tests
- fake service tests

Note this does not lower the max heap usage by a lot (instead make memory increase gradual). I will make further improvements on how we process the data.


src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -253,7 +253,7 @@ DatabricksResultSet execute(
             getResultSetResp(
                 response.getStatus(),
                 response.getOperationHandle(),
-                response.toString(),
```
jayantsing-db, line 256: when the driver is run in debug log mode, stringifying thrift model classes incur a huge memory allocation cost (~700 MB to ~170 MB improvement). henceforth, removing these in the favour of telemetry and other statement id logging

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -595,9 +605,27 @@ public boolean isFirst() throws SQLException {
...
+   *
+   * @return {@code true} if the cursor is on the last row, {@code false} otherwise
+   * @throws SQLException if the result set is closed or an error occurs
+   */
   @Override
   public boolean isLast() throws SQLException {
     checkIfClosed();
+    if (executionResult instanceof LazyThriftResult) {
+      return !executionResult.hasNext();
+    }
```
vikrantpuppala, line 628: do we need to check instanceof? can we not simply not do !hasNext() for all implementations?

src/main/java/com/databricks/jdbc/api/impl/ExecutionResultFactory.java
```diff
@@ -96,7 +94,7 @@ private static IExecutionResult getResultHandler(
     LOGGER.info("Processing result of format {} from Thrift server", resultFormat);
     switch (resultFormat) {
       case COLUMN_BASED_SET:
-        return getResultSet(convertColumnarToRowBased(resultsResp, parentStatement, session));
```
vikrantpuppala, line 99: can we not remove these older function implementations?

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
```diff
@@ -0,0 +1,256 @@
...
+
+  /**
+   * Gets the number of rows in the current batch.
+   *
+   * @return the number of rows in the current batch
+   */
+  @Override
+  public long getRowCount() {
+    // Return the number of rows in the current batch
+    return currentBatch != null ? currentBatch.size() : 0;
```
vikrantpuppala, line 199: isn't this incorrect? other implementations return the total row count (from metadata, etc.)

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
```diff
@@ -0,0 +1,256 @@
...
+
+  /**
+   * Gets the chunk count. Always returns 0 for thrift columnar results.
+   *
+   * @return 0 (thrift results don't use chunks like Arrow)
+   */
+  @Override
+  public long getChunkCount() {
+    // For thrift columnar results, we don't have chunks in the same sense as Arrow
+    return 0;
```
vikrantpuppala, line 210: should we instead throw an error since this is an internal interface?

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
```diff
@@ -0,0 +1,256 @@
...
+
+  private void loadCurrentBatch() throws DatabricksSQLException {
+    currentBatch = extractRowsFromColumnar(currentResponse.getResults());
+    currentBatchIndex = -1; // Reset batch index
+    totalRowsFetched += currentBatch.size();
+    LOGGER.debug(
+        "Loaded batch with {} rows, total fetched: {}", currentBatch.size(), totalRowsFetched);
+  }
+
+  private void fetchNextBatch() throws DatabricksSQLException {
```
vikrantpuppala, line 221: let's add javadoc across all functions

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
```diff
@@ -0,0 +1,256 @@
...
+      // So, this is a safeguard to ensure we don't move past the end
+      return false;
+    }
+
+    // Check if we've reached the maxRows limit
+    boolean hasRowLimit = maxRows > 0;
+    if (hasRowLimit && globalRowIndex + 1 >= maxRows) {
+      hasReachedEnd = true;
+      return false;
+    }
```
shivam2680, line 124: we're already checking this in hasNext. can remove this block ig

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -595,9 +605,27 @@ public boolean isFirst() throws SQLException {
...
+   * </ul>
+   *
+   * @return {@code true} if the cursor is on the last row, {@code false} otherwise
+   * @throws SQLException if the result set is closed or an error occurs
+   */
   @Override
   public boolean isLast() throws SQLException {
     checkIfClosed();
+    if (executionResult instanceof LazyThriftResult) {
+      return !executionResult.hasNext();
```
shivam2680, line 627: isLast should return true only when we are on last row, not when there are no more rows
```suggestion
      return executionResult.getCurrentRow() >= 0 && ! executionResult.hasNext();
```

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
jayantsing-db: couple of reasons this is like this:
- this function updates the state of hasReachedEnd
- hasNext and next are mostly independent and there is no strict enforcement on client code on how they use them together. So, for safety/early-exits, you may find some duplicate checks

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
jayantsing-db: >true only when we are on last row

isn't  !executionResult.hasNext() doing the same thing? or are you referring to cases like empty result set, etc.?

src/main/java/com/databricks/jdbc/api/impl/ExecutionResultFactory.java
jayantsing-db: This function getResultSet(List<List<Object>> rows) is currently being used elsewhere too so keeping it for now.

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
jayantsing-db: Yes you are right. This is different from other execution results in the sense that it does not return the whole row count. I have documented this limitation [here](https://github.com/databricks/databricks-jdbc/blob/29197e89801ac854a1a036d581660688ad154093/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java#L589) in the usage (only one usage in DatabricksResultSet) of this method in the case of LazyThriftResult. From this execution result perspective, it only maintains a certain row count at a time and this method returns the row count maintained at this point. I am a bit hesitant to throw exception here. 

If the limitation turns out to be a pain point for customer (i don't expect customers use `isAfterLast()` that often), the fix will be to send all fetch requests just to get the row count.

src/main/java/com/databricks/jdbc/api/impl/LazyThriftResult.java
jayantsing-db: done

PR #973, opened 2025-08-29
[Bugfix] Fix backward incompatible change in SDK : browser auth flow

samikshya-db:
## Description
- This is a breaking change for browser based auth, this PR fixes it.

## Testing
- E2E manual testing
- Added a Unit test too

- I am only fixing the bleeding issue as part of this PR. And not the long term context flag addition in the driver. Added a TODO for the same.
- Additional context : https://docs.google.com/document/d/17wkI_I16_k6Igl0QPDWZkgbg-7VApebyO4D1yVeXfpM/edit?tab=t.0#heading=h.ig8zh4bt1u8z 

NO_CHANGELOG=true

### DIFF
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/common/ClientConfigurator.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/common/ClientConfigurator.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.net.ServerSocket;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


shivam2680 (approved):

src/main/java/com/databricks/jdbc/dbclient/impl/common/ClientConfigurator.java
```diff
@@ -196,6 +197,8 @@ public void setupU2MConfig() throws DatabricksParsingException {
         .setAuthType(DatabricksJdbcConstants.U2M_AUTH_TYPE)
         .setHost(host)
         .setClientId(clientId)
+        .setOAuthBrowserAuthTimeout(
+            Duration.ofHours(1)) // TODO : add a browser timeout connection config
```
vikrantpuppala, line 201: 1. what was the previous default?
2. can we remove the todo? since this is a user interactive flow, i don't think we ever need this to be configurable?

samikshya-db: Also added this in the PR description for more context. Thanks

vikrantpuppala: Thank you

vikrantpuppala (approved):

PR #978, opened 2025-09-02
ResultSchema refactor

sreekanth-db:
## Description

Refactoring the code base to use local classes for ResultSchema, ColumnInfo and ColumnInfoTypeName instead of using SDK classes.

Reason for refactoring: To support geospatial datatypes (GEOGRAPHY, GEOMETRY) in ColumnInfoTypeName class. 

NO_CHANGELOG=true


src/main/java/com/databricks/jdbc/model/core/ColumnInfoTypeName.java
```diff
@@ -0,0 +1,25 @@
+package com.databricks.jdbc.model.core;
+
```
gopalldb, line 2: similar comment here as well

sreekanth-db: Added the comment

gopalldb (approved):

PR #979, opened 2025-09-02
Fix getFunctions in SEA + add a workaround for getFunctions in Thrift

samikshya-db:

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
@@ -2,6 +2,7 @@ package com.databricks.jdbc.api.impl;
 
 import static com.databricks.jdbc.common.MetadataResultConstants.*;
 import static com.databricks.jdbc.dbclient.impl.common.CommandConstants.METADATA_STATEMENT_ID;
+import static com.databricks.jdbc.dbclient.impl.sqlexec.CommandName.LIST_FUNCTIONS;
 import static com.databricks.jdbc.dbclient.impl.sqlexec.ResultConstants.CLIENT_INFO_PROPERTIES_RESULT;
 
 import com.databricks.jdbc.api.impl.converters.ConverterHelper;
@@ -9,8 +10,10 @@ import com.databricks.jdbc.api.internal.IDatabricksConnectionInternal;
 import com.databricks.jdbc.api.internal.IDatabricksSession;
 import com.databricks.jdbc.common.*;
 import com.databricks.jdbc.common.util.DriverUtil;
+import com.databricks.jdbc.common.util.WildcardUtil;
 import com.databricks.jdbc.dbclient.impl.common.MetadataResultSetBuilder;
 import com.databricks.jdbc.dbclient.impl.common.StatementId;
+import com.databricks.jdbc.dbclient.impl.sqlexec.CommandBuilder;
 import com.databricks.jdbc.exception.DatabricksSQLException;
 import com.databricks.jdbc.log.JdbcLogger;
 import com.databricks.jdbc.log.JdbcLoggerFactory;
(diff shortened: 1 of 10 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
```diff
@@ -1534,11 +1537,22 @@ public ResultSet getFunctions(String catalog, String schemaPattern, String funct
...
+          functionNamePattern =
+              "%"; // This is because functionName is a required parameter in thrift flow.
+        }
+        return session
+            .getDatabricksMetadataClient()
+            .listFunctions(session, catalog, schemaPattern, functionNamePattern);
+      }
     } catch (Exception e) {
-      LOGGER.error(e, "Unable to fetch functions, returning empty result set");
+      LOGGER.error(e, "Unable to fetch functions with error {}, returning empty result set", e);
```
jayantsing-db, line 1555: nit: e is already passed to log statement, any specific need to re-add and stringify it.

src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
```diff
@@ -1618,4 +1632,18 @@ private void throwExceptionIfConnectionIsClosed() throws SQLException {
...
+    Statement internalStatement = connection.createStatement();
+    String showFunctionsSqlCommand =
+        new CommandBuilder(catalog, session)
+            .setSchemaPattern(schemaPattern)
+            .setFunctionPattern(functionNamePattern)
+            .getSQLString(LIST_FUNCTIONS);
+    DatabricksResultSet rs =
+        (DatabricksResultSet) internalStatement.executeQuery(showFunctionsSqlCommand);
+    return new MetadataResultSetBuilder(connection.getConnectionContext())
+        .getFunctionsResult(rs, catalog);
```
jayantsing-db, line 1647: should we use iDatabrickClient#executeStatement directly as it is an internal impl detail? This may save us from creating a full statement object and then closing the statement object.

jayantsing-db, line 1647: please consider closing these resources like result-set, statement for any leaks, etc.

jayantsing-db (approved):
Some suggestions inline.

src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
```diff
@@ -26,4 +26,5 @@ public class CommandConstants {
       "SHOW KEYS" + IN_CATALOG_SQL + IN_ABSOLUTE_SCHEMA_SQL + IN_ABSOLUTE_TABLE_SQL;
   public static final String SHOW_FOREIGN_KEYS_SQL =
       "SHOW FOREIGN KEYS" + IN_CATALOG_SQL + IN_ABSOLUTE_SCHEMA_SQL + IN_ABSOLUTE_TABLE_SQL;
+  public static final String USE_CATALOG_SQL = "USE CATALOG %s";
```
gopalldb, line 29: where is this used?

src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
```diff
@@ -1534,11 +1537,22 @@ public ResultSet getFunctions(String catalog, String schemaPattern, String funct
...
-          .listFunctions(session, catalog, schemaPattern, functionNamePattern);
+      if (connection.getConnectionContext().enableShowCommandsForGetFunctions()) {
+        return getFunctionsFromCommand(
+            catalog,
+            schemaPattern,
+            functionNamePattern); // TODO : remove this once thrift implementation is fixed.
+      } else {
+        if (WildcardUtil.isNullOrEmpty(functionNamePattern)) {
+          functionNamePattern =
+              "%"; // This is because functionName is a required parameter in thrift flow.
```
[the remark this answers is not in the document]
samikshya-db: In case of thrift get function request, it is a required field. We can't get past the null error in the model itself. Hence, adding this.

src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
samikshya-db: Good catch. done

src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
samikshya-db: We want to log the error too 

`Unable to fetch functions, returning empty result set java.lang.NullPointerException: Cannot invoke "com.databricks.jdbc.api.internal.IDatabricksStatementInternal.getStatementId()" because "parentStatement" is null`

inplace of 

`2025-09-10 19:31:37 SEVERE com.databricks.jdbc.api.impl.DatabricksDatabaseMetaData#getFunctions - Unable to fetch functions, returning empty result set`

Issue #981, opened 2025-09-05
[BUG] MetadataParser unable to parse Struct<Decimal(10, 20)>


sshpuntoff:
**Describe the bug**
 The driver throws an ArrayIndexOutOfBoundsException when parsing STRUCT metadata containing DECIMAL fields with precision and scale (e.g., DECIMAL(18,2)). The
  MetadataParser.parseStructMetadata() method incorrectly splits STRUCT field definitions at commas inside parentheses, causing malformed field parsing.

**To Reproduce**
You can simply test the logic in MetaDataParserTest
```
  @Test
  @DisplayName("parseStructMetadata with DECIMAL precision and scale")
  public void testParseStructMetadata_WithDecimalPrecisionScale() {
    String metadata = "STRUCT<name:STRING, amount:DECIMAL(18,2)>";
    Map<String, String> expected = new LinkedHashMap<>();
    expected.put("name", "STRING");
    expected.put("amount", "DECIMAL(18,2)");

    Map<String, String> actual = MetadataParser.parseStructMetadata(metadata);
    assertEquals(
        expected,
        actual,
        "Parsed struct metadata with DECIMAL precision and scale should handle parentheses correctly.");
  }
```

**Client side logs**
```
  java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
      at com.databricks.jdbc.api.impl.MetadataParser.parseStructMetadata(MetadataParser.java:30)
      at com.databricks.jdbc.api.impl.ComplexDataTypeParser.parseToStruct(ComplexDataTypeParser.java:104)
      at com.databricks.jdbc.api.impl.ComplexDataTypeParser.convertValueNode(ComplexDataTypeParser.java:131)
      at com.databricks.jdbc.api.impl.ComplexDataTypeParser.parseToArray(ComplexDataTypeParser.java:73)
```

**Client Environment (please complete the following information):**
 - OS: MAC/Linux
 - Java version 21
 - Java vendor OpenJDK
 - Driver Version 

**Additional context**
Complex datatypes and Arrow must be enabled for this to happen in a run time environment

sshpuntoff:
Here is my suggested fix - it is working for me locally
https://github.com/databricks/databricks-jdbc/pull/982

josecsotomorales:
@jayantsing-db I was able to reproduce this issue. Can you please take a look at it? 🙏🏻

And excellent work @sshpuntoff!! 🚀

[...]

PR #982, opened 2025-09-05
fix: struct metadata parsing for nested decimal fields

sshpuntoff:
## Description
When parsing Struct<Decimal(10,20)>, the driver currently raises an exception: 
```
  java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
      at com.databricks.jdbc.api.impl.MetadataParser.parseStructMetadata(MetadataParser.java:30)
      at com.databricks.jdbc.api.impl.ComplexDataTypeParser.parseToStruct(ComplexDataTypeParser.java:104)
      at com.databricks.jdbc.api.impl.ComplexDataTypeParser.convertValueNode(ComplexDataTypeParser.java:131)
      at com.databricks.jdbc.api.impl.ComplexDataTypeParser.parseToArray(ComplexDataTypeParser.java:73)
```
This occurs because the current implementation does not expect nested comma's in datatypes. 
By tracking the Depth of the `()` characters, we can successfully parse complex datatypes.

Issue:
https://github.com/databricks/databricks-jdbc/issues/981

## Testing
Unit tests expanded and local testing done.

This is a bug in the implementation of the MetaDataParser and is pretty direct to re-produce and correct.

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/MetadataParser.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/MetadataParser.java
@@ -1,16 +1,12 @@
 package com.databricks.jdbc.api.impl;
 
 import com.databricks.jdbc.exception.DatabricksDriverException;
-import com.databricks.jdbc.log.JdbcLogger;
-import com.databricks.jdbc.log.JdbcLoggerFactory;
 import com.databricks.jdbc.model.telemetry.enums.DatabricksDriverErrorCode;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
 /** Utility class for parsing metadata descriptions into structured type mappings. */
 public class MetadataParser {
-
-  private static final JdbcLogger LOGGER = JdbcLoggerFactory.getLogger(MetadataParser.class);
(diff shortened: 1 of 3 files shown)


src/main/java/com/databricks/jdbc/api/impl/MetadataParser.java
```diff
@@ -94,25 +94,33 @@ public static String parseMapMetadata(String metadata) {
    * @return an array of field definitions in the STRUCT
    */
   private static String[] splitFields(String metadata) {
-    int depth = 0;
+    int angleBracketDepth = 0;
+    int parenDepth = 0;
```
sshpuntoff, line 98: there is a case to be made that just a single
`int depth = 0;` would work correctly here, and the code below just needs to be 
```
if (ch == '<' || ch == '(') {
 depth ++;
}
else if (ch == '>' || ch == ')') {
    depth --;
}
```
Which would simplify the change set even more.
Let me know if you prefer a minimal diff over the explicit variable names.

PR #987, opened 2025-09-09
[PECOBLR-747] Thin jar

shivam2680:
## Description
Customer reported that the Databricks thin JAR "specifies no dependencies at all" in its POM, preventing dependency introspection tools from working. Users cannot:
  - View dependency trees with mvn dependency:tree or Clojure's -Stree
  - Pin, override, or exclude specific dependency versions
  - Resolve dependency conflicts using standard build tools

Solution:
Deploy the existing thin JAR as a separate artifact (databricks-jdbc-thin) with a POM that declares all runtime dependencies:

  1. Created thin_public_pom.xml: Declares runtime dependencies (Arrow, BouncyCastle, Jackson, Thrift, etc.) with explicit versions. exclude provide and test scope deps.
  2. Updated release workflow: Deploy thin JAR using mvn deploy-file with dependency metadata
  3. Updated README.md: Added thin JAR usage instructions

  - Enables introspection: clj -Sdeps '{:deps {com.databricks/databricks-jdbc-thin {:mvn/version "1.0.9-oss"}}}' -Stree now works
  - Separate artifact: com.databricks:databricks-jdbc-thin vs com.databricks:databricks-jdbc
  - Standard dependency management: Users can exclude/override versions using Maven/Gradle
  - No build changes: Uses existing thin JAR (2.5MB), just adds deployment metadata

## Testing
  - mvn validate -f thin_public_pom.xml - POM structure is valid
  - Existing thin JAR builds correctly: target/databricks-jdbc-*-thin.jar (2.5MB)

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
@@ -969,6 +969,11 @@ public class DatabricksConnectionContext implements IDatabricksConnectionContext
     return null;
   }
 
+  @Override
+  public boolean enableShowCommandsForGetFunctions() {
+    return getParameter(DatabricksJdbcUrlParams.ENABLE_SHOW_COMMAND_FOR_GET_FUNCTIONS).equals("1");
+  }
+
   private static boolean nullOrEmptyString(String s) {
     return s == null || s.isEmpty();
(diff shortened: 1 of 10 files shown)


samikshya-db (approved):
I have some comments, please merge after that. Thanks for making the changes 🫙

Issue #958, opened 2025-08-25
[BUG] - query that projects various columns intermittently throws Cannot invoke "org.apache.hive.service.rpc.thrift.TGetFunctionsReq.isSetOperationId()" because "req$9" is null


dqmdev:
OSS JDBC driver 109

Simple select-project queries  may execute or fail with

Error: HTTP request failed by code: 500, status line: HTTP/1.1 500 Internal Server Error. Thrift Header : Cannot invoke "org.apache.hive.service.rpc.thrift.TGetFunctionsReq.isSetOperationId()" because "req$9" is null
SQLState:  08000
ErrorCode: 0

i.e. select * from tarrbint where tarrbint

create table  if not exists TARRBINT (RNUM int , CARRBINT ARRAY< BIGINT >)
using  parquet
location 's3://..../tarrbint/tarrbint.parquet'

select * from cognos.gosalesdw1021.go_org_dim can fail with the same error as above.

create external table if not exists gosalesdw1021.GO_ORG_DIM ( ORGANIZATION_KEY int , ORGANIZATION_CODE1 string , ORGANIZATION_CODE2 string , ORGANIZATION_CODE3 string , ORGANIZATION_CODE4 string , ORGANIZATION_CODE string , ORGANIZATION_PARENT string , ORGANIZATION_LEVEL smallint )
using parquet  
location 's3://......' 

This issue is occurring a lot and makes the driver unusable.

dqmdev:
[databricks_jdbc.log.0.zip](https://github.com/user-attachments/files/21972077/databricks_jdbc.log.0.zip)

samikshya-db:
The initial issue you reported has now been resolved by fixing the null case in getFunctions. However, this surfaced a deeper underlying concern related to state leakage in TCLIService.Client. We’ve identified that sequential calls using the same client instance can lead to state being unintentionally carried over if a previous call encountered a server error.

To address this, we have already raised a pull request that ensures a new TCLIService.Client is created for every call, thereby preventing such state leakage. This fix is being reviewed and will be included in the release after the next.


Closing the issue as you will no longer face this issue in the next release that is scheduled this week. Feel free to open in case you have concerns.

PR #999, opened 2025-09-15
Fix race condition in thrift client 

samikshya-db:
## Description
- This issue was uncovered as part of bug raised in #958. Although the issue was resolved as part of PR #979 - the cause is deeper than that - TCLIService.client causes state leakage in case previous server call  
- This PR fixes the issue.
- Internal doc on root-cause and analysis: https://docs.google.com/document/d/12kciNIqy-kL1e0HOQs9JOjJlb2zWeNFl02e9kO772PE/edit?tab=t.0

## Testing
- Reset code before [this commit](https://github.com/databricks/databricks-jdbc/commit/fcfa625756c00d8f8740faa999b448dec1ab7a85) where we fixed getFunctions, and the following code no longer throws error 

```
for (int i = 0; i < 10; i++) {
      try {
        DatabaseMetaData metaData = con.getMetaData();
        metaData.getFunctions("main", "testschema", null).close(); // Metadata call

        Statement stmt = con.createStatement();
        stmt.execute("select * from samikshyachand.default.hello_table"); // SQL query
        stmt.close();

        Thread.sleep(100); // Add a small delay
      } catch (Exception e) {
        e.printStackTrace();
      }
    }
```

### DIFF
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksHttpTTransport.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksHttpTTransport.java
@@ -87,7 +87,9 @@ public class DatabricksHttpTTransport extends TTransport {
 
   @Override
   public void write(byte[] buf, int off, int len) {
-    requestBuffer.write(buf, off, len);
+    synchronized (requestBuffer) {
+      requestBuffer.write(buf, off, len);
+    }
   }
 
   @Override
@@ -115,9 +117,13 @@ public class DatabricksHttpTTransport extends TTransport {
       LOGGER.debug("Thrift tracing header: " + traceHeader);
       request.addHeader(TracingUtil.TRACE_HEADER, traceHeader);
     }
-
+    byte[] requestPayload;
+    synchronized (requestBuffer) {
+      requestPayload = requestBuffer.toByteArray();
+      requestBuffer.reset();
+    }
     // Set the request entity
-    request.setEntity(new ByteArrayEntity(requestBuffer.toByteArray()));
+    request.setEntity(new ByteArrayEntity(requestPayload));
 
     // Execute the request and handle the response
     long httpRequestStartTime = System.currentTimeMillis();
@@ -145,9 +151,6 @@ public class DatabricksHttpTTransport extends TTransport {
       LOGGER.error(e, errorMessage);
       throw new TTransportException(TTransportException.UNKNOWN, errorMessage, e);
     }
-
-    // Reset the request buffer
-    requestBuffer.reset();
   }
 
   @Override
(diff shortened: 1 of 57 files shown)


src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -51,13 +47,15 @@ final class DatabricksThriftAccessor {
...
-  private final ThreadLocal<TCLIService.Client> thriftClient;
   private final DatabricksConfig databricksConfig;
   private final boolean enableDirectResults;
   private final int asyncPollIntervalMillis;
   private final int maxRowsPerBlock;
   private final String connectionUuid;
+  private final String endpointUrl;
+  private final IDatabricksConnectionContext connectionContext;
   private TProtocolVersion serverProtocolVersion = JDBC_THRIFT_VERSION;
+  private ThreadLocal<TCLIService.Client> FAKE_SHARED_CLIENT;
```
gopalldb, line 58: nit: don't use caps for non final variables

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksHttpTTransport.java
```diff
@@ -115,9 +117,13 @@ public void flush() throws TTransportException {
       LOGGER.debug("Thrift tracing header: " + traceHeader);
       request.addHeader(TracingUtil.TRACE_HEADER, traceHeader);
     }
-
+    byte[] requestPayload;
+    synchronized (requestBuffer) {
```
jayantsing-db, line 121: If i am understanding correctly, a JDBC connection will now have a single thrift client. So, when a JDBC connection is used in different threads concurrently, it will use the same thrift client. Now, if we don't put a sync mechanism, different threads using the same thrift client will fail because:
- thread A can increment the seq number of thrift client from x to x + 1
- thread B can receive the response for the seq number x

This section of the code with all good intent and purposes tries to put a sync mechanism in place. But I am afraid, this might not be sufficient.

So the thrift calls happen in this order sequentially for a thread:
``` 
TCLIService.Client.ExecuteStatement -> seqid ++ -> transport.write -> transport.flush -> <wait for the http response> -> transport.read (when reading it checks the seqid with that present in message)
```

The current sync mechanism only protects transport.write and transport.read. And things could fail in exciting ways because the whole chain is not isolated/protected.

Happy to chat if this does not make sense at all.

jayantsing-db: Ideally, JDBC is a sync protocol and client app is not advised to use the same connection in different threads but there is no strict enforcement. Also i wrote a basic test on top of these changes that uses a JDBC connection across many threads and executes statements. I didn't encounter any errors (but i believe like other threading issues that's by chance that threads didn't step over each other)

jayantsing-db:
I see that Samikshya has included a reproducer and these changes fix the reproducer while the earlier thread-local changes were failing. Do we have a understanding of why thread-local changes were working and these are passing (in case there is any misunderstanding)? I can try to run the reproducer with earlier changes and these changes to understand more.

PR #1001, opened 2025-09-16
Default value of RowsFetchedPerBlock changed to 100K

jayantsing-db:
## Description
RowsFetchedPerBlock denotes the max rows that JDBC buffers in memory when running in Arrow disabled inline mode. 100K default captures typical use cases with an efficient balance between execution time and memory usage.


jayantsing-db:
Why are fake service integration tests are failing for this?

PR #1010, opened 2025-09-17
Fixed ResultSetMetadata for SEA inline mode

msrathore-db:
## Description

For SEA this type is parsed later using a function. So we need not strip the base type here. So reverted the code to use stripTypeName instead. But the stripTypeName function isn't correct either since it will incorrectly convert ARRAY<DECIMAL(10,2)> to ARRAY<DECIMAL and later parse it incorrectly.  
## Testing

Tested the SQL smoke test locally

NO_CHANGELOG=true

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
@@ -108,7 +108,7 @@ public class DatabricksResultSetMetaData implements ResultSetMetaData {
               .columnTypeClassName(DatabricksTypeUtil.getColumnTypeClassName(columnTypeName))
               .columnType(columnType)
               .columnTypeText(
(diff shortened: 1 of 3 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
```diff
@@ -108,7 +108,7 @@ public DatabricksResultSetMetaData(
               .columnTypeClassName(DatabricksTypeUtil.getColumnTypeClassName(columnTypeName))
               .columnType(columnType)
               .columnTypeText(
-                  metadataResultSetBuilder.stripBaseTypeName(
+                  metadataResultSetBuilder.stripTypeName(
```
samikshya-db, line 111: can you please add a test case which broke the multiDBRs to begin with?

msrathore-db: Added

PR #1012, opened 2025-09-18
Fixed stripTypeName function

msrathore-db:
## Description

Updated the stripTypeName function to strip the type only when it contains the close bracket in the end. This would parse ARRAY<DECIMAL(10,2)> correctly. 
## Testing

Added unit test

[PECOBLR-951](https://databricks.atlassian.net/browse/PECOBLR-951)


jayantsing-db (commented):
Could you please remind me that why we need such client side parsing for metadata?

PR #1015, opened 2025-09-18
Fix thin jar creation and pushing to maven

samikshya-db:


shivam2680 (approved):

samikshya-db:
> let's check with workflow dispatch.

I won't be able to - as the workflow itself does not exist in the list of actions yet (as the code is not on main)

PR #1021, opened 2025-09-25
[PECOBLR-874] Handle close gracefully during statement timeout

gopalldb:
## Description

Currently error during closing statement while timeout will change the errorType from timeout, we want to preserve the same


samikshya-db (approved):
Thanks for making this change!

nit : add unit tests

PR #1025, opened 2025-09-26
Added support for enableMultipleCatalogSupport param

msrathore-db:
## Description

Added support for the connection param of enableMultipleCatalogSupport.
The default value is kept to 1.
The behaviour remains the same when the param is enabled. When the param
is disabled, the behaviour is as follows:
- getCatalogs() returns only the current catalog
- Metadata operations are done on the current catalog
- setCatalog() method does nothing (doesn't execute USE CATALOG)
- The SQL query of USE CATALOG can still be executed.

To maintain the current catalog, we make the query to get the current
catalog everytime when this param is disabled. 
 
[PECOBLR-938](https://databricks.atlassian.net/browse/PECOBLR-938)

[PECOBLR-938]:
https://databricks.atlassian.net/browse/PECOBLR-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PECOBLR-938]: https://databricks.atlassian.net/browse/PECOBLR-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClient.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClient.java
@@ -48,8 +48,7 @@ public class DatabricksMetadataSdkClient implements IDatabricksMetadataClient {
   }
 
   private String autoFillCatalog(String catalog, IDatabricksSession session) throws SQLException {
-    if (isMultipleCatalogSupportDisabled()
-        && (WildcardUtil.isNullOrWildcard(catalog) || catalog.isEmpty())) {
+    if (isMultipleCatalogSupportDisabled()) {
       String currentCatalog = session.getCurrentCatalog();
       return (currentCatalog != null && !currentCatalog.isEmpty()) ? currentCatalog : "";
     }
--- a/src/test/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClientTest.java
+++ b/src/test/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClientTest.java
@@ -934,39 +934,4 @@ public class DatabricksMetadataSdkClientTest {
     assertEquals(METADATA_STATEMENT_ID, actualResult.getStatementId());
     assertEquals(2, ((DatabricksResultSetMetaData) actualResult.getMetaData()).getTotalRows());
   }
-
-  @Test
-  void testListSchemasWithMultipleCatalogSupportDisabledAndSpecificCatalog() throws SQLException {
-    when(session.getComputeResource()).thenReturn(mockedComputeResource);
-    when(mockClient.getConnectionContext()).thenReturn(mock(IDatabricksConnectionContext.class));
-    when(mockClient.getConnectionContext().getEnableMultipleCatalogSupport()).thenReturn(false);
-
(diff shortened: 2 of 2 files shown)


src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClient.java
```diff
@@ -42,6 +42,43 @@ public DatabricksMetadataSdkClient(IDatabricksClient sdkClient) {
...
+  private boolean isMultipleCatalogSupportDisabled() {
+    return sdkClient.getConnectionContext() != null
+        && !sdkClient.getConnectionContext().getEnableMultipleCatalogSupport();
+  }
+
+  private String autoFillCatalog(String catalog, IDatabricksSession session) {
+    if (isMultipleCatalogSupportDisabled()
+        && (catalog == null || catalog.isEmpty() || catalog.equals("%") || catalog.equals("*"))) {
+      String currentCatalog = getCurrentCatalog(session);
+      return (currentCatalog != null && !currentCatalog.isEmpty()) ? currentCatalog : "";
```
gopalldb, line 54: is there a way to check if this is empty for existing driver?

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftServiceClient.java
```diff
@@ -67,6 +67,32 @@ void setServerProtocolVersion(TProtocolVersion serverProtocolVersion) {
     this.serverProtocolVersion = serverProtocolVersion;
   }
 
+  private boolean isMultipleCatalogSupportEnabled() {
+    return connectionContext == null || connectionContext.getEnableMultipleCatalogSupport();
+  }
+
+  private String getCurrentCatalog(IDatabricksSession session) {
+    try {
+      DatabricksResultSet resultSet =
```
gopalldb, line 76: This is common to both SEA and Thrift, we can refactor this to some common class later

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftServiceClient.java
```diff
@@ -332,6 +358,20 @@ public DatabricksResultSet listCatalogs(IDatabricksSession session) throws SQLEx
         String.format("Fetching catalogs using Thrift client. Session {%s}", session.toString());
     LOGGER.debug(context);
     DatabricksThreadContextHolder.setSessionId(session.getSessionId());
+
+    // If multiple catalog support is disabled, return only the current catalog
+    if (!isMultipleCatalogSupportEnabled()) {
```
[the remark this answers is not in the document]
msrathore-db: No it isn't needed for thrift since it already handles it in the thrift logic.

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClient.java
msrathore-db: We've a function for isNullOrWildcard. Will use that. Thanks

gopalldb (approved):

PR #1034, opened 2025-10-07
[PECOBLR-642] Add header to metadata commands in SEA mode

vikrantpuppala:
## Description

Add header to metadata commands in SEA mode

## Testing

Unit + manual testing

### DIFF
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
@@ -192,7 +192,7 @@ public class DatabricksSdkClient implements IDatabricksClient {
     ExecuteStatementResponse response;
     try {
       Request req = new Request(Request.POST, STATEMENT_PATH, apiClient.serialize(request));
-      req.withHeaders(getHeaders("executeStatement"));
+      req.withHeaders(getHeaders("executeStatement", statementType, false));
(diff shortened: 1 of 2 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
```diff
@@ -1077,4 +1077,9 @@ public TelemetryLogLevel getTelemetryLogLevel() {
     return TelemetryLogLevel.parse(
         getParameter(DatabricksJdbcUrlParams.TELEMETRY_LOG_LEVEL), TelemetryLogLevel.DEBUG);
   }
+
+  @Override
+  public boolean isSeaSyncMetadataEnabled() {
```
nikhilsuri-db: Should we also check for "true" value?

vikrantpuppala: all of our params are 0/1, if we want to support true/false, we'd have to do it across all params

samikshya-db (approved):

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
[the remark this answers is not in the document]
vikrantpuppala: We don't expect this connection flag to be used in normal scenarios. Only in excruciating circumstances - will be undocumented. Is telemetry still warranted?

PR #1039, opened 2025-10-11
Incorporate telemetry logLevels across the driver

samikshya-db:
## Description
- LogLevels were added to telemetry as part of #1032 , This PR integrates it across the driver
- Add telemetry for timeout exceptions (it was not added before)
- For error throws for fallback mechanisms we are only throwing a warn rather than error/fatal (Example : fallback to Thrift, try-catch to build resultSet etc.)

## Testing
- Manually tested logLevels
- Added unit testing

- The following are not included as part of the PR/ will be included in later PRs : 
  - We plan to add more logs to other telemetry logLevels subsequently. Also see[ internal doc](https://docs.google.com/document/d/1EmsW4jYMtpvjlLovyRQsOwqIBAN-MAcOZKAS0Bm0c7Q/edit?tab=t.0)
  - Testing telemetry for driver retry logic and make sure it works well
  -  Group null statementIds before sending it to the endpoint. 
  - Addition of redacted stackTrace in telemetry 
  - For errors in batch operation, we will have to figure out how to log only once in fatal telemetry log.
  - Add logLevel field in server proto

### DIFF
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/common/TimeoutHandler.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/common/TimeoutHandler.java
@@ -68,7 +68,7 @@ public class TimeoutHandler {
               "Statement execution timed-out after %d seconds. Operation: %s",
               timeoutSeconds, operationDescription);
       LOGGER.error(timeoutErrorMessage);
-      throw new DatabricksTimeoutException(timeoutErrorMessage, null, internalErrorCode);
+      throw new DatabricksTimeoutException(timeoutErrorMessage, /*cause*/ null, internalErrorCode);
     }
   }
 


src/main/java/com/databricks/jdbc/dbclient/impl/common/TimeoutHandler.java
```diff
@@ -59,8 +68,7 @@ public void checkTimeout() throws DatabricksTimeoutException {
               "Statement execution timed-out after %d seconds. Operation: %s",
               timeoutSeconds, operationDescription);
       LOGGER.error(timeoutErrorMessage);
-
-      throw new DatabricksTimeoutException(timeoutErrorMessage);
+      throw new DatabricksTimeoutException(timeoutErrorMessage, null, internalErrorCode);
```
gopalldb: nit: add /* paramName */ for null

Issue #1047, opened 2025-10-22
[FEATURE] Support for More Than 256 Parameters or Option to Disable Parameterized Queries for Batched Inserts


josecsotomorales:
**Summary**

The Databricks OSS JDBC driver, as of v3.0.1, currently limits prepared statements to a maximum of 256 parameters. This limitation prevents the driver from efficiently executing large batched inserts, as each statement must be sent individually.

This issue proposes either:
	1.	Lifting or expanding the 256-parameter limit, or
	2.	Providing a connection/property flag to disable parameterized queries — allowing client-side interpolation and enabling true multi-row batch inserts.

⸻

**Background**

In its current form, the OSS driver always relies on parameterized execution for batches. However, this approach:
	•	Caps the total number of bound parameters to 256.
	•	Forces each statement in a batch to be executed individually.
	•	Severely limits throughput for large datasets.

For example, inserting a large DataFrame results in multiple round-trips, each limited by the 256-parameter constraint:

```sql
INSERT INTO table (col1, col2, col3) VALUES (?, ?, ?)
```

→ repeated many times, instead of a single optimized multi-row insert like:

```sql
INSERT INTO table (col1, col2, col3)
VALUES (1, 2, 3), (4, 5, 6), (7, 8, 9)
```

⸻

**Problem**
	•	The 256-parameter cap prevents scaling inserts efficiently.
	•	The driver does not provide a configuration option to disable parameterized queries or to switch to interpolated mode.
	•	As a result, large batch writes perform poorly, especially in data-intensive workloads (e.g., Spark-to-JDBC writes).

⸻

**Proposed Enhancement**

Introduce one or both of the following:
	1.	Configurable parameter limit — allow increasing the number of parameters supported per prepared statement.
	2.	Property to disable query parameters (e.g., DisableParameterizedQueries=true) — enabling client-side interpolation and multi-row value construction for batched inserts.

This would provide parity with how other JDBC drivers (e.g., Snowflake, PostgreSQL, MySQL) handle large batches efficiently.

⸻

**Benefits**
	•	Significantly improved batch insert performance.
	•	Reduced network round-trips.
	•	Enables scalability for large data ingestion workloads.

⸻

**Next Steps**
	•	Evaluate feasibility of increasing or removing the 256-parameter cap.
	•	Discuss adding a driver-level flag to control parameterized query behavior.
	•	Test performance improvements with Spark’s JDBC write path once implemented.

[...]

PR #1049, opened 2025-10-24
Support batched writes

sshpuntoff:
## Description

Added a new `BATCH_INSERT_SIZE` parameter that controls the number of rows to batch while performing an insert.
We were previously limited to `256 Parameters` which was not very performant for large numbers of writes.

By calling the `interpolateParameters` method, (which is flagged off by supportManyParameters), we are able to write larger batches of records. Setting the default value to 1000 rows is meant to be conservative and should prevent users from hitting the max payload size. 

```
Error : com.databricks.internal.sdk.core.error.platform.InvalidParameterValue: The statement field must not exceed a length of 16777216 bytes.
```

In our testing, we discovered that timestamps were not properly converted by the `interpolateSQL` method, so that logic been patched/unit tested as well.

## Testing

Additional unit tests that verify the batching of records.
Local behavior verification and - writing to databricks via jdbc.

You currently need the following parameters to enable this behavior:

`supportManyParameters=1` + `EnableBatchedInserts=1` + `BatchInsertSize=5000`

There is an existing `MAX_BATCH_SIZE` which is used by similar code, but I think they are separate use cases that warrant different parameters.

Related Issue
https://github.com/databricks/databricks-jdbc/issues/1047

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -27,5 +27,6 @@
 - Fix: SQLInterpolator failing to escape temporal fields and special characters.
 - Fixed: Errors in table creation when using BIGINT, SMALLINT, TINYINT, or VOID types.
 - Fixed: PreparedStatement.getMetaData() now correctly reports TINYINT columns as Types.TINYINT (java.lang.Byte) instead of Types.SMALLINT (java.lang.Integer).
+- Fixed: TINYINT to String conversion to return numeric representation (e.g., "65") instead of character representation (e.g., "A").
 ---
 *Note: When making changes, please add your change under the appropriate section with a brief description.* 
--- a/src/main/java/com/databricks/jdbc/api/impl/converters/ByteConverter.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/converters/ByteConverter.java
@@ -66,6 +66,6 @@ public class ByteConverter implements ObjectConverter {
 
   @Override
   public String toString(Object object) throws DatabricksSQLException {
-    return new String(new byte[] {toByte(object)});
+    return String.valueOf(toByte(object));
   }
 }
--- a/src/test/java/com/databricks/jdbc/api/impl/converters/ByteConverterTest.java
+++ b/src/test/java/com/databricks/jdbc/api/impl/converters/ByteConverterTest.java
@@ -7,6 +7,8 @@ import com.databricks.jdbc.exception.DatabricksSQLException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 
 public class ByteConverterTest {
   private final byte NON_ZERO_OBJECT = 65;
@@ -74,9 +76,10 @@ public class ByteConverterTest {
     assertTrue(exception.getMessage().contains("Unsupported char conversion operation"));
   }
 
-  @Test
-  public void testConvertToString() throws DatabricksSQLException {
-    assertEquals(new ByteConverter().toString(NON_ZERO_OBJECT), "A");
+  @ParameterizedTest
+  @CsvSource({"-128, -128", "-1, -1", "0, 0", "1, 1", "10, 10", "65, 65", "127, 127"})
+  public void testConvertToString(byte input, String expected) throws DatabricksSQLException {
+    assertEquals(new ByteConverter().toString(input), expected);
   }
 
   @Test


src/main/java/com/databricks/jdbc/common/util/SQLInterpolator.java
```diff
@@ -10,9 +10,42 @@
 import java.util.regex.Pattern;
 
 public class SQLInterpolator {
-  private static String escapeApostrophes(String input) {
+  protected static String escapeInputs(String input) {
```
sshpuntoff: changed this to make it easier to test.

Added support for things like emojis (verified in our env with 🧱)
We were having some issues with `\n` characters. 

I think the parametrization does a lot of this cleansing automatically, so this is needed for the behavior to match.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -147,72 +147,248 @@ private boolean canUseBatchedInsert() {
...
+        }
+
+        LOGGER.debug("Successfully processed chunk {}-{}", startIndex + 1, endIndex);
+        return chunkSize;
+
+      } catch (Exception e) {
+        lastException = e;
+        LOGGER.error(
+            "Failed to execute chunk {}-{} (attempt {}/{}): {}",
+            startIndex + 1,
```
jayantsing-db: inclusive/exclusive? Can we please correct other log statements.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -147,72 +147,248 @@ private boolean canUseBatchedInsert() {
...
+        // When using parameterized queries, respect the 256 parameter limit from Databricks backend
+        int maxRowsByParameterLimit =
+            DatabricksJdbcConstants.MAX_QUERY_PARAMETERS / parametersPerRow;
+
+        // Ensure we have at least 1 row per chunk
+        if (maxRowsByParameterLimit < 1) {
+          maxRowsPerChunk = 1;
+        } else {
+          maxRowsPerChunk = maxRowsByParameterLimit;
+        }
```
jayantsing-db: Why is the maxRowsPerChunk in this case not capped by the the total number of rows, databricksBatchParameterMetaData.size()?

Ig this is handled later on by?
```
int endIndex =
            Math.min(startIndex + maxRowsPerChunk, databricksBatchParameterMetaData.size());
```

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -147,72 +147,248 @@ private boolean canUseBatchedInsert() {
...
       LOGGER.debug("Successfully processed {} rows in chunks", processedRows);
+      // Clear the batch after successful execution to prevent accidental re-execution
+      clearBatch();
       return allUpdateCounts;
 
+    } catch (DatabricksBatchUpdateException e) {
+      // Re-throw batch update exceptions (these already have proper update counts)
+      // Clear the batch to prevent duplicate inserts if application retries
+      try {
+        clearBatch();
```
jayantsing-db: clearBatch is a public API within JDBC spec. Is this a correct use? What if users explicitly do not want to clear the batch on failure and want to manually execute the same batch again? Is that possible? If so, maybe we gate this using a connection prop: ClearBatchOnError. Default false but for your use-case, set to true.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -242,10 +418,24 @@ private long[] executeIndividualStatements() throws DatabricksBatchUpdateExcepti
         for (int i = sqlQueryIndex + 1; i < largeUpdateCount.length; i++) {
           largeUpdateCount[i] = Statement.EXECUTE_FAILED;
         }
+        // Clear the batch to prevent duplicate execution on retry
+        // WARNING: Due to lack of transaction support, any successfully executed statements
+        // before this failure have already been committed and cannot be rolled back
+        try {
+          clearBatch();
```
jayantsing-db: same as above.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
sshpuntoff: Oh I see. This is specifically the "Non Interpolated" code path. 

We are limited in this case to the 
("Number of parameters per row")  / (256 parameter limit)

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -147,72 +147,248 @@ private boolean canUseBatchedInsert() {
...
-      LOGGER.error("Error executing batched INSERT: {}", e.getMessage(), e);
+      // Unexpected exception - mark all as failed
+      LOGGER.error("Unexpected error executing batched INSERT: {}", e.getMessage(), e);
       long[] failedCounts = new long[databricksBatchParameterMetaData.size()];
       for (int i = 0; i < failedCounts.length; i++) {
         failedCounts[i] = Statement.EXECUTE_FAILED;
       }
+      // Clear the batch to prevent duplicate inserts on retry
+      try {
+        clearBatch();
```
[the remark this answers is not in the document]
sshpuntoff: I am seeing substantial duplication of writes when I do bulk writes without these clearBatch calls, I ended up pushing these clearBatch calls back up,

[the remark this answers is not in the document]
jayantsing-db: If you are seeing duplicates because of your client application explicitly retrying the same batch insert, I would rather recommend to call the clearBatch in the client code rather than the JDBC code.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -385,6 +385,8 @@ private long[] executeBatchedInsert() throws DatabricksBatchUpdateException {
       }
 
       LOGGER.debug("Successfully processed {} rows in chunks", processedRows);
+      // Clear the batch after successful execution per JDBC spec
+      clearBatch();
```
sshpuntoff: @jayantsing-db - as described in the longer thread, I have removed the finally block
and am now only clearing the batch on successful writes per the JDBC doc.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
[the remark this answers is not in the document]
jayantsing-db: This makes a lot more sense to me now. Unlike your custom code, we can't call clearBatch in spark writer API. Also, the spec recommends the clearing batch on success. Will take a look at the new changes.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
jayantsing-db: >I saw you added support for transactions recently, do you think these should be automatically wrapped in transactions?

appreciate, you noticed. The way I think about batch and transaction is that when statements are executed in a batch (executeBatch), individual statements follow the current auto-commit state. So if auto-commit is false, all the statements in a batch are part of the current transaction. However, when the batch is successful, my understanding the user has to call rollback or commit and that the executeBatch itself should not do so behind the scenes automatically.

src/main/java/com/databricks/jdbc/api/impl/DatabricksPreparedStatement.java
```diff
@@ -147,68 +147,254 @@ private boolean canUseBatchedInsert() {
...
+   * @param insertInfo Parsed INSERT statement information
+   * @param startIndex Starting row index (inclusive)
+   * @param endIndex Ending row index (exclusive)
+   * @param allUpdateCounts Array to store update counts
+   * @param interpolateParameters Whether to interpolate parameters into SQL
+   * @param depth Current recursion depth (0 = top level)
+   * @return Number of rows successfully processed
+   * @throws DatabricksBatchUpdateException if a single row fails after all retries
+   */
+  private int executeChunkWithRetryAndSplit(
```
samikshya-db: I am only worried about this causing duplicate insertions without idempotency!

sshpuntoff: What I have observed is that the databricks API will reject a payload outright.
I have not seen any "Partial writes" - I have done some testing which triggers this logic explicitly (by using overly large batch sizes).

The validation query I run (using tpch datasets) is 
```
SELECT COUNT(*), COUNT(DISTINCT *) FROM <catalog>.<table>
```
I have done this with multiple spark executors and have not seen duplicates generated.

We can consider making the retries here less opinionated (ie. let the batches fail if the BATCH_SIZE is too big)
I am certain we should have some retries to account for server size back-pressure (rate limiting etc)..

sshpuntoff: Ack on the refactor. Will take care of that.

sshpuntoff: @samikshya-db - Following up on the duplicate data concern. 

Multi-row INSERT statements should be atomic in Databricks. A statement like this should insert all 3 rows or none:
```sql
INSERT INTO users VALUES (1, 'Alice'), (2, 'Bob'), (3, 'Charlie')
```

The retry logic only retries failed statements (which should have committed zero rows). Successful statements are never retried. Example with 10 rows where row 5 fails:
```
INSERT rows 0-9 (×4) → FAIL (0 rows committed)
  ├─ INSERT rows 0-4 → ✅ 5 committed, done
  └─ INSERT rows 5-9 → FAIL (0 committed)
      ├─ INSERT rows 5-6 → FAIL (0 committed)
      │   ├─ INSERT row 5 → ❌ FAIL
      │   └─ INSERT row 6 → ✅ 1 committed, done
      └─ INSERT rows 7-9 → ✅ 3 committed, done

Result: 9 rows inserted
```

Assuming INSERT atomicity, we only retry statements that were rejected by the server, so duplicates shouldn't occur. Is there's a scenario where Databricks SQL can partially succeed?

The prior duplicate data bug was related to a missing `clearBatch` call after a successful execute batch statement, which violated the JDBC spec. 

Please let me know if you have any questions.

sshpuntoff: logic removed.

samikshya-db (approved):
Thanks for the great contribution, @sshpuntoff!  I appreciate your thoroughness and how patiently you addressed all the comments!

samikshya-db:
I will merge your changes (most likely by tomorrow) after the 3.0.4 release.

josecsotomorales:
@samikshya-db @jayantsing-db all tests are passing! Can we please merge this PR to avoid it from getting stale? Thanks!

jayantsing-db (approved):

PR #1052, opened 2025-10-29
Fix regression in sdk (changes in token federation)

samikshya-db:
## Description
- The issue : 
  - An [E2E test](https://docs.google.com/document/d/1McX4IgD-ZBTtiNXNUrEemsjbVj3oeoQYDwegjjln6UA/edit?ouid=113104269373381935368&tab=t.0#heading=h.k832j8h2svg) conducted with JDBC, with additional SDK-level logging, confirmed that the SDK was refreshing tokens on every call—even when successive calls occurred within seconds. This behavior occurred despite Databricks M2M tokens being valid for 59 minutes. As a result, the token endpoint was hit excessively, eventually triggering global rate limits and throttling for the IP.
  -  Each [getToken call in the SDK](https://github.com/databricks/databricks-jdbc/blob/a17b84c1a0418094a8434f56246c764fa235d19b/src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksHttpTTransport.java#L166) fetched a new token from the server because the SDK did not cache tokens correctly. This was traced to a regression in the SDK’s token management logic.
- This PR ensures that the SDK is configured once in the constructor, preventing repeated token endpoint calls. We also plan to perform a broader SDK code audit to identify and address any similar issues going forward.

## Testing
- Tested manually using M2M flow : The token retrieval is now performed only once when M2M creds are used.
- Unit tests

- Internal doc : https://docs.google.com/document/d/1McX4IgD-ZBTtiNXNUrEemsjbVj3oeoQYDwegjjln6UA/edit?tab=t.g07tag19b223

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -10,6 +10,7 @@
 * Updated sdk version from 0.65.0 to 0.67.3
 
 ### Fixed
+- Fixed SQL syntax error when LIKE queries contain empty ESCAPE clauses.
 - Fix: driver failing to authenticate on token update in U2M flow.
 - Fix: driver failing to parse complex data types with nullable attributes.
 - Fixed: Resolved SDK token-caching regression causing token refresh on every call. SDK is now configured once to avoid excessive token endpoint hits and rate limiting.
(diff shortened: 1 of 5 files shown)


src/test/java/com/databricks/jdbc/dbclient/impl/common/ClientConfiguratorTest.java
```diff
@@ -170,7 +170,7 @@ void testM2MWithJWT() throws DatabricksSQLException {
     assertEquals("https://sample-host.18.azuredatabricks.net", config.getHost());
     assertEquals("test-client", config.getClientId());
     assertEquals("custom-oauth-m2m", provider.authType());
-    assertEquals(DatabricksJdbcConstants.M2M_AUTH_TYPE, config.getAuthType());
```
gopalldb: no other tests needed?

PR #1053, opened 2025-10-29
Suggestion: modify tinyint to string conversion

vankovivan:
## Description
Currently the conversion from tinyint to string is done using `String(byte[])` which depends on the platform's default charset and also is [unspecified](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-byte:A-) when the bytes are not valid in the default charset (e.g. may produce "?" instead of a sensible value).

## Testing
Tested with the existing unit tests.

-


src/test/java/com/databricks/jdbc/api/impl/converters/ByteConverterTest.java
```diff
@@ -76,7 +76,7 @@ public void testConvertToChar() {
 
   @Test
   public void testConvertToString() throws DatabricksSQLException {
-    assertEquals(new ByteConverter().toString(NON_ZERO_OBJECT), "A");
+    assertEquals(new ByteConverter().toString(NON_ZERO_OBJECT), "65");
   }
```
samikshya-db: I may be skeptical here (and just so we don't add a breaking change later)

Can you add a more elaborate test for negative boundary values too? 


```suggestion
  @ParameterizedTest
  @CsvSource({
      "-128, -128",
      "-1, -1",
      "0, 0",
      "1, 1",
      "10, 10",
      "65, 65",
      "127, 127"
  })
  public void testConvertToString(byte input, String expected) throws DatabricksSQLException {
      assertEquals(new ByteConverter().toString(input), expected);
  }
```

samikshya-db (commented):
Thanks for the valuable fix @vankovivan ! LGTM overall


Added a minor comment on testing. Can you also add an entry in NEXT_CHANGELOG.md file ? something like : 
`Fixed: TINYINT values converted to correct numeric strings`

PR #1070, opened 2025-11-10
Shade SLF4J

jayantsing-db:


pom.xml
```diff
@@ -552,6 +552,10 @@
                   <pattern>org.osgi</pattern>
                   <shadedPattern>com.databricks.internal.osgi</shadedPattern>
                 </relocation>
```
tejassp-db: The convention I have seen is to retain the full original package name in the shaded jar and prefix it with the org name in this case `com.databricks.shaded`. In this instance it will be renamed to `com.databricks.shaded.org.slf4j`.

NEXT_CHANGELOG.md
```diff
@@ -8,13 +8,14 @@
...
-- Updated sdk version from 0.65.0 to 0.67.3
+- Updated sdk version from 0.65.0 to 0.67.3.
 
 ### Fixed
 - Fixed SQL syntax error when LIKE queries contain empty ESCAPE clauses.
 - Fix: driver failing to authenticate on token update in U2M flow.
 - Fix: driver failing to parse complex data types with nullable attributes.
 - Fixed: Resolved SDK token-caching regression causing token refresh on every call. SDK is now configured once to avoid excessive token endpoint hits and rate limiting.
 - Fixed: TimestampConverter.toString() returning ISO8601 format with timezone conversion instead of SQL standard format.
+- Shade SLF4J to avoid conflicts with user applications.
```
tejassp-db: How will SLF4J providers like `log4j` plug in? The SLF4J providers will be implementing the interfaces in `org.slf4j` package path and our internal code will be referencing `shaded.org.slf4j` package path? Do we have a deeper understanding of how this will work?

Examples of things breaking when shading SLF4J 
- https://github.com/snowflakedb/snowflake-jdbc/issues/16
- https://github.com/aws/aws-sdk-java-v2/issues/5108

pom.xml
```diff
@@ -552,6 +552,10 @@
...
+                <relocation>
+                  <pattern>org.slf4j</pattern>
```
tejassp-db: Can't we just use maven scope `provided` and expect the customer to provide a version of `slf4j`?

tejassp-db (changes requested):
I have reservations on whether this is the right thing to do.

NEXT_CHANGELOG.md
jayantsing-db: Let me try making it part of POM and use exclude

pom.xml
jayantsing-db: We have tried this and this breaks out-of-the-box functioning of the driver like in the other Databricks driver

NEXT_CHANGELOG.md
tejassp-db: What does "shaded log4j plugin" mean?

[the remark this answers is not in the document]
tejassp-db: But isn't the point that they are using slf4j for logging and the driver version conflicts with their application. Wouldn't they want to log driver logs with slf4j as well?

What about all the customers who are ok with the included slf4j version - do they need to shade a provider as well? It will not work out of the box for them anymore.

Isn't this the same as not including slf4j logging in our driver?

PR #1073, opened 2025-11-11
Unified release workflow for shaded and thin JARs

gopalldb:
## Summary
This PR implements a unified release workflow that publishes both the shaded uber JAR and thin JAR to Maven Central in a single GitHub Actions workflow.

## Changes
1. **Added `build-helper-maven-plugin`** to `pom.xml`:
   - Attaches the unshaded JAR with `-thin` classifier before the shade plugin runs
   - Preserves the original JAR as `databricks-jdbc-${VERSION}-thin.jar`

2. **Updated `.github/workflows/release.yml`**:
   - Added GPG configuration for signing
   - Kept existing step to deploy shaded JAR (main artifact)
   - Added new step to deploy thin JAR as separate artifact with custom POM

3. **Removed `.github/workflows/release-thin.yml`**:
   - No longer needed - both JARs now published in single unified workflow

## Published Artifacts
- **Shaded JAR**: `com.databricks:databricks-jdbc:3.0.1` (37MB, includes all dependencies)
- **Thin JAR**: `com.databricks:databricks-jdbc-thin:3.0.1` (2.4MB, driver classes only)

## Testing
```bash
mvn clean && mvn -B -DskipTests package
# Verified both JARs created successfully:
# - target/databricks-jdbc-3.0.1.jar (37MB shaded)
# - target/databricks-jdbc-3.0.1-thin.jar (2.4MB thin)
```

## Backward Compatibility
✅ Existing users are unaffected - the main artifact remains the shaded uber JAR

---
NO_CHANGELOG=true


pom.xml
```diff
@@ -474,6 +474,30 @@
           </excludes>
         </configuration>
       </plugin>
+      <!-- Attach unshaded JAR before shade plugin runs -->
```
shivam2680: why do we need this block?

.github/workflows/release.yml
```diff
@@ -27,9 +27,15 @@ jobs:
           gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg-passphrase: GPG_PASSPHRASE
 
-      - name: Publish to the Maven Central Repository
+      - name: Configure GPG
         run: |
-          # Deploy main artifacts (uber JAR, sources, javadoc)
+          echo "allow-loopback-pinentry" >> ~/.gnupg/gpg-agent.conf
+          echo "pinentry-mode loopback" >> ~/.gnupg/gpg.conf
+          gpg-connect-agent reloadagent /bye
```
jayantsing-db: As far as i remember, i don't think this is needed. But no problem trying out with this 😅

PR #1087, opened 2025-11-17
Fix download status in chunk download 

samikshya-db:
## Description
- There were 2 issues with chunk download status : 
  -   `DOWNLOAD_SUCCEEDED` was set before the data was actually available locally, making the state misleading. This is because we marked as success even before reading the response stream. This PR ensures DOWNLOAD_SUCCEEDED is only set when the data fully read
  - Not all errors are caught when chunk is download/processing fails. any subsequent processing failures correctly transition to PROCESSING_FAILED instead of attempting an invalid transition to DOWNLOAD_FAILED.

## Testing
- added unit tests

### DIFF
--- a/.github/workflows/release-thin.yml
+++ b/.github/workflows/release-thin.yml
@@ -31,28 +31,96 @@ jobs:
           echo "pinentry-mode loopback" >> ~/.gnupg/gpg.conf
           gpg-connect-agent reloadagent /bye
 
-      - name: Build thin JAR
+      - name: Build thin JAR with sources and javadocs
         run: |
-          mvn -B -DskipTests package
+          # Build main artifacts including sources and javadocs
+          mvn -B -DskipTests package source:jar javadoc:jar
 
-      - name: Verify thin jar exists
+      - name: Sign all thin JAR artifacts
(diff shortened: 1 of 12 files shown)


src/test/java/com/databricks/jdbc/api/impl/arrow/ArrowResultChunkStatusTest.java
```diff
@@ -0,0 +1,276 @@
...
+/** Tests around when DOWNLOAD_SUCCEEDED is set and error handling during body read. */
+public class ArrowResultChunkStatusTest {
+
+  @AfterEach
+  void clearTelemetry() {
+    TelemetryCollector.getInstance().exportAllPendingTelemetryDetails();
+  }
+
+  @Test
+  void setsDownloadSucceededOnlyAfterBodyRead_successfulFlow() throws Exception {
```
gopalldb: Thanks for elaborate tests!

jayantsing-db:
Looks like the PR got merged. This will not only not-fix the issue at hand but also cause regressions in memory allocations.

PR #1088, opened 2025-11-18
Fix omission of specific complex nested type

madhav-db:
## Description

Adds specific nesting type to complex data types

## Testing

Added unit tests and tested manually


gopalldb (approved):
Thanks for fixing!

PR #1093, opened 2025-11-19
Improve logging when jdbc is shaded

vikrantpuppala:
## Description

Improve logging when jdbc is shaded
## Testing

Unit tests + manually in benchmarking


samikshya-db (approved):
I have always needed this, thanks.

PR #1101, opened 2025-11-24
Fix deadlock by completing futures for upfront-fetched chunk links

jayantsing-db:
## Description
NO_CHANGELOG=true

Complete link futures for upfront-fetched chunks to prevent deadlock

When chunk links are fetched upfront, the corresponding futures were never completed, causing threads to wait indefinitely. Now we complete these futures in the constructor for all pre-fetched chunks.

## Testing

- Unit tests
- Manual testing


gopalldb (approved):

jayantsing-db:
Why are tests failing?

PR #1107, opened 2025-11-25
[ES-1614194] Include OAuth HTML resources

jayantsing-db:
## Description

OAuth U2M is broken in databricks-jdbc for AWS and GCP. This PR fixes it by including the required resources needed by SDK classes.

## Testing

Tested by locally installing the jar and running the reproducer on the ticket.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixed
 
 - Fix driver crash when using `INTERVAL` types.
+- Fix U2M by including SDK OAuth HTML callback resources.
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -578,7 +578,6 @@
                     <exclude>javax/**</exclude>
                     <exclude>jakarta/**</exclude>
                     <exclude>net/jcip/**</exclude>
-                    <exclude>oauth/**</exclude>
```
jayantsing-db: This only adds:
- success html page
- failure html template page

both are required by sdk.

PR #1109, opened 2025-11-26
[PECOBLR-954] Close statements based on Execute Statement Response SEA

madhav-db:
## Description

Close statements based on Execute Statement Response SEA

Flow:
1. User calls `executeQuery`
2. `executeInternal` in `DatabricksStatement` is called, and creates direct same thread executor
3. Runs on same thread, calls `getResultFromClient`
4. After execution, server responds with `status: CLOSED`, we detect the state and call `markAsClosed` which soft closes the statement (does not shut down the executor)
5. We return result set to user
6. User calls `statement.close()`
7. `DatabricksStatement` sees `isClosed=true`, skips the server close (since it's already closed), closes result set and shuts down the executor.

NO_CHANGELOG=true


src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -274,6 +274,12 @@ public DatabricksResultSet executeStatement(
     if (responseState != StatementState.SUCCEEDED && responseState != StatementState.CLOSED) {
       handleFailedExecution(response, statementId, sql);
     }
+
+    if (responseState == StatementState.CLOSED && parentStatement != null) {
```
gopalldb: I see an issue here. We are marking statement as closed here, which in turn will be closing the resultSet. Now, we are also returning the resultSet here. How will that work?

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -816,6 +816,26 @@ void checkIfClosed() throws DatabricksSQLException {
...
+      try {
+        this.resultSet.close();
+      } catch (DatabricksSQLException e) {
+        LOGGER.warn("Error closing result set: {}", e.getMessage());
+      }
+      this.resultSet = null;
+    }
+    this.connection.closeStatement(this);
+    DatabricksThreadContextHolder.clearStatementInfo();
+    shutDownExecutor();
```
gopalldb: as discussed offline, we are on executor thread already, which is actually the same thread executor. We can remove this also.

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
madhav-db: Removed, added the flow to the description

Issue #1103, opened 2025-11-25
[BUG] Settings RowsFetchedPerBlock to 0 hangs the driver


vankovivan:
**Describe the bug**
Settings configuration property RowsFetchedPerBlock to 0 hangs various driver methods related to execution and result set iteration (resultset.next(), statement.executeUpdate()).

**To Reproduce**
```java
        // ...
        props.setProperty("RowsFetchedPerBlock", "0");

        Driver driver = DriverManager.getDriver(url);
        try (Connection conn = driver.connect(url, props)) {
            try (PreparedStatement st = conn.prepareStatement("create or replace view default.jjj(id) as select 1")) {
                st.executeUpdate();
            }
        }
```

**Expected behavior**
Driver validates the provided configuration properties.

**Client Environment (please complete the following information):**
 - Driver Version 3.0.3

gopalldb:
We will look into this.

gopalldb:
Tracking ticket: PECOBLR-1309

PR #1112, opened 2025-11-27
[PECOBLR-1309] Add validation in user input (in case of integers)

samikshya-db:
## Description
- This addresses the issue in #1103 (i.e., rowsFetchedPerBlock field) and also adds validation across other integer fields that require to be positive. 

## Testing
- Added unit tests

### DIFF
--- a/src/test/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClientTest.java
+++ b/src/test/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksMetadataSdkClientTest.java
@@ -17,15 +17,11 @@ import com.databricks.jdbc.api.internal.IDatabricksSession;
 import com.databricks.jdbc.common.CommandName;
 import com.databricks.jdbc.common.IDatabricksComputeResource;
 import com.databricks.jdbc.common.StatementType;
-import com.databricks.jdbc.common.TelemetryLogLevel;
-import com.databricks.jdbc.common.util.DatabricksThreadContextHolder;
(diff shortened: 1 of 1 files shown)


src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
```diff
@@ -83,13 +88,18 @@ connectionContext, getTelemetryExecutorService(), databricksConfig),
             key,
             (k, existing) -> {
               if (existing == null) {
-                return new TelemetryClientHolder(
-                    new TelemetryClient(connectionContext, getTelemetryExecutorService()), 1);
+                try {
+                  return new TelemetryClientHolder(
+                      new TelemetryClient(connectionContext, getTelemetryExecutorService()), 1);
+                } catch (Exception e) {
+                  // Validation or other errors during client creation - fail silently
```
gopalldb: lets log also

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
```diff
@@ -175,6 +185,7 @@ private static final class TelemetryClientHolder {
   }
 
   private static String keyOf(IDatabricksConnectionContext context) {
-    return context.getHostForOAuth();
+    String host = context.getHostForOAuth();
```
gopalldb: shall we do context.host instead?

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
samikshya-db: Done ✅

PR #1125, opened 2025-12-09
Add tests for StreamingChunkProvider and ChunkLinkFetcher

jayantsing-db:
## Description

- Add comprehensive unit tests for `StreamingChunkProvider`
- Add unit tests for `ChunkLinkFetcher`
- Minor fix in `StreamingChunkProvider`

## Testing

- Unit tests


src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
```diff
@@ -344,6 +344,15 @@ private void linkPrefetchLoop() {
...
         break;
+      } catch (Exception e) {
+        LOGGER.error("Unexpected error in link prefetch: {}", e.getMessage(), e);
+        prefetchError =
+            new DatabricksSQLException(
+                "Unexpected error in link prefetch: " + e.getMessage(),
+                e,
+                DatabricksDriverErrorCode.CHUNK_READY_ERROR);
+        notifyChunkCreated(); // Wake up any waiting consumer to check the error
+        break;
```
jayantsing-db: This guarantees that the link fetch thread does not simply die without informing the system on runtime exceptions like NPE, etc. This was caught during unit testing! I have added tests on this.

PR #1129, opened 2025-12-10
Add tokenCache in custom credentialProviders

samikshya-db:
## Description
- We should be caching the tokens for custom oauth providers
- This is to incorporate caching as mentioned in the[ internal code audit ](https://docs.google.com/document/d/1O6cmsqYw6JMYIzW6NrJR_RLZw6KQkzXNy0oKzRK7xAM/edit?tab=t.0)
- There are 2 options to add tokenCache : one is re-using the persistent token cache being used in refresh flow, another is to extend cachedTokenSource (in-memory). The decision is made by benchmarking both : [internal doc](https://docs.google.com/document/d/1aO6befanIuO-OIJ4NZMx3SK3R7JXmwtV54LCpFo02EA/edit?tab=t.0). 

Screenshot 2025-12-12 at 5 18 17 PM
 

## Testing
- added unit tests
- Tested each of the flow end to end.

### DIFF
--- a/src/test/java/com/databricks/jdbc/api/impl/arrow/ChunkDownloadTaskTest.java
+++ b/src/test/java/com/databricks/jdbc/api/impl/arrow/ChunkDownloadTaskTest.java
@@ -172,6 +172,11 @@ public class ChunkDownloadTaskTest {
 
     when(remoteChunkProvider.getCompressionCodec()).thenReturn(CompressionCodec.NONE);
 
+    // Mock chunkLinkDownloadService.getLinkForChunk() in case the chunk link becomes invalid
+    CompletableFuture<ExternalLink> linkFuture =
+        CompletableFuture.completedFuture(mockExternalLink);
+    when(chunkLinkDownloadService.getLinkForChunk(anyLong())).thenReturn(linkFuture);
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/databricks/jdbc/dbclient/impl/common/ClientConfigurator.java
```diff
@@ -107,6 +109,34 @@ private static String createUniqueIdentifier(String host, String clientId, List<
...
+   * preventing cache collisions between different connections.
+   *
+   * @param clientId The OAuth client ID (used for encryption passphrase generation)
+   * @param scopes The OAuth scopes (used for cache path identification if non-empty)
+   * @return An EncryptedFileTokenCache instance
+   */
+  private TokenCache createTokenCache(String clientId, List<String> scopes) {
+    String userHome = System.getProperty("user.home");
+    Path homeDir = Paths.get(userHome);
+    Path databricksDir = homeDir.resolve(".config/databricks-jdbc/oauth");
```
gopalldb: does this work on windows?

gopalldb (approved):
Thanks, make sure that you verify these manually

PR #1144, opened 2025-12-16
PECOBLR-1121 Tests to validate Arrow patch across JVM versions.

tejassp-db:
## Description
Added tests to validate Arrow patch code paths. Added Maven profiles to validate the behaviour across JVM versions and with/without "--add-opens=java.base/java.nio=ALL-UNNAMED" JVM arguments.

By default, JVM version 11 is assumed. To use other JVM versions, the toolchain needs to be setup to point to the correct Java versions on the local machine in .m2/toolchains.xml.

## Testing
Tests to validate Arrow patch behaviour.

Stacked PR.


src/test/java/com/databricks/jdbc/api/impl/arrow/ArrowBufferAllocatorUnknownManagerTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.condition.JRE;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Test that the fallback {@code DatabricksBufferAllocator} is used with an unknown allocation
+ * manager type when the creation of {@code RootAllocator} is not possible in the current JVM.
+ *
+ * <p>This test is in a separate class to ensure it runs in a fresh JVM with the system property set
+ * before Arrow's static initialization.
```
vikrantpuppala: are we sure this is true? I think maven surefire defaults to reusing the same jvm fork: https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#reuseForks

src/test/java/org/apache/arrow/memory/DatabricksArrowPatchTest.java
```diff
@@ -0,0 +1,243 @@
...
+   * on JVM >= 17.
+   */
+  @Test
+  @Tag("Jvm17PlusAndArrowToNioReflectionDisabled")
+  @EnabledOnJre({JRE.JAVA_17, JRE.JAVA_21})
+  public void testArrowThrowsExceptionOnMissingAddOpensJvmArgs() {
+    Throwable throwable = null;
+    try {
+      RootAllocator allocator = new RootAllocator();
+      allocator.close(); // Unreachable code.
```
vikrantpuppala: I think similar to the first PR we should try to write here?

src/test/resources/arrow/chunk_1.arrow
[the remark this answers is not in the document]
vikrantpuppala: I think we should. Even if do not generate on every test run, it would be useful to retain the generator.

PR #1148, opened 2025-12-17
Remove custom provider as sdk is now fixed

samikshya-db:
## Description
- The custom azure u2m was a pain because we had to make the sdk flow for aws inline with the custom flow : this is not possible. Now that sdk has fixed their flows, we should start using it for azure too. 

## Testing
- Tested it manually using Azure U2M


gopalldb:
Thanks for cleaning this up :)

PR #1156, opened 2025-12-22
PECOBLR-1121 Arrow patch to circumvent Arrow issues with JDk 16+.

tejassp-db:
## Description
Databricks server shares query results in Arrow format for easy cross language functionality. The JDBC driver experiences compatibility issues with JDK 16 and later versions when processing Arrow results. 

This problem arises from stricter encapsulation of internal APIs in newer Java versions, which affects the driver's use of the Apache Arrow result format consumption with the Apache Arrow library. The JDBC driver is used in partner solutions, where they do not have control of the runtime environment, and the workaround of setting JVM arguments is not feasible. 

## Testing
Tests are added in other stacked PRs.

Its a stacked PR.


src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -0,0 +1,1265 @@
...
+  private final long id = idGenerator.incrementAndGet();
+  private final ReferenceManager referenceManager;
+  private final BufferManager bufferManager;
+  private final long addr;
+  private long readerIndex;
+  private long writerIndex;
+
+  // ---- Databricks patch start ----
+  private final HistoricalLog historicalLog =
+      DEBUG ? new HistoricalLog(DEBUG_LOG_LENGTH, "ArrowBuf[%d]", id) : null;
```
vikrantpuppala: do we even need to worry about the historical log? seems like its usage is null checked everywhere anyway?

src/main/java/org/apache/arrow/memory/DatabricksAllocationReservation.java
```diff
@@ -0,0 +1,104 @@
...
+    if (nBytes < 0) {
+      return false;
+    }
+    // Check if reservation would exceed limits
+    long currentReservation = reservedSize.get();
+    long newReservation = currentReservation + nBytes;
+    if (newReservation > allocator.getHeadroom() + currentReservation) {
+      return false;
+    }
+    reservedSize.addAndGet(nBytes);
```
vikrantpuppala: this is not thread safe, should we use compareAndSet?

src/main/java/org/apache/arrow/memory/DatabricksArrowBuf.java
```diff
@@ -0,0 +1,738 @@
...
+
+  @Override
+  public ReferenceManager getReferenceManager() {
+    return referenceManager;
+  }
+
+  @Override
+  public long capacity() {
+    return capacity;
+  }
```
vikrantpuppala: i think a bunch of these methods do not have a changed override behaviour, can we remove them so that it is easy to review and maintain?

src/main/java/org/apache/arrow/memory/DatabricksArrowBuf.java
```diff
@@ -0,0 +1,738 @@
...
+    this.bufferManager = bufferManager;
+    this.capacity = capacity;
+    this.offset = 0;
+    this.readerIndex = 0;
+    this.writerIndex = 0;
+
+    if (capacity > Integer.MAX_VALUE) {
+      throw new IllegalArgumentException(
+          "DatabricksArrowBuf does not support capacity > Integer.MAX_VALUE");
+    }
```
vikrantpuppala: - why this limit?
- this is missing from the other constructor
- can we reuse constructors

src/main/java/org/apache/arrow/memory/util/MemoryUtil.java
```diff
@@ -0,0 +1,263 @@
...
+      // This exception will get swallowed, but it's necessary for the static analysis that ensures
+      // the static fields above get initialized
+      final RuntimeException failure =
+          new RuntimeException(
+              "Failed to initialize MemoryUtil. You must start Java with "
+                  + "`--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED` "
+                  + "(See https://arrow.apache.org/docs/java/install.html)",
+              e);
+      // ---- Databricks patch start ----
+      // ---- Remove 'failure.printStackTrace();'
```
gopalldb: can we log the stack trace?

src/main/java/org/apache/arrow/memory/DatabricksAllocationReservation.java
```diff
@@ -0,0 +1,104 @@
...
+    return true;
+  }
+
+  @Override
+  public ArrowBuf allocateBuffer() {
+    assertNotUsed();
+    if (!used.compareAndSet(false, true)) {
```
gopalldb: can we add logging?

src/main/java/org/apache/arrow/memory/DatabricksArrowBuf.java
tejassp-db: Idea is to have all behaviour captured within a single class and make all things explicit. Also it is not dependent on any changes in the base class `ArrowBuf`.

src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -0,0 +1,1265 @@
...
+public class ArrowBuf implements AutoCloseable {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ArrowBuf.class);
+
+  // ---- Databricks patch start ----
+  // ---- Copied verbatim from BaseAllocator. We avoid initializing static fields of BaseAllocator
+  // ---- to avoid unsafe allocation initialization errors.
+  public static final String DEBUG_ALLOCATOR = "arrow.memory.debug.allocator";
+  public static final int DEBUG_LOG_LENGTH = 6;
+  public static final boolean DEBUG;
```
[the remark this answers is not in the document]
tejassp-db: This code is copied verbatim from another Arrow class `BaseAllocator`.

src/main/java/org/apache/arrow/memory/util/MemoryUtil.java
tejassp-db: This code explicitly removes the print of stack trace to `stderr` to prevent customers from thinking something is broken, since our fix works with the native Arrow being absent as well. The exception is caught and error message is logged in static initiliazer of `ArrowBufferAllocator`.

src/main/java/org/apache/arrow/memory/DatabricksAllocationReservation.java
tejassp-db: The interface that this class implements `AllocationReservation`, is explicitly marked as not thread-safe. So we are following contract. Also this code is generally called from a single thread.

src/main/java/org/apache/arrow/memory/DatabricksArrowBuf.java
tejassp-db: `ByteBuffer.allocate` constructor takes an integer argument. It is an inherent limitation of Java nio ByteBuffer. In JDBC case the maximum allocation will be a chunk size, which is 20MiB as of today. 

Reuse constructor - I will give it more thought and get back.

src/main/java/com/databricks/jdbc/api/impl/arrow/ArrowBufferAllocator.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final boolean canCreateRootAllocator;
+
+  /** Logger instance. */
+  private static final JdbcLogger LOGGER = JdbcLoggerFactory.getLogger(ArrowBufferAllocator.class);
+
+  /* Check if the RootAllocator can be instantiated. */
+  static {
+    RootAllocator rootAllocator = null;
+    try {
+      rootAllocator = new RootAllocator();
```
jayantsing-db: Is instantiating a root allocator object sufficient/robust signal to fallback to custom allocator? For example, for resource/IO wrapper objects like root allocator, Arrow may choose to just create a lightweight object/handle and the off-heap memory is not unsafe-accessed until a memory is explicitly buffered using that root allocator object. So, just creating a root allocator object may succeed but as the arrow reader proceeds to buffer off-heap memory during runtime, then it fails.

Should we do something like this to be deliberate about the fallback logic?
```
Class<?> unsafeClass = Class.forName....
Field f = ....
f.setAccessible(true);
```

src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -0,0 +1,1265 @@
...
+
+  // Initialize this before DEFAULT_CONFIG as DEFAULT_CONFIG will eventually initialize the
+  // allocation manager,
+  // which in turn allocates an ArrowBuf, which requires DEBUG to have been properly initialized
```
jayantsing-db: Since this is already within the patch block, we can remove this potentially confusing comment from the BaseAllocator class.

src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -0,0 +1,1265 @@
...
+ * <p>The management (allocation, deallocation, reference counting etc) for the memory chunk is not
+ * done by ArrowBuf. Default implementation of ReferenceManager, allocation is in {@link
+ * BaseAllocator}, {@link BufferLedger} and {@link AllocationManager}
+ */
+public class ArrowBuf implements AutoCloseable {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ArrowBuf.class);
+
+  // ---- Databricks patch start ----
+  // ---- Copied verbatim from BaseAllocator. We avoid initializing static fields of BaseAllocator
```
jayantsing-db: qq: What was the procedure to determine that "we need to break the static dependency chain to BaseAllocator" because it ends up using unsafe? Is it based on empirical analysis or some deterministic code scan?

The reason I am asking is that there could be a whole bunch of classes loading in this arrow parsing path which might end up using unsafe and ultimately break in exciting new ways.

src/main/java/org/apache/arrow/memory/util/MemoryUtil.java
```diff
@@ -0,0 +1,263 @@
...
+        }
+      } finally {
+        if (address != -1) {
+          UNSAFE.freeMemory(address);
+        }
+      }
+      DIRECT_BUFFER_CONSTRUCTOR = directBufferConstructor;
+    } catch (Throwable e) {
+      // This exception will get swallowed, but it's necessary for the static analysis that ensures
+      // the static fields above get initialized
```
jayantsing-db: nit: Do we need to ensure that this exception is indeed correctly swallowed? If unsure, should we just stop throwing exception in the patch (although i agree with the above philosophy of avoiding as much patch lines as possible)

src/main/java/org/apache/arrow/memory/DatabricksReferenceManager.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private ArrowBuf deriveBuffer(DatabricksArrowBuf srcBuffer) {
+    return deriveBuffer(srcBuffer, 0, srcBuffer.capacity());
+  }
+
+  @Override
+  public ArrowBuf deriveBuffer(ArrowBuf sourceBuffer, long index, long length) {
+    Preconditions.checkArgument(
+        length <= Integer.MAX_VALUE,
```
jayantsing-db: Should this be index + length?

src/main/java/org/apache/arrow/memory/DatabricksReferenceManager.java
```diff
@@ -0,0 +1,131 @@
...
+            + index
+            + " and length="
+            + length
+            + " exceeds source buffer capacity="
+            + sourceBuffer.capacity());
+
+    // Create a new DatabricksArrowBuf sharing the same byte buffer.
+    DatabricksArrowBuf buf = checkBufferType(sourceBuffer);
+    return new DatabricksArrowBuf(
+        this, null, buf.getByteBuffer(), buf.getOffset() + (int) index, length);
```
jayantsing-db: This always sets buffer manager as null. The reallocIfNeeded will always throw error on slices.

```
@Override
  public ArrowBuf reallocIfNeeded(final long size) {
    Preconditions.checkArgument(size >= 0, "reallocation size must be non-negative");
    if (this.capacity() >= size) {
      return this;
    }
    if (bufferManager != null) {
      return bufferManager.replace(this, size);
    } else {
      throw new UnsupportedOperationException(
          "Realloc is only available in the context of operator's UDFs");
    }
  }
  ```

File: src/main/java/org/apache/arrow/memory/DatabricksReferenceManager.java [hunk 14]
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Override
+  public boolean release() {
+    return release(1);
+  }
+
+  @Override
+  public boolean release(int decrement) {
+    return getRefCount() == 0;
```
jayantsing-db: This will always return false. Could this cause any unknown issues? For example, in the chain of calls, when an arrow reader is closed -> vectors are closed -> buffer is released; could this method (always returning false), cause any unintended leaks by short-circuiting any close chain? I understand that this doesn't matter for on-heap arrow bufs.

jayantsing-db:
Quick question (while I am excited to see this go to production) how do we want to handle fragility and maintainability here? My preference would be to put this patch in the arrow-java repo under the Databricks module so we minimize the chances of things breaking or drifting over time.

src/main/java/org/apache/arrow/memory/DatabricksBufferAllocator.java
```diff
@@ -0,0 +1,236 @@
...
+   *
+   * @param name the allocator name
+   * @param parent the parent allocator (null for root)
+   */
+  public DatabricksBufferAllocator(String name, DatabricksBufferAllocator parent) {
+    this.name = name;
+    this.parent = parent;
+
+    // Create an empty buffer with a no-op reference manager
+    this.emptyBuffer = new DatabricksArrowBuf(ReferenceManager.NO_OP, null, 0);
```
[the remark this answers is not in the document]
jayantsing-db: Ah, I got rusty on Java class loading. LLM says that the RootAllocator class will not be loaded until the method is invoked. Currently the method is not invoked anywhere in arrow codebase so the risk is low but I would see it as a latent risk and suggest to get rid of NO_OP reference manager because it directly uses RootAllocator.

src/main/java/org/apache/arrow/memory/DatabricksBufferAllocator.java
```diff
@@ -0,0 +1,236 @@
...
+
+    // Remove from parent's children list
+    if (parent != null) {
+      parent.children.remove(this);
+    }
+  }
+
+  @Override
+  public long getAllocatedMemory() {
+    return 0;
```
jayantsing-db: nit: We have some debug logs in JDBC read path like below:
```
private void logAllocatorStats(String event) {
    long allocatedMemory = rootAllocator.getAllocatedMemory();
    long peakMemory = rootAllocator.getPeakMemoryAllocation();
    long headRoom = rootAllocator.getHeadroom();
    long initReservation = rootAllocator.getInitReservation();

    LOGGER.debug(
        "Chunk allocator stats Log - Event: {}, Chunk Index: {}, Allocated Memory: {}, Peak Memory: {}, Headroom: {}, Init Reservation: {}",
        event,
        chunkIndex,
        allocatedMemory,
        peakMemory,
        headRoom,
        initReservation);
  }
```
rootAllocator.getAllocatedMemory(); this and others would now always return 0 (which is a bit misleading).

src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -0,0 +1,1265 @@
...
+      historicalLog.buildHistory(sb, indent + 1, verbosity.includeStackTraces);
+    }
+  }
+
+  /**
+   * Print detailed information of this buffer into <code>sb</code>.
+   *
+   * <p>Most information will only be present if BaseAllocator.DEBUG is true.
+   */
+  public void print(StringBuilder sb, int indent) {
```
jayantsing-db: nit: If this method is called, BaseAllocator will be loaded in runtime and break because we are still using BaseAllocator's Verbosity.LOG_WITH_STACKTRACE

src/main/java/org/apache/arrow/vector/util/DecimalUtility.java
```diff
@@ -0,0 +1,247 @@
...
+            DatabricksArrowBuf buf = (DatabricksArrowBuf) bytebuf;
+            final int startIdx = index * byteWidth;
+            if (LITTLE_ENDIAN) {
+                buf.setLong(startIdx, value);
+                for (int i = 1; i <= (byteWidth - 8) / 8; i++) {
+                    buf.setLong(startIdx + Long.BYTES * i, padValue);
+                }
+            } else {
+                for (int i = 0; i < (byteWidth - 8) / 8; i++) {
+                    MemoryUtil.putLong(startIdx + Long.BYTES * i, padValue);
```
jayantsing-db: Is this correct?
-  MemoryUtil#putLong expects absolute memory address
- The patched buff iiuc should not concern itself with Memory util at all

src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -0,0 +1,1265 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// -------------------------------------------------------------------------
+// MODIFICATION NOTICE:
+// This file was modified by Databricks, Inc on 16-December-2025.
+// Description of changes:
+//        - Patched ArrowBuf to be non-final and extensible.
+//        - Patched ArrowBuf to remove dependency on BaseAllocator.
```
jayantsing-db: I think this modification notice is incorrect: We are still importing `import org.apache.arrow.memory.BaseAllocator.Verbosity;`

src/main/java/org/apache/arrow/vector/util/DecimalUtility.java
```diff
@@ -0,0 +1,247 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// -------------------------------------------------------------------------
+// MODIFICATION NOTICE:
+// This file was modified by Databricks, Inc on 16-December-2025.
+// Description of changes: Patched method writeLongToArrowBuf to handle DatabricksArrowBuf.
```
jayantsing-db: If i understand the reason of this patch correctly, this is most likely because this utility class uses memoryAddress with MemoryUtil and would have otherwise break when used with patched buff class.

Could there be other classes that are incompatible with patched buff class? I asked LLM to search Arrow java and it found few other classes. The assessment is that right now those classes are not directly used by the read path in this JDBC. Is that a safe assumption? Attaching the summary below:

```
NOT Patched - Vector Classes                                                                                                                                                                                                                                                                                    
  ┌──────────────────────┬───────────────────────────────────────────────────────────────────────┬────────────────────────┐                                                                                                                                                                                             
  │        Class         │              Methods Using memoryAddress() + MemoryUtil               │     Risk for JDBC      │                                                                                                                                                                                             
  ├──────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────┤                                                                                                                                                                                             
  │ DecimalVector        │ setBigEndian(), setSafe(int, long, ArrowBuf, int), setBigEndianSafe() │ Low (write-only)       │                                                                                                                                                                                             
  ├──────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────┤                                                                                                                                                                                             
  │ Decimal256Vector     │ setBigEndian(), setSafe(int, long, ArrowBuf, int), setBigEndianSafe() │ Low (write-only)       │                                                                                                                                                                                             
  ├──────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────┤                                                                                                                                                                                             
  │ BaseFixedWidthVector │ copyFrom(), copyFromSafe()                                            │ Low (not used in JDBC) │                                                                                                                                                                                             
  ├──────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────┤                                                                                                                                                                                             
  │ BitVectorHelper      │ checkAllBitsEqualTo(), concatBits()                                   │ Low (utility methods)  │                                                                                                                                                                                             
  ├──────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────┤                                                                                                                                                                                             
  │ VectorAppender       │ Multiple visit() methods                                              │ Low (not used in JDBC) │                                                                                                                                                                                             
  └──────────────────────┴───────────────────────────────────────────────────────────────────────┴────────────────────────┘                                                                                                                                                                                             
NOT Patched - Hash/Utility Classes                                                                                                                                                                                                                                                                               
  ┌──────────────┬────────────────────────────────────────────┬────────────────────────────────────────────────┐                                                                                                                                                                                                        
  │    Class     │ Methods Using memoryAddress() + MemoryUtil │                 Risk for JDBC                  │                                                                                                                                                                                                        
  ├──────────────┼────────────────────────────────────────────┼────────────────────────────────────────────────┤                                                                                                                                                                                                        
  │ MurmurHasher │ hashCode(ArrowBuf, long, long)             │ Medium - could be used for dictionary encoding │                                                                                                                                                                                                        
  ├──────────────┼────────────────────────────────────────────┼────────────────────────────────────────────────┤                                                                                                                                                                                                        
  │ SimpleHasher │ hashCode(ArrowBuf, long, long)             │ Medium - could be used for dictionary encoding │                                                                                                                                                                                                        
  └──────────────┴────────────────────────────────────────────┴────────────────────────────────────────────────┘ 
```

src/main/java/org/apache/arrow/memory/DatabricksArrowBuf.java
```diff
@@ -0,0 +1,742 @@
...
+  }
+
+  @Override
+  public ByteBuffer nioBuffer(long index, int length) {
+    chk(index, length);
+    ByteBuffer duplicate = byteBuffer.duplicate();
+    duplicate.order(BYTE_ORDER);
+    duplicate.position(offset + (int) index);
+    duplicate.limit(offset + (int) index + length);
+    return duplicate;
```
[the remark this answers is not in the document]
jayantsing-db: ByteBuffer#slice

jayantsing-db (commented):
I completed my review and put some comments. I will take a final look at discussions before stamping. I will check test PRs too.

src/main/java/org/apache/arrow/memory/DatabricksBufferAllocator.java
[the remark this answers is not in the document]
tejassp-db: Fixed - added a Databricks specifc NOOP `ReferenceManager` which returns `DatabricksBufferAllocator`.

src/main/java/org/apache/arrow/memory/DatabricksBufferAllocator.java
```diff
@@ -0,0 +1,236 @@
...
+
+    DatabricksBufferAllocator child = new DatabricksBufferAllocator(name, this);
+
+    children.add(child);
+
+    return child;
+  }
+
+  @Override
+  public void close() {
```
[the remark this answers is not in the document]
tejassp-db: I have this memory test `DatabricksArrowPatchMemoryUsageTest.java`. It is part of PR https://github.com/databricks/databricks-jdbc/pull/1144

src/main/java/org/apache/arrow/memory/ArrowBuf.java
tejassp-db: Fixed.

src/main/java/org/apache/arrow/vector/util/DecimalUtility.java
tejassp-db: LLM findings for MurmurHasher

```
  Read Path Analysis                                                                                                                                                                                                                                      
                  
  The ArrowStreamReader read path follows this sequence:
  1. ArrowStreamReader.loadNextBatch() (vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamReader.java:142-176)
  2. → MessageSerializer.deserializeRecordBatch()
  3. → ArrowReader.loadRecordBatch() or loadDictionary()
  4. → VectorLoader.load() or loadFieldBuffers()

  No hashing operations occur in this path.

  Where Hashing IS Used

  Hashing (including MurmurHasher) is only used in the encoding/write path:

  - DictionaryHashTable uses ArrowBufHasher for the encode() operation in DictionaryEncoder
  - The decode() method does NOT use the hash table - it simply copies dictionary values using index lookups

  MurmurHasher Usage in Practice

  - MurmurHasher is only instantiated in tests (TestBitVector.java, TestArrowBufHasher.java)
  - Production code defaults to SimpleHasher.INSTANCE for DictionaryHashTable operations
  - No production code directly instantiates new MurmurHasher()

  Summary Table
  ┌───────────────────────────────────┬───────────────┬────────────────────────┐
  │             Operation             │ Uses Hashing? │      Hasher Used       │
  ├───────────────────────────────────┼───────────────┼────────────────────────┤
  │ ArrowStreamReader.loadNextBatch() │ ❌ No         │ N/A                    │
  ├───────────────────────────────────┼───────────────┼────────────────────────┤
  │ Dictionary decoding (read)        │ ❌ No         │ N/A                    │
  ├───────────────────────────────────┼───────────────┼────────────────────────┤
  │ Dictionary encoding (write)       │ ✅ Yes        │ SimpleHasher (default) │
  └───────────────────────────────────┴───────────────┴────────────────────────┘
  The separation is intentional - reading just loads buffers and copies data, while encoding requires hash lookups to build dictionary indices.
```

src/main/java/org/apache/arrow/memory/ArrowBuf.java
tejassp-db: fixed.

src/main/java/org/apache/arrow/memory/DatabricksBufferAllocator.java
tejassp-db: Yes, it would be incorrect.

jayantsing-db (approved):

PR #1159, opened 2025-12-26
Fixes multichunk test by only counting the unique urls 

msrathore-db:
## Description

Fixes multichunk test by only counting the unique urls requested that'll ensure that the test is not counting the retries
## Testing

Tested locally

NO_CHANGELOG=true

### DIFF
--- a/development/.release-freeze.json
+++ b/development/.release-freeze.json
@@ -1,4 +1,4 @@
 {
-  "freeze": true,
-  "reason": "3.0.7 release in progress"
+  "freeze": false,
+  "reason": "Release freeze disabled"
 }


samikshya-db:
Is the retry because of incomplete PR #1150  [[code link](https://github.com/databricks/databricks-jdbc/blob/4ed9125c43b54d1c685a17ec541d0922bea57b87/src/main/java/com/databricks/jdbc/api/impl/arrow/ChunkLinkDownloadService.java#L421)]

should we try fixing that first?

msrathore-db:
> Is the retry because of incomplete PR #1150 [[code link](https://github.com/databricks/databricks-jdbc/blob/4ed9125c43b54d1c685a17ec541d0922bea57b87/src/main/java/com/databricks/jdbc/api/impl/arrow/ChunkLinkDownloadService.java#L421)]
> 
> should we try fixing that first?

Yeah makes sense. That would be a better fix. Updated. Thanks!!

PR #1175, opened 2026-01-09
Add UserAgent for connector service calls

gopalldb:
## Problem

  Custom user agent from useragententry parameter wasn't included in connector service HTTP requests for feature flag retrieval.

 ## Root Cause

  Method execution order issue in UserAgentManager.setUserAgent() - custom user agent was set AFTER the connector service request was made.

 ## Solution

  Reordered the method to set custom user agent before calling getClientUserAgent() (which triggers feature flag fetch).

## Testing

  - Added testCustomUserAgentIncludedBeforeClientTypeEvaluation() test
  
  NO_CHANGELOG=true

### DIFF
--- a/src/main/java/com/databricks/jdbc/common/DatabricksJdbcConstants.java
+++ b/src/main/java/com/databricks/jdbc/common/DatabricksJdbcConstants.java
@@ -108,6 +108,7 @@ public final class DatabricksJdbcConstants {
   public static final String QUERY_EXECUTION_TIMEOUT_SQLSTATE = "57KD0";
   public static final int TEMPORARY_REDIRECT_STATUS_CODE = 307;
   public static final String REDACTED_TOKEN = "****";
+  public static final String QUERY_TAGS = "query_tags";
   public static final int MAX_DEFAULT_STRING_COLUMN_LENGTH = 32767;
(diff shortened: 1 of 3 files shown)


src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
```diff
@@ -17,33 +17,96 @@ public class UserAgentManager {
...
+  private static String[] parseCustomerUserAgent(String customerUserAgent) {
+    try {
+      String decodedUA = URLDecoder.decode(customerUserAgent, StandardCharsets.UTF_8);
+      int i = decodedUA.indexOf('/');
+      String customerName = (i < 0) ? decodedUA : decodedUA.substring(0, i);
+      String customerVersion = (i < 0) ? VERSION_FILLER : decodedUA.substring(i + 1);
+      return new String[] {customerName, customerVersion};
+    } catch (Exception e) {
+      LOGGER.debug("Failed to parse customer userAgent entry {}, Error {}", customerUserAgent, e);
+      return null;
```
jprakash-db: Instead of null, can we give a generic name. So we can later using telemetry identity how many requests are coming from incorrect User Agent arguments

jprakash-db (approved):
Overall LGTM. Some minor comments

Issue #1178, opened 2026-01-12
[BUG]


ovna01:
### Issue with metadata.getTypeInfo ResultSet
**Describe the bug**
The ResultSet from metadata call getTypeInfo and getClientInfoProperties can only be used once per session

**To Reproduce**
For getTypeInfo after connect:
DatabaseMetaData dm = connection.getMetaData();
ResultSet rs = dm.getTypeInfo();
while(rs.next()) {} // Loop to end of rs
// Second call returns the same ResultSet instance that is closed and cannot be used
ResultSet rs = dm.getTypeInfo();

**Expected behavior**
Each call to getTypeInfo() (and getClientInfoProperties) should return a new (or usable) ResultSet.

**Client Environment (please complete the following information):**
 - OS: Mac OS
 - Java 21
 - Java vendor OpenJDK
 - Driver Version 3.01 - 3.0.7
 - BI Tool - DbVisualizer
 - BI Tool version  25.3.1 

**Additional context**
This was detected in DbVisualizer by the developers and we are working on a work around. 
In the tool we call getTypeInfo in several occasions and as the second call does not work some major 
features in DbVisualizer does not work e.g. Create Table.

msrathore-db:
The fix was merged and would be a part of the next release

PR #1187, opened 2026-01-26
Added support for getClientInfoProperties and getTypeInfo to return a new resultset

msrathore-db:
Added support for getClientInfoProperties and getTypeInfo to return a new resultset and not return the same resultset matching the JDBC spec

## Description

Added support for getClientInfoProperties and getTypeInfo to return a new resultset and not return the same resultset matching the JDBC spec
## Testing

Added tests

Fixes: #1178


src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
```diff
@@ -1529,7 +1528,7 @@ public boolean autoCommitFailureClosesAllResultSets() throws SQLException {
   public ResultSet getClientInfoProperties() throws SQLException {
     LOGGER.debug("public ResultSet getClientInfoProperties()");
     throwExceptionIfConnectionIsClosed();
-    return CLIENT_INFO_PROPERTIES_RESULT;
+    return metadataResultSetBuilder.getClientInfoPropertiesResult();
```
vikrantpuppala: Please check if there are any other functions with the similar issue

msrathore-db: Couldn't find

PR #1189, opened 2026-01-29
Add NonRowcountQueryPrefixes configuration parameter

gopalldb:
## Description
Implements NonRowcountQueryPrefixes flag to match exiting JDBC driver behavior. This allows users to specify comma-separated query prefixes (like INSERT, UPDATE, DELETE) that should return result sets instead of row counts.

Changes:
- Added NON_ROWCOUNT_QUERY_PREFIXES parameter to DatabricksJdbcUrlParams
- Added getNonRowcountQueryPrefixes() method to connection context interface and implementation
- Updated shouldReturnResultSet() logic to check configured prefixes before SQL patterns
- Added 11 comprehensive unit tests covering various scenarios
- Updated NEXT_CHANGELOG.md with feature description

Usage:
NonRowcountQueryPrefixes=INSERT,UPDATE,DELETE,MERGE

## Testing
Tests: All 75 tests pass (64 existing + 11 new)

### DIFF
--- a/src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
+++ b/src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
@@ -1132,4 +1132,65 @@ public class DatabricksStatementTest {
     String query = "SELECT * FROM table";
     assertTrue(DatabricksStatement.shouldReturnResultSet(query, prefixes));
   }
+
+  @Test
+  public void testExecuteInsertWithNonRowcountQueryPrefixes() throws Exception {
+    // Create connection with NonRowcountQueryPrefixes=INSERT
+    String jdbcUrlWithPrefix = JDBC_URL + "NonRowcountQueryPrefixes=INSERT";
+    IDatabricksConnectionContext connectionContext =
+        DatabricksConnectionContext.parse(jdbcUrlWithPrefix, new Properties());
+    DatabricksConnection connection = new DatabricksConnection(connectionContext, client);
+    DatabricksStatement statement = new DatabricksStatement(connection);
+
+    String insertStatement = "INSERT INTO table VALUES (1, 2, 3)";
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -681,8 +682,23 @@ static String trimCommentsAndWhitespaces(String query) {
...
     String trimmedQuery = trimCommentsAndWhitespaces(query);
 
+    // Check configured non-rowcount prefixes first
+    if (nonRowcountQueryPrefixes != null && !nonRowcountQueryPrefixes.isEmpty()) {
+      String upperQuery = trimmedQuery.toUpperCase();
+      for (String prefix : nonRowcountQueryPrefixes) {
+        if (upperQuery.startsWith(prefix)) {
+          return true;
+        }
+      }
```
vikrantpuppala: can we use a simple lambda to write this

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
```diff
@@ -1162,6 +1162,19 @@ public boolean isBatchedInsertsEnabled() {
     return getParameter(DatabricksJdbcUrlParams.ENABLE_BATCHED_INSERTS).equals("1");
   }
 
+  @Override
+  public List<String> getNonRowcountQueryPrefixes() {
+    String prefixesStr = getParameter(DatabricksJdbcUrlParams.NON_ROWCOUNT_QUERY_PREFIXES);
+    if (prefixesStr == null || prefixesStr.trim().isEmpty()) {
+      return Collections.emptyList();
+    }
```
vikrantpuppala: this will actually never be null because default value is empty? and empty condition will be handled by the lambda below

PR #1202, opened 2026-02-06
Fix: try cancelling statement in all cases before throing an error

zashroof:
## Description
If thread is inturrpted or a statment timesout, an exception is thrown while sometimes the statement continues to execute on the server. Also we see lots of exceptions that cancelling a statement failed (404).

Adding a tryCancelStatement to always try to cancel the statement on the server even if the chances the cancelStatement succeeds are slim. And not throwing an exception if canceling a statement fails.

## Testing
Added tests

Cursor helped me create this PR. This is my first PR in this repo as well so probably I got a few things wrong.

### DIFF
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
@@ -254,6 +254,8 @@ public class DatabricksSdkClient implements IDatabricksClient {
         try {
           Thread.sleep(connectionContext.getAsyncExecPollInterval());
         } catch (InterruptedException e) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -254,6 +254,8 @@ public DatabricksResultSet executeStatement(
         try {
           Thread.sleep(connectionContext.getAsyncExecPollInterval());
         } catch (InterruptedException e) {
+          // Cancel the statement on the server before throwing - it may still be running
```
gopalldb: We have cancel being fired from checkTimeout above also. We can keep a check that if cancel was already attempted, don't do cancel again.

PR #1206, opened 2026-02-11
[PECOBLR-1943] Fix rollback in auto-commit mode and handle numeric autocommit responses

vikrantpuppala:
## Summary
- **rollback() now throws** `DatabricksTransactionException` when called while the connection is in auto-commit mode (no active transaction), aligning behavior with the JDBC spec
- **fetchAutoCommitStateFromServer()** now accepts both `"1"`/`"0"` and `"true"`/`"false"` responses from the `SET AUTOCOMMIT` query, since different server implementations return different formats
- Updated E2E and unit tests to reflect the corrected behavior

## Context
1. `connection.rollback()` in auto-commit mode should throw an exception, not silently succeed
2. `SET AUTOCOMMIT` query returns `"1"`/`"0"` rather than `"true"`/`"false"` depending on the server

## Test plan
- [x] Unit tests pass: `DatabricksConnectionTest` (42/42 pass)
- [ ] E2E tests: `TransactionTests` and `ExplicitTransactionStatementTests` against staging warehouse

### DIFF
--- a/.github/workflows/prCheckJDK8.yml
+++ b/.github/workflows/prCheckJDK8.yml
@@ -4,6 +4,21 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
     branches: [jdk-8]
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to checkout'
+        required: false
+        default: 'jdk-8'
+      repository:
+        description: 'Repository to checkout (e.g., user/repo)'
+        required: false
+        default: 'databricks/databricks-jdbc'
+      reason:
+        description: 'Reason for manual run'
+        required: false
+        default: 'Manual health check for JDK-8'
+
 
 jobs:
   jdk8-test:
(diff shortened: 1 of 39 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
```diff
@@ -252,13 +252,13 @@ private boolean fetchAutoCommitStateFromServer() throws SQLException {
...
       if (rs.next()) {
-        // The result should contain the value = "true" or "false"
+        // The result may contain "true"/"false" or "1"/"0" depending on the server
         String value = rs.getString(1); // Column 1: value
 
         LOGGER.debug(
             "Fetched autoCommit state from server: value={}. Updating session cache.", value);
 
-        boolean autoCommitState = "true".equalsIgnoreCase(value);
+        boolean autoCommitState = "true".equalsIgnoreCase(value) || "1".equals(value);
```
gopalldb: it means that even values like "yes" or "enabled" would result in false. I think, can we explicitly throw error in those cases instead of blindly treating those as false? Not blocker for this PR though

PR #1230, opened 2026-02-24
[PECOBLR-1928] Add AI coding agent detection to User-Agent header

vikrantpuppala:
## Summary
- Adds `AgentDetector` utility class that detects 7 AI coding agents (Claude Code, Cursor, Gemini CLI, Cline, Codex, OpenCode, Antigravity) by checking well-known environment variables they set in spawned shell processes
- Integrates detection into `UserAgentManager` to append `agent/<product>` to the User-Agent header on both SDK-based and connector-service UA paths
- Uses exactly-one detection rule: if zero or multiple agent env vars are set, no agent is attributed (avoids ambiguity)

## Approach
Mirrors the implementation in [databricks/cli#4287](https://github.com/databricks/cli/pull/4287) and aligns with the latest agent list in [`libs/agent/agent.go`](https://github.com/databricks/cli/blob/main/libs/agent/agent.go#L35).

| Agent | Product String | Environment Variable |
|-------|---------------|---------------------|
| Google Antigravity | `antigravity` | `ANTIGRAVITY_AGENT` |
| Claude Code | `claude-code` | `CLAUDECODE` |
| Cline | `cline` | `CLINE_ACTIVE` |
| OpenAI Codex | `codex` | `CODEX_CI` |
| Cursor | `cursor` | `CURSOR_AGENT` |
| Gemini CLI | `gemini-cli` | `GEMINI_CLI` |
| OpenCode | `opencode` | `OPENCODE` |

Adding a new agent requires only a new constant and a new entry in the `KNOWN_AGENTS` array.

## Changes
- **New**: `AgentDetector.java` — environment-variable-based agent detection with injectable env lookup for testability
- **Modified**: `UserAgentManager.java` — calls `AgentDetector.detect()` in both `setUserAgent()` and `buildUserAgentForConnectorService()`
- **New**: `AgentDetectorTest.java` — 12 test cases covering single agent, no agent, multiple agents, empty values, null values, and full coverage of all known agents

## Test plan
- [x] `AgentDetectorTest` — 12 tests pass (single agent detection for all 7 agents, no agent, multiple agents, empty/null values)
- [x] `UserAgentManagerTest` — all 7 existing tests continue to pass
- [x] Manual: verify User-Agent header contains `agent/claude-code` when run from Claude Code

### DIFF
--- a/src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
+++ b/src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
@@ -1,0 +1,79 @@
+package com.databricks.jdbc.common.util;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * Detects whether the JDBC driver is being invoked by an AI coding agent by checking for well-known
+ * environment variables that agents set in their spawned shell processes.
+ *
+ * <p>Detection only succeeds when exactly one agent environment variable is present, to avoid
+ * ambiguous attribution when multiple agent environments overlap.
+ *
+ * <p>Adding a new agent requires only a new constant and a new entry in {@link #KNOWN_AGENTS}.
+ *
+ * <p>References for each environment variable:
+ *
+ * <ul>
+ *   <li>ANTIGRAVITY_AGENT: Closed source. Google Antigravity sets this variable.
+ *   <li>CLAUDECODE: https://github.com/anthropics/claude-code (sets CLAUDECODE=1)
+ *   <li>CLINE_ACTIVE: https://github.com/cline/cline (shipped in v3.24.0)
+ *   <li>CODEX_CI: https://github.com/openai/codex (part of UNIFIED_EXEC_ENV array in codex-rs)
+ *   <li>CURSOR_AGENT: Closed source. Referenced in a gist by johnlindquist.
+ *   <li>GEMINI_CLI: https://google-gemini.github.io/gemini-cli/docs/tools/shell.html (sets
+ *       GEMINI_CLI=1)
+ *   <li>OPENCODE: https://github.com/opencode-ai/opencode (sets OPENCODE=1)
+ * </ul>
+ */
+public class AgentDetector {
+
+  public static final String ANTIGRAVITY = "antigravity";
+  public static final String CLAUDE_CODE = "claude-code";
+  public static final String CLINE = "cline";
+  public static final String CODEX = "codex";
+  public static final String CURSOR = "cursor";
+  public static final String GEMINI_CLI = "gemini-cli";
+  public static final String OPEN_CODE = "opencode";
+
+  static final String[][] KNOWN_AGENTS = {
+    {"ANTIGRAVITY_AGENT", ANTIGRAVITY},
+    {"CLAUDECODE", CLAUDE_CODE},
+    {"CLINE_ACTIVE", CLINE},
+    {"CODEX_CI", CODEX},
+    {"CURSOR_AGENT", CURSOR},
+    {"GEMINI_CLI", GEMINI_CLI},
+    {"OPENCODE", OPEN_CODE},
+  };
+
+  /**
+   * Detects which AI coding agent (if any) is driving the current process.
+   *
+   * @return the agent product string if exactly one agent is detected, or empty otherwise
+   */
+  public static Optional<String> detect() {
+    return detect(System::getenv);
+  }
+
+  /**
+   * Detects which AI coding agent (if any) is present, using the provided function to look up
+   * environment variables. This overload exists for testability.
+   *
+   * @param envLookup function that returns the value of an environment variable, or null if unset
+   * @return the agent product string if exactly one agent is detected, or empty otherwise
+   */
+  static Optional<String> detect(Function<String, String> envLookup) {
+    List<String> detected = new ArrayList<>();
+    for (String[] entry : KNOWN_AGENTS) {
+      String value = envLookup.apply(entry[0]);
+      if (value != null && !value.isEmpty()) {
+        detected.add(entry[1]);
+      }
+    }
+    if (detected.size() == 1) {
+      return Optional.of(detected.get(0));
+    }
+    return Optional.empty();
+  }
+}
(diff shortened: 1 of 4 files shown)


simonfaltum (approved):
Looks right

src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
```diff
@@ -62,6 +63,12 @@ public static void setUserAgent(IDatabricksConnectionContext connectionContext)
         }
       }
     }
+
+    // Detect AI coding agent and append to user agent
+    String agentProduct = AgentDetector.detect();
```
samikshya-db: I hope this does not break the user-agent parsing functionality in https://github.com/databricks-eng/universe/commit/d74e4f8861d34f726b035e15653110ba1879583c

As long as that is the case, I am good

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
```diff
@@ -0,0 +1,78 @@
...
+    {"GEMINI_CLI", GEMINI_CLI},
+    {"OPENCODE", OPEN_CODE},
+  };
+
+  /**
+   * Detects which AI coding agent (if any) is driving the current process.
+   *
+   * @return the agent product string if exactly one agent is detected, or an empty string otherwise
+   */
+  public static String detect() {
```
tejassp-db: General convention is to use `null` or better yet `java.util.Optional` to signify a missing value. Empty string can be misleading.

nikhilsuri-db (approved):

gopalldb (commented):
As discussed offline, lets not merge this before server side changes

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
```diff
@@ -0,0 +1,78 @@
...
+    for (String[] entry : KNOWN_AGENTS) {
+      String value = envLookup.apply(entry[0]);
+      if (value != null && !value.isEmpty()) {
+        detected.add(entry[1]);
+      }
+    }
+    if (detected.size() == 1) {
+      return detected.get(0);
+    }
+    return "";
```
gopalldb: why not null? will that break?

gopalldb (approved):
Verify once with server side reporting, a minor comment on empty string

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
```diff
@@ -0,0 +1,78 @@
...
+   * Detects which AI coding agent (if any) is present, using the provided function to look up
+   * environment variables. This overload exists for testability.
+   *
+   * @param envLookup function that returns the value of an environment variable, or null if unset
+   * @return the agent product string if exactly one agent is detected, or an empty string otherwise
+   */
+  static String detect(Function<String, String> envLookup) {
+    List<String> detected = new ArrayList<>();
+    for (String[] entry : KNOWN_AGENTS) {
+      String value = envLookup.apply(entry[0]);
```
gopalldb, line 69: do we need to do case insensitive check?

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
```diff
@@ -0,0 +1,78 @@
...
+  public static final String ANTIGRAVITY = "antigravity";
+  public static final String CLAUDE_CODE = "claude-code";
+  public static final String CLINE = "cline";
+  public static final String CODEX = "codex";
+  public static final String CURSOR = "cursor";
+  public static final String GEMINI_CLI = "gemini-cli";
+  public static final String OPEN_CODE = "opencode";
+
+  static final String[][] KNOWN_AGENTS = {
+    {"ANTIGRAVITY_AGENT", ANTIGRAVITY},
```
gopalldb, line 41: Can we just use enum here?

enum Agent {
  CLAUDECODE("claude-code"),
  CLINE_ACTIVE("cline"), 

  ...
}

src/main/java/com/databricks/jdbc/common/util/UserAgentManager.java
vikrantpuppala: Addressed — the server-side UA parsers have been updated to handle the agent/ suffix: https://github.com/databricks-eng/universe/pull/1694145 (merged and deployed to dev). Verified e2e that query history shows correct driver name and BI tool fields.

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
vikrantpuppala: Good call — updated to return `Optional<String>` in the latest push.

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public static final String ANTIGRAVITY = "antigravity";
+  public static final String CLAUDE_CODE = "claude-code";
+  public static final String CLINE = "cline";
+  public static final String CODEX = "codex";
+  public static final String CURSOR = "cursor";
+  public static final String GEMINI_CLI = "gemini-cli";
+  public static final String OPEN_CODE = "opencode";
+
+  static final String[][] KNOWN_AGENTS = {
```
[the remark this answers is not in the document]
vikrantpuppala: Yes — the env var name differs from the product name (e.g. `CLAUDECODE` → `claude-code`, `CLINE_ACTIVE` → `cline`), so we need the mapping. The product names are what get sent in the UA header and need to match the server-side convention.

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
vikrantpuppala: Updated to `Optional<String>` per tejassp's suggestion — `detect()` now returns `Optional.empty()` instead of `""`. Callers use `ifPresent()`.

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
vikrantpuppala: No — environment variable names are case-sensitive on Linux, and we define the exact names to check. The CLI reference implementation also does exact-match lookups.

src/main/java/com/databricks/jdbc/common/util/AgentDetector.java
vikrantpuppala: Considered this but kept the data-driven list for consistency — the same pattern (list of tuples/pairs) is used across all 4 driver implementations (Python, Go, Node.js, Java) and the CLI reference. An enum would work but doesn't reduce complexity since each entry still needs two strings (env var name + product name).

sreekanth-db (approved):

PR #1238, opened 2026-02-27
[PECOBLR-1746] Implementing support for listing procedures

msrathore-db:
## Description

Implements `getProcedures` and `getProcedureColumns` in `DatabricksDatabaseMetaData` by querying `information_schema.routines` and `information_schema.parameters` via SQL.

Unlike other metadata operations that use `SHOW` commands or Thrift RPCs, these use direct SQL SELECT queries against `information_schema` views. This works for both Thrift and SEA transports.

**getProcedures:**
- Queries `information_schema.routines` filtered by `routine_type = 'PROCEDURE'`
- Returns 9-column JDBC-spec result set (PROCEDURE_CAT, PROCEDURE_SCHEM, PROCEDURE_NAME, reserved x3, REMARKS, PROCEDURE_TYPE, SPECIFIC_NAME)
- PROCEDURE_TYPE is always `procedureNoResult` (1)

**getProcedureColumns:**
- Queries `information_schema.parameters` JOINed with `information_schema.routines` to filter for procedures
- Returns 20-column JDBC-spec result set with parameter metadata
- Maps `parameter_mode` (IN/OUT/INOUT) to JDBC COLUMN_TYPE constants (1/4/2)
- Maps Databricks type names to `java.sql.Types` codes via existing `getCode()`

**Catalog resolution:**
- NULL catalog → queries `system.information_schema.routines` (cross-catalog)
- Specific catalog → queries `<catalog>.information_schema.routines`
- Empty string → returns empty result set

**Shared SQL builders** in `CommandConstants` eliminate duplication between SDK and Thrift clients.

## Testing

**Unit tests** (`DatabricksMetadataSdkClientTest`):
- 4 parameterized tests for `listProcedures` SQL generation (catalog+schema+name, null schema, null name, null catalog)
- 3 parameterized tests for `listProcedureColumns` SQL generation (all filters, partial filters, all nulls)

**Integration test** (`MetadataIntegrationTests#testGetProceduresAndProcedureColumns`):
- Creates a procedure `jdbc_test_compute_area(x DOUBLE, y DOUBLE, OUT area DOUBLE)` with COMMENT
- Verifies `getProcedures` returns correct name, schema, catalog, remarks, type
- Verifies `getProcedureColumns` returns 3 params with correct COLUMN_TYPE (IN=1, OUT=4), DATA_TYPE (DOUBLE=8), ordinal positions
- Tests column name filtering
- Cleans up procedure after test
- WireMock stubs recorded for REPLAY mode

**Existing tests:** All 248 `DatabricksDatabaseMetaDataTest` + 51 `DatabricksMetadataSdkClientTest` pass.

- `information_schema.parameters` is used (not `routine_columns`) because `routine_columns` contains table-valued function output columns, not procedure parameters.
- NULLABLE is always `procedureNullableUnknown` (2) since the server does not track parameter nullability.
- When catalog is NULL, `system.information_schema` is queried which requires system table access permissions. If the user lacks this permission, the driver returns an empty result set. This will be addressed server-side in a future release.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Added `DatabaseMetaData.getProcedures()` and `DatabaseMetaData.getProcedureColumns()` to discover stored procedures and their parameters. Queries `information_schema.routines` and `information_schema.parameters` using parameterized SQL for both SEA and Thrift transports.
 - Added connection property `OAuthWebServerTimeout` to configure the OAuth browser authentication timeout for U2M (user-to-machine) flows, and also updated hardcoded 1-hour timeout to default 120 seconds timeout.
 - Added connection property `UseQueryForMetadata` to use SQL SHOW commands instead of Thrift RPCs for metadata operations (getCatalogs, getSchemas, getTables, getColumns, getFunctions). This fixes incorrect wildcard matching where `_` was treated as a single-character wildcard in Thrift metadata pattern filters.
 - Added connection property `TreatMetadataCatalogNameAsPattern` to control whether catalog names are treated as patterns in Thrift metadata RPCs. When disabled (default), unescaped `_` in catalog names is escaped to prevent single-character wildcard matching. This aligns with JDBC spec which treats catalogName as identifier and not pattern.


src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1125,6 +1163,118 @@ private List<List<Object>> getRowsForFunctions(
...
+        return (short) procedureColumnUnknown;
+    }
+  }
+
+  private static String getStringOrNull(DatabricksResultSet resultSet, String columnName)
+      throws SQLException {
+    try {
+      Object val = resultSet.getObject(columnName);
+      return val != null ? val.toString() : null;
+    } catch (SQLException e) {
```
gopalldb: add logging

src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
```diff
@@ -26,4 +28,70 @@ public class CommandConstants {
...
+
+    StringBuilder sql = new StringBuilder();
+    sql.append("SELECT ").append(ROUTINES_SELECT_COLUMNS);
+    sql.append(" FROM ").append(routinesTable);
+    sql.append(" WHERE ").append(PROCEDURE_TYPE_FILTER);
+    if (schemaPattern != null) {
+      sql.append(" AND routine_schema LIKE '").append(schemaPattern).append("'");
+    }
+    if (procedureNamePattern != null) {
+      sql.append(" AND routine_name LIKE '").append(procedureNamePattern).append("'");
```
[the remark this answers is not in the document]
msrathore-db: Done. Now using `?` placeholders with `ImmutableSqlParameter` for server-side binding.

src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
```diff
@@ -26,4 +28,70 @@ public class CommandConstants {
...
+    if (columnNamePattern != null) {
+      sql.append(" AND p.parameter_name LIKE '").append(columnNamePattern).append("'");
+    }
+    sql.append(
+        " ORDER BY p.specific_catalog, p.specific_schema, p.specific_name, p.ordinal_position");
+    return sql.toString();
+  }
+
+  private static String getCatalogPrefix(String catalog) {
+    return (catalog == null) ? "system" : "`" + catalog + "`";
```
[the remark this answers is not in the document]
msrathore-db: Catalog names cannot contain backticks, so the user input would be invalid in that case. No escaping needed.

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1125,6 +1163,118 @@ private List<List<Object>> getRowsForFunctions(
...
+      row.add(
+          dataType != null
+              ? getCode(stripBaseTypeName(dataType.toUpperCase()))
+              : null); // DATA_TYPE
+      row.add(dataType != null ? dataType.toUpperCase() : null); // TYPE_NAME
+      Integer numericPrecision = getIntOrNull(resultSet, "numeric_precision");
+      Integer charMaxLength = getIntOrNull(resultSet, "character_maximum_length");
+      Integer charOctetLength = getIntOrNull(resultSet, "character_octet_length");
+      row.add(numericPrecision != null ? numericPrecision : charMaxLength); // PRECISION
+      row.add(charOctetLength != null ? charOctetLength : numericPrecision); // LENGTH
```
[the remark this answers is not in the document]
msrathore-db: Correct — if both `numericPrecision` and `charMaxLength` are null, COLUMN_SIZE returns null, which is the expected behavior per JDBC spec for types where precision is not applicable.

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1125,6 +1163,118 @@ private List<List<Object>> getRowsForFunctions(
...
+      row.add(mapParameterModeToColumnType(parameterMode, isResult)); // COLUMN_TYPE
+      row.add(
+          dataType != null
+              ? getCode(stripBaseTypeName(dataType.toUpperCase()))
+              : null); // DATA_TYPE
+      row.add(dataType != null ? dataType.toUpperCase() : null); // TYPE_NAME
+      Integer numericPrecision = getIntOrNull(resultSet, "numeric_precision");
+      Integer charMaxLength = getIntOrNull(resultSet, "character_maximum_length");
+      Integer charOctetLength = getIntOrNull(resultSet, "character_octet_length");
+      row.add(numericPrecision != null ? numericPrecision : charMaxLength); // PRECISION
```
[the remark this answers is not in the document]
msrathore-db: Same as above — null fallback is intentional. Returns null when neither numeric precision nor character length applies to the type.

src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
```diff
@@ -26,4 +32,107 @@ public class CommandConstants {
...
+      String procedureNamePattern,
+      String columnNamePattern,
+      Map<Integer, ImmutableSqlParameter> params) {
+    String catalogPrefix = getCatalogPrefix(catalog);
+    String parametersTable = catalogPrefix + "." + INFORMATION_SCHEMA_PARAMETERS + " p";
+    String routinesTable = catalogPrefix + "." + INFORMATION_SCHEMA_ROUTINES + " r";
+    int paramIndex = 1;
+
+    StringBuilder sql = new StringBuilder();
+    sql.append("SELECT ").append(PARAMETERS_SELECT_COLUMNS);
```
gopalldb: can we use utility like mybatis-3 for creating such SQL strings?

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1129,6 +1167,127 @@ private List<List<Object>> getRowsForFunctions(
     return rows;
   }
 
+  private List<List<Object>> getRowsForProcedures(DatabricksResultSet resultSet)
+      throws SQLException {
+    LOGGER.debug("Building rows for getProcedures result set");
+    List<List<Object>> rows = new ArrayList<>();
+    while (resultSet.next()) {
+      List<Object> row = new ArrayList<>();
+      row.add(getStringOrNull(resultSet, "routine_catalog")); // PROCEDURE_CAT
```
gopalldb: declare as constants, all of hard coded values

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1129,6 +1167,127 @@ private List<List<Object>> getRowsForFunctions(
...
+      throws SQLException {
+    LOGGER.debug("Building rows for getProcedureColumns result set");
+    List<List<Object>> rows = new ArrayList<>();
+    while (resultSet.next()) {
+      String dataType = getStringOrNull(resultSet, "data_type");
+      String parameterMode = getStringOrNull(resultSet, "parameter_mode");
+      String isResult = getStringOrNull(resultSet, "is_result");
+
+      List<Object> row = new ArrayList<>();
+      row.add(getStringOrNull(resultSet, "specific_catalog")); // PROCEDURE_CAT
```
gopalldb: is null value allowed for catalog and schema?

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1129,6 +1167,127 @@ private List<List<Object>> getRowsForFunctions(
...
+      row.add(numericPrecision != null ? numericPrecision : charMaxLength); // COLUMN_SIZE
+      row.add(charOctetLength); // BUFFER_LENGTH
+      row.add(getShortOrNull(resultSet, "numeric_scale")); // SCALE
+      row.add(getShortOrNull(resultSet, "numeric_precision_radix")); // RADIX
+      row.add((short) procedureNullableUnknown); // NULLABLE
+      row.add(getStringOrNull(resultSet, "comment")); // REMARKS
+      row.add(getStringOrNull(resultSet, "parameter_default")); // COLUMN_DEF
+      row.add(null); // SQL_DATA_TYPE (reserved)
+      row.add(null); // SQL_DATETIME_SUB (reserved)
+      row.add(getIntOrNull(resultSet, "character_octet_length")); // CHAR_OCTET_LENGTH
```
gopalldb: is this same as buffer_length? and you are reading it again

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
```diff
@@ -1129,6 +1167,127 @@ private List<List<Object>> getRowsForFunctions(
...
+    }
+    return rows;
+  }
+
+  private List<List<Object>> getRowsForProcedureColumns(DatabricksResultSet resultSet)
+      throws SQLException {
+    LOGGER.debug("Building rows for getProcedureColumns result set");
+    List<List<Object>> rows = new ArrayList<>();
+    while (resultSet.next()) {
+      String dataType = getStringOrNull(resultSet, "data_type");
```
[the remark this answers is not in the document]
msrathore-db: Done — see above.

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
msrathore-db: Yes, per the JDBC spec both PROCEDURE_CAT and PROCEDURE_SCHEM are nullable. From the spec: "PROCEDURE_CAT String => procedure catalog (may be null)" and "PROCEDURE_SCHEM String => procedure schema (may be null)".

src/main/java/com/databricks/jdbc/dbclient/impl/common/MetadataResultSetBuilder.java
msrathore-db: Good catch. Fixed — now reusing the charOctetLength variable for both BUFFER_LENGTH (col 9) and CHAR_OCTET_LENGTH (col 17) instead of calling getIntOrNull twice.

PR #1243, opened 2026-03-02
[PECOBLR-1121] Arrow patch to circumvent Arrow issues with JDk 16+

tejassp-db:
Databricks server shares query results in Arrow format for easy cross language functionality. The JDBC driver experiences compatibility issues with JDK 16 and later versions when processing Arrow results. 

This problem arises from stricter encapsulation of internal APIs in newer Java versions, which affects the driver's use of the Apache Arrow result format consumption with the Apache Arrow library. The JDBC driver is used in partner solutions, where they do not have control of the runtime environment, and the workaround of setting JVM arguments is not feasible. 

This PR patches some of the Arrow code to provide alternative JVM Heap based byte allocators that do not use native `MemoryUtil` based direct reads from off-heap memory. This implementation uses the native Arrow code path if feasible, else falls back to the patched code.

All the code has been tested for read compatibility with all Arrow types, latency benchmarks have been tested, and automated tests have been added as well.

During the course of this change it became necessary to also convert the project into a multi-module maven project

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Updated
 - Fat jar now routes SDK and Apache HTTP client logs through Java Util Logging (JUL), removing the need for external logging libraries.
+- PECOBLR-1121 Arrow patch to circumvent Arrow issues with JDK 16+.
 
 ### Fixed
 - Fixed `rollback()` to throw `SQLException` when called in auto-commit mode (no active transaction), aligning with JDBC spec. Previously it silently sent a ROLLBACK command to the server.


tejassp-db:
Current github actions wont pass, because the current github workflows  are setup for a single module maven project. I have a separate branch to enable these test runs and I have run it from there.

- https://github.com/databricks/databricks-jdbc/actions/runs/22570820195
- https://github.com/databricks/databricks-jdbc/actions/runs/22570826007
- https://github.com/databricks/databricks-jdbc/actions/runs/22570816060
- https://github.com/databricks/databricks-jdbc/actions/runs/22570810937

PR #1255, opened 2026-03-06
Fix the JDK8 github action + make local build developer friendly + fix coverage

samikshya-db:
## Description
1. The JDK8 GH action is broken, this PR fixes it. 
2. This PR also adds a maven profile that does compile, run unit tests, and package and nothing else (similar to our old `mvn clean test`)
    1. What -Plocal skips:
      - NVD/dependency-check scan
      - Arrow patch tests (org.apache.arrow.memory.*)
      - Integration & fake service tests
      - Proxy, SSL, and logging tests

## Testing
- Run `mvn -pl jdbc-core -Plocal clean test`
- For packaging, `mvn -pl jdbc-core,assembly-uber install -Plocal`

Note : arrow specific tests are not relevant in jdk8 branch as it is specific for JDK16+

NO_CHANGELOG=true

### DIFF
--- a/jdbc-core/pom.xml
+++ b/jdbc-core/pom.xml
@@ -415,10 +415,15 @@
             <exclude>**/model/**</exclude>
             <exclude>**/thrift/generated/**</exclude>
 
-            <!-- Exclude patched Arrow classes from coverage -->
+            <!-- Exclude patched Arrow classes from coverage. All Arrow tests are tagged with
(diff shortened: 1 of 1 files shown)


samikshya-db:
@tejassp-db , For your comment
`If it works here then there is no need to add a profile for local`
No it does not work on local if the repo secrets are not exported locally. I do not want to run e2e locally that is. Moreover, the NVD checks take a long time if cache-key is not present. 

 In short - the local profile is just a convenience shortcut for developers here

Issue #1253, opened 2026-03-05
[BUG] jdk-8 branch has java 9 references


Ankitrana1:
**Describe the bug**
jdk-8 branch can't be compiled with Java v8

**To Reproduce**
1.Checkout jdk-8 branch 
2. Set Java home to jdk-8
3. Build the maven project.

**Expected behavior**
Should compile with Java v8

**Screenshots**

**Client side logs**
ShardingKey can't be found in DatabricksConnection class
Mapping.of feature is used from Java-9

**Client Environment (please complete the following information):**
 - OS: Windows
 - Java version Java 8
 - Java vendor Azul Jdk
 - Driver Version 

**Additional context**

PR #1266, opened 2026-03-10
Remove ShardingKey methods incompatible with Java 8

samikshya-db:
## Summary

- `java.sql.ShardingKey` was introduced in Java 9 (JDBC 4.3) and does not exist in Java 8. This PR fixes it.
- Fixes https://github.com/databricks/databricks-jdbc/issues/1253

## Test plan
- Github action run for JDK8

## Additional notes

- No unit tests in jdk-8 branch (tests will be added when merging with main branch)

SKIP_COVERAGE_CHECK=No unit tests in jdk-8 branch
NO_CHANGELOG=true


src/main/java/com/databricks/jdbc/api/impl/volume/VolumeUploadCallback.java
```diff
@@ -175,93 +178,102 @@ public void cancelled() {
...
+                            + e.getMessage();
+                    LOGGER.error(e, errorMessage);
+                    uploadFuture.complete(
+                        new VolumePutResult(500, VolumeOperationStatus.FAILED, errorMessage));
+                    return null;
+                  });
+        },
+        delayMs,
+        TimeUnit.MILLISECONDS);
+    scheduler.shutdown();
```
gopalldb: We are calling shutdown immediately after schedule, it is not waiting for task to be completed

PR #1280, opened 2026-03-16
Bind to localhost when finding available ports

joeljons:
## Description
Change so that checking for available ports checks by binding to localhost.

When connecting with OAuth Browser flow (`AuthMech=11;Auth_Flow=2`) a local webserver is started to catch the redirect from the OAuth flow.
If port 8020 (or the port(s) configured by `OAuth2RedirectUrlPort`) was occupied and bound to localhost the port availability check though it was free because it checked using the wildcard adress.
This led to the jdbc connect failed with `DatabricksDriverException: ExternalHeaderFactory not initialized. Call configure() first.`

## Testing
* Updated tests in ClientConfiguratorTest to bind to localhost.
* Added a new test case for where the occupied port is only bound to the wildcard address.
* Manual testing of having other services occupying the port, including the HttpServer started by the OAuth Browser flow from another session.

Multiple servers can be started on the same port if bound to different hosts or the wildcard address. The HttpServer started in `Consent.getOAuthCallbackParameters` binds to localhost, so `ClientConfigurator.isPortAvailable` should also bind to localhost when checking. Although `redirectUrl` in `Consent` can be configured it is always localhost set by `ClientConfigurator.setupU2MConfig`.


joeljons:
It would be great if someone could have a look at this PR

Issue #1233, opened 2026-02-25
[BUG] Leaked socket prevents CRaC checkpointing


NathanEckert:
**Describe the bug**

A leaked socket prevent an application connected to Databricks to be checkpoint by CRaC.
This issue affect all the 3.X driver, while it was possible to do so with the SIMBA 2.X driver

**To Reproduce**

You need a CRaC enabled JDK, (Azul for example)

pom.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.example</groupId>
    <artifactId>databricks-socket-leak-reproducer</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>17</maven.compiler.source>
        <maven.compiler.target>17</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>

    <dependencies>
        <dependency>
            <groupId>com.databricks</groupId>
            <artifactId>databricks-jdbc</artifactId>
            <version>3.2.1</version>
        </dependency>
        <dependency>
            <groupId>org.crac</groupId>
            <artifactId>crac</artifactId>
            <version>1.5.0</version>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>exec-maven-plugin</artifactId>
                <version>3.5.0</version>
                <configuration>
                    <executable>java</executable>
                    <arguments>
                        <argument>--add-opens=java.base/java.nio=ALL-UNNAMED</argument>
                        <argument>-classpath</argument>
                        <classpath/>
                        <argument>com.example.DatabricksSocketLeakReproducer</argument>
                    </arguments>
                </configuration>
            </plugin>
        </plugins>
    </build>

    <profiles>
        <profile>
            <id>crac</id>
            <build>
                <plugins>
                    <plugin>
                        <groupId>org.codehaus.mojo</groupId>
                        <artifactId>exec-maven-plugin</artifactId>
                        <version>3.5.0</version>
                        <configuration>
[...]

PR #1333, opened 2026-03-27
Fix telemetry HTTP client socket leak preventing CRaC checkpoint

gopalldb:
## Summary

Fixes #1325 (follow-up to #1233). After `Connection.close()`, delayed telemetry flush tasks could re-create TELEMETRY HTTP clients that were never closed, leaking TCP sockets and preventing CRaC checkpoint.

**Two cross-thread race conditions fixed:**

1. **TelemetryClient re-creation**: `getTelemetryClient()` after `closeTelemetryClient()` created an orphaned `TelemetryClient` with a periodic flush scheduler that nobody closes. Fixed by tracking closed connection UUIDs and returning `NoopTelemetryClient`, and by reordering `closeTelemetryClient()` to export pending collector events before closing the client.

2. **HTTP client re-creation**: `getClient(ctx, TELEMETRY)` after `removeClient(ctx)` re-created a `DatabricksHttpClient` via `computeIfAbsent` that nobody closes. Fixed by adding `closeConnection()` which permanently marks a connection as closed, causing `getClient()` to return `null`.

### Files changed

| File | Change |
|------|--------|
| `TelemetryClientFactory.java` | Added `closedConnectionUuids` guard, reordered close sequence |
| `DatabricksHttpClientFactory.java` | Added `closedConnections` guard, new `closeConnection()` method |
| `DatabricksConnection.java` | Use `closeConnection()` instead of `removeClient()` |
| `TelemetryPushClient.java` | Null guard for `getClient()` return value |
| `TelemetryHttpClientLeakTest.java` | 3 reproduction/verification tests |
| `RCA_SOCKET_LEAK_TELEMETRY_HTTP_CLIENT.md` | Full RCA with reproduction and verification plan |

### Test results

- 3085 unit tests pass, 0 failures, 0 errors
- 3 new tests specifically reproduce and verify the fix for both race conditions

## Test plan

- [x] `TelemetryHttpClientLeakTest#testGetTelemetryClientAfterCloseReCreatesClient` — verifies `NoopTelemetryClient` returned after close
- [x] `TelemetryHttpClientLeakTest#testGetClientReturnsNullAfterCloseConnection` — verifies `null` returned from HTTP client factory after close
- [x] `TelemetryHttpClientLeakTest#testCloseTelemetryClientWithPendingCollectorEventsReCreatesClient` — verifies pending collector events don't cause re-creation
- [x] All 152 existing telemetry tests pass
- [x] Full unit test suite (3085 tests) passes
- [ ] Manual verification with CRaC-enabled JDK (0 sockets after `Connection.close()`)


src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -31,17 +40,42 @@ public IDatabricksHttpClient getClient(IDatabricksConnectionContext context) {
...
       IDatabricksConnectionContext context, HttpClientType type) {
+    // Prevent creating new HTTP clients for connections that have been closed.
+    // This guards against delayed TelemetryPushTask executions that call
+    // getClient(ctx, TELEMETRY) after removeClient(ctx) has already run.
+    String connectionUuid = context.getConnectionUuid();
+    if (connectionUuid != null && closedConnections.contains(connectionUuid)) {
+      LOGGER.debug(
+          "Rejecting getClient() for closed connection {} with type {}",
+          context.getConnectionUuid(),
+          type);
```
vikrantpuppala: **[F1] TOCTOU race — the PR does not actually close the race it claims to fix** (High)

The guard reads `closedConnections.contains(uuid)` then falls through to `instances.computeIfAbsent(...)`. These are two separate operations on two separate maps with no mutual exclusion. Interleaving:
1. Thread A (delayed `TelemetryPushTask` on the 10-thread pool): evaluates `closedConnections.contains(uuid)` → false.
2. Thread B (`DatabricksConnection.close()`): runs `closeConnection()` — adds UUID, wipes `instances`.
3. Thread A resumes `computeIfAbsent` → creates a new `DatabricksHttpClient(TELEMETRY)` → leaks a TCP socket.

This is literally the race the RCA describes on the old code. The new code narrows but does not close it.

**Fix — fuse guard and map op atomically:**
```java
String uuid = context.getConnectionUuid();
return instances.compute(
    getClientKey(uuid, type),
    (k, existing) -> {
      if (existing != null) return existing;
      if (uuid != null && closedConnections.contains(uuid)) return null;
      return new DatabricksHttpClient(context, type);
    });
```

Apply the same pattern in `TelemetryClientFactory.getTelemetryClient()`.

Flagged independently by 4 reviewers (language, security, performance, devils-advocate).

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -17,6 +18,14 @@ public class DatabricksHttpClientFactory {
   private final ConcurrentHashMap<SimpleEntry<String, HttpClientType>, DatabricksHttpClient>
       instances = new ConcurrentHashMap<>();
 
+  /**
+   * Tracks connection UUIDs for which removeClient() has been called. Prevents getClient() from
+   * re-creating HTTP clients for closed connections via computeIfAbsent. Without this guard,
+   * delayed TelemetryPushTask executions can create orphaned HTTP clients that leak TCP sockets.
+   * See GitHub issue #1325.
```
vikrantpuppala: **[F2] Unbounded heap growth — PR trades a bounded socket leak for an unbounded heap leak** (High)

`closedConnections` is a process-wide `ConcurrentHashMap.newKeySet()` that is append-only for the JVM lifetime — no TTL, no cap, no eviction, no removal path. `TelemetryClientFactory.closedConnectionUuids` has the same shape and clears only in `reset()` (test-only).

Per-entry cost ≈ 120 B × 2 sets ≈ **~240 B per closed connection**. 100 closes/sec (realistic for HikariCP under load) → ~2 GB/month. This hits exactly the long-lived CRaC workloads this PR targets.

**Fix options (in preference order):**
1. Bound via Caffeine `Cache` with `expireAfterWrite(5-10 minutes)` — the race window is seconds, not JVM lifetime.
2. **Flip to an allowlist**: insert UUID in `getClient` on first use, remove in `closeConnection`. Bounded to live connections and solves F1's TOCTOU for free.
3. Encode the tombstone in the existing `instances` map.

Flagged by 6 reviewers (performance, maintainability, security, ops, devils-advocate, architecture).

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -31,17 +40,42 @@ public IDatabricksHttpClient getClient(IDatabricksConnectionContext context) {
 
   public IDatabricksHttpClient getClient(
       IDatabricksConnectionContext context, HttpClientType type) {
+    // Prevent creating new HTTP clients for connections that have been closed.
+    // This guards against delayed TelemetryPushTask executions that call
+    // getClient(ctx, TELEMETRY) after removeClient(ctx) has already run.
+    String connectionUuid = context.getConnectionUuid();
+    if (connectionUuid != null && closedConnections.contains(connectionUuid)) {
+      LOGGER.debug(
```
vikrantpuppala: **[F3] `getClient()` silently widened to nullable return — 7 unverified callers** (High)

Pre-PR `getClient()` was non-null. Post-PR it can return null. Only `TelemetryPushClient.pushEvent()` was updated. Verified call sites that store or dereference without a null check:

| File:line | Usage | Null-safe? |
|---|---|---|
| `DatabricksThriftAccessor.java:662` | passed to `DatabricksHttpTTransport` ctor | No |
| `DatabricksTokenFederationProvider.java:74` | stored in `this.hc` | No |
| `AzureMSICredentialProvider.java:43` | stored in `this.httpClient` | No |
| `PrivateKeyClientCredentialProvider.java:27` | stored | No |
| `DatabricksDriverFeatureFlagsContext.java:92` | runs on a background scheduler; `execute()` caught by `catch(Exception)` at TRACE → **silent failure of FF refresh** | No |
| `ArrowStreamResult.java:53, 153` | ctor param → NPE later in chunk download | No |

No `@Nullable` annotation despite the codebase using `javax.annotation.Nullable` elsewhere (`SqlParameter.java:10`, `DatabricksSession.java:102`). Contract change is invisible to callers.

**Fix (preferred)**: return a rejecting sentinel `IDatabricksHttpClient` whose `execute()` throws `ConnectionClosedException` (symmetric to `NoopTelemetryClient`). All callers stay safe by default.
**Alternative**: split into `getClient()` (throws) + `getClientOrNull()` (used only by `TelemetryPushClient`), or add `@Nullable` + Javadoc and null-check every caller.

Flagged by 6 reviewers.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -31,17 +40,42 @@ public IDatabricksHttpClient getClient(IDatabricksConnectionContext context) {
...
+   */
   public void removeClient(IDatabricksConnectionContext context) {
     for (HttpClientType type : HttpClientType.values()) {
       removeClient(context, type);
     }
   }
 
+  /**
+   * Permanently closes all HTTP clients for the given connection and prevents new ones from being
+   * created. This should be called from DatabricksConnection.close() to prevent delayed
```
vikrantpuppala: **[F5] `closeConnection()` NPEs on null UUID — inconsistent with `getClient()`'s null guard** (High)

`ConcurrentHashMap.newKeySet()` rejects null elements — `closedConnections.add(null)` throws NPE. Yet `getClient()` in the same file at the new lines `if (connectionUuid != null && closedConnections.contains(connectionUuid))` explicitly null-guards, implying nulls are considered possible. The two paths disagree on the invariant.

Same issue in `TelemetryClientFactory.closeTelemetryClient` at the `closedConnectionUuids.add(connectionUuid)` calls inside the `computeIfPresent` lambdas — no null-guard.

**Fix**: Pick one invariant. Either (a) assert non-null in `closeConnection`/`closeTelemetryClient` and document the constraint on `IDatabricksConnectionContext.getConnectionUuid()`, or (b) mirror the `getClient` guard:
```java
String uuid = context.getConnectionUuid();
if (uuid != null) closedConnections.add(uuid);
```

Flagged by 2 reviewers (architecture, language).

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
```diff
@@ -137,41 +153,52 @@ public ITelemetryClient getTelemetryClient(IDatabricksConnectionContext connecti
...
    * Closes telemetry client for a connection. Thread-safe: computeIfPresent ensures atomic locking,
    * preventing race conditions between connection removal and addition.
+   *
+   * <p>The connection UUID is added to closedConnectionUuids FIRST to prevent getTelemetryClient()
+   * from re-creating a TelemetryClient during or after the close sequence. Pending
+   * TelemetryCollector events are exported BEFORE the TelemetryClient is closed, so they are
+   * flushed through the existing client. See GitHub issue #1325.
    */
   public void closeTelemetryClient(IDatabricksConnectionContext connectionContext) {
     String key = TelemetryHelper.keyOf(connectionContext);
```
vikrantpuppala, line 160: **[F6] Reordered close is not fail-safe — one exception skips ALL cleanup and re-opens the entire leak** (High)

The new ordering runs `exportAllPendingTelemetryDetails()` BEFORE the two `computeIfPresent` blocks that (a) mark UUID closed, (b) close the client, (c) remove the holder.

If `exportAllPendingTelemetryDetails()` or the `exportTelemetryLog` it calls throws **any unchecked exception**, execution exits `closeTelemetryClient` and propagates up to `DatabricksConnection.close()`. Consequences:
- `TelemetryClient` stays in the holder map with its periodic flush scheduler alive
- `closedConnectionUuids` never updated → future `getTelemetryClient()` won't return Noop
- `DatabricksHttpClientFactory.closeConnection()` on the next line of `DatabricksConnection.close()` **never runs**

One export-time exception silently re-opens the entire leak the PR set out to fix. The old ordering was actually more robust (cleanup ran first, export was a trailer).

**Fix**:
1. Mark UUIDs closed **unconditionally up-front** (before export).
2. Run holder cleanup.
3. Run `exportAllPendingTelemetryDetails` **last**, wrapped in `try/catch` that logs-and-swallows.
4. In `DatabricksConnection.close()` wrap the cleanup chain in try/finally so `closeConnection(ctx)` always runs even if earlier steps throw.

Flagged by ops reviewer.

src/test/java/com/databricks/jdbc/telemetry/TelemetryHttpClientLeakTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+      // After closeTelemetryClient(), getTelemetryClient() should NOT create a new
+      // real TelemetryClient. It should return NoopTelemetryClient to prevent leaks.
+      assertInstanceOf(
+          NoopTelemetryClient.class,
+          recreated,
+          "LEAK BUG (issue #1325): getTelemetryClient() after closeTelemetryClient() "
+              + "created a new TelemetryClient instead of returning NoopTelemetryClient. "
+              + "This orphaned client will never be closed, and its periodic flush creates "
+              + "TELEMETRY HTTP clients that leak TCP sockets.");
```
vikrantpuppala: **[F4] Tests are single-threaded; do not reproduce the cross-thread race** (High)

The RCA describes a race between main thread (`DatabricksConnection.close()`) and pool thread (delayed `TelemetryPushTask.run()` on the 10-thread `telemetryExecutorService`). All three tests run sequentially on the JUnit main thread — no `CountDownLatch`/`CyclicBarrier`/real executor/parallel stress.

- **Test #1 (`testGetTelemetryClientAfterCloseReCreatesClient`)** is **tautological**: `closedConnectionUuids.add(uuid)` runs inside `closeTelemetryClient` first, so a sequential `getTelemetryClient` call trivially returns `NoopTelemetryClient`. Cannot fail against a partial revert of the fix.
- **Test #3** mocks `TelemetryHelper.exportTelemetryLog` with a lambda that calls `getTelemetryClient(ctx)`. But the real `exportTelemetryLog` reads from `DatabricksThreadContextHolder.getConnectionContext()` — a thread-local that `Connection.close()` clears. The real re-creation vector is `TelemetryPushClient.pushEvent()` capturing `ctx` at ctor time (verified in source at `TelemetryPushClient.java:47-49`) — which no test covers.

**Fix — add a real cross-thread test:**
```java
@Test
void testRaceBetweenCloseAndDelayedPushTask() throws Exception {
  CountDownLatch blockInPush = new CountDownLatch(1);
  CountDownLatch closeDone = new CountDownLatch(1);
  // Spy push client blocks inside pushEvent until main thread calls closeConnection
  // Submit flush task on real pool
  // Main thread: await blockInPush entered, then closeConnection(ctx)
  // Release latch; assert instances map has no TELEMETRY entry for uuid after join
}
```

Delete or rewrite Tests #1 and #3 so they would fail against a partial revert of the fix.

Flagged by 2 reviewers (test, devils-advocate).

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -31,17 +40,42 @@ public IDatabricksHttpClient getClient(IDatabricksConnectionContext context) {
...
+   * Removes and closes all HTTP clients for the given connection. Does NOT mark the connection as
+   * closed — the client can be re-created by a subsequent getClient() call.
+   */
   public void removeClient(IDatabricksConnectionContext context) {
     for (HttpClientType type : HttpClientType.values()) {
       removeClient(context, type);
     }
   }
 
+  /**
```
vikrantpuppala: **[F7] No `@VisibleForTesting reset()` on `DatabricksHttpClientFactory` — test pollution** (Medium)

`TelemetryClientFactory.reset()` (in this PR at the hunk around line 243) clears `closedConnectionUuids`. `DatabricksHttpClientFactory` has no equivalent. It's a JVM singleton and `closedConnections` is an instance field.

Consequence: any test that calls `closeConnection(ctx)` permanently adds UUID to `closedConnections` for the rest of the test JVM. A later test reusing the same UUID gets null from `getClient()` with no symptom except "my test fails mysteriously when run after another test."

**Fix**: Add `@VisibleForTesting` `void reset()` on `DatabricksHttpClientFactory` that clears both `instances` and `closedConnections`. Call in `@BeforeEach`/`@AfterEach` of `TelemetryHttpClientLeakTest`.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -31,17 +40,42 @@ public IDatabricksHttpClient getClient(IDatabricksConnectionContext context) {
...
+          context.getConnectionUuid(),
+          type);
+      return null;
+    }
     return instances.computeIfAbsent(
         getClientKey(context.getConnectionUuid(), type),
         k -> new DatabricksHttpClient(context, type));
   }
 
+  /**
```
vikrantpuppala, line 54: **[F8] `removeClient(ctx)` is a public footgun with zero production callers** (Medium)

After this PR, the no-arg `removeClient(IDatabricksConnectionContext)` overload has exactly one caller in the repo: `DatabricksHttpClientTest.java:240` (a test). All production call sites were migrated to `closeConnection()` by this PR.

The RCA explicitly identifies `removeClient`'s "allows re-creation via `computeIfAbsent`" behavior as the root cause of #1325. Shipping two nearly-identically-named public methods where one silently re-enables the bug this PR fixes is a future-maintainer trap — the name "removeClient" reads as the obvious choice on IDE autocomplete.

**Fix**: Delete the no-arg overload and migrate the test; or mark `@VisibleForTesting` / `@Deprecated` with a pointer to `closeConnection`. At minimum, rename to signal intent (e.g., `evictClientsForReconnect`).

Flagged by 4 reviewers (maintainability, architecture, devils-advocate, agent-compat).

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
```diff
@@ -137,41 +153,52 @@ public ITelemetryClient getTelemetryClient(IDatabricksConnectionContext connecti
   /**
    * Closes telemetry client for a connection. Thread-safe: computeIfPresent ensures atomic locking,
    * preventing race conditions between connection removal and addition.
+   *
+   * <p>The connection UUID is added to closedConnectionUuids FIRST to prevent getTelemetryClient()
+   * from re-creating a TelemetryClient during or after the close sequence. Pending
+   * TelemetryCollector events are exported BEFORE the TelemetryClient is closed, so they are
```
vikrantpuppala: **[F9] Reordered close has a concurrency regression under concurrent close of same UUID** (Medium)

New ordering: (1) `removeCollector`, (2) `exportAllPendingTelemetryDetails()` which calls `getTelemetryClient(ctx)` through the factory, (3) `computeIfPresent` removes holder and marks UUID closed.

Single-threaded: fine. Under **concurrent** close of the same UUID (legal via the public API):
- Thread A starts step 2's export loop.
- Thread B completes step 3 first (removes holder, marks UUID closed).
- Thread A's next `getTelemetryClient(ctx)` now returns `NoopTelemetryClient` → events silently dropped.
- Or a concurrent re-open races in between, and step 2's `getTelemetryClient` creates a new `TelemetryClient` (the exact original bug).

Old ordering was order-insensitive because holder-remove+close was atomic under `computeIfPresent`.

**Fix**: Mark UUID closed BEFORE export, and export via a directly-held `TelemetryClient` reference (either look it up from the holder map first, or have the removal return the client). Don't route the export back through the factory.

Flagged by architecture reviewer.

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
```diff
@@ -422,7 +422,7 @@ public void close() throws SQLException {
     TelemetryClientFactory.getInstance().closeTelemetryClient(connectionContext);
     DatabricksClientConfiguratorManager.getInstance().removeInstance(connectionContext);
     DatabricksDriverFeatureFlagsContextFactory.removeInstance(connectionContext);
-    DatabricksHttpClientFactory.getInstance().removeClient(connectionContext);
+    DatabricksHttpClientFactory.getInstance().closeConnection(connectionContext);
```
vikrantpuppala, line 427: **[F10] No regression test for the `Connection.close()` post-condition** (Medium)

This one-line change alters observable behavior: after `connection.close()`, `DatabricksHttpClientFactory.getInstance().getClient(ctx, TELEMETRY)` now returns **null** (previously always returned a new client). Telemetry pushes submitted during close are now silently dropped.

No existing test was modified; grep shows no test asserts the new post-close invariant. The "3085 tests pass" claim only proves nothing else broke, not that the new behavior is covered.

**Fix**: Add one focused assertion in the existing DatabricksConnection test class:
```java
@Test
void getClientReturnsNullAfterConnectionClose() {
  DatabricksConnection conn = ...;
  conn.close();
  assertNull(DatabricksHttpClientFactory.getInstance()
      .getClient(conn.getConnectionContext(), HttpClientType.TELEMETRY));
}
```

Flagged by test reviewer.

src/main/java/com/databricks/jdbc/telemetry/TelemetryPushClient.java
```diff
@@ -47,6 +47,11 @@ public void pushEvent(TelemetryRequest request) throws Exception {
     IDatabricksHttpClient httpClient =
         DatabricksHttpClientFactory.getInstance()
             .getClient(connectionContext, HttpClientType.TELEMETRY);
+    if (httpClient == null) {
+      // Connection was closed — HTTP client factory rejected the request to prevent socket leaks.
+      LOGGER.debug("Skipping telemetry push: connection has been closed");
```
vikrantpuppala, line 52: **[F12] Rejection log at DEBUG; no metric for dropped telemetry** (Medium)

Two invisibility paths added by this PR:
1. `DatabricksHttpClientFactory.getClient` — rejection logged at `LOGGER.debug("Rejecting getClient() for closed connection {} with type {}")`. Below default production log level. If a non-TELEMETRY caller hits this path due to a future bug or code change, operators never see it.
2. `TelemetryPushClient.pushEvent` on null — `LOGGER.debug("Skipping telemetry push: connection has been closed")`. In aggressive close scenarios (shutdown, failover, pool eviction), every dropped push is invisible. No counter. Operators cannot answer "is telemetry working?"

**Fix**:
- Raise the `getClient` rejection log to WARN for non-TELEMETRY types (preserves DEBUG for the expected path).
- Add an `AtomicLong droppedOnClosedCount` on the factory; emit a periodic aggregated WARN when > 0, or expose via JMX/existing metrics.
- Include `Thread.currentThread().getName()` in the rejection log for diagnosis.

Flagged by 3 reviewers (ops, agent-compat, maintainability).

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
```diff
@@ -34,6 +34,14 @@ public class TelemetryClientFactory {
   @VisibleForTesting
   final Map<String, TelemetryClientHolder> noauthTelemetryClientHolders = new ConcurrentHashMap<>();
 
+  /**
+   * Tracks connection UUIDs that have been closed. When a connection is closed, its UUID is added
+   * here so that subsequent getTelemetryClient() calls (e.g., from delayed flush tasks or collector
+   * exports) return NoopTelemetryClient instead of re-creating an orphaned TelemetryClient. This
+   * prevents the socket leak described in GitHub issue #1325.
+   */
```
vikrantpuppala, line 42: **[F13] Duplicated registry pattern across two factories** (Medium)

`closedConnectionUuids` here and `closedConnections` in `DatabricksHttpClientFactory` are the same concept implemented twice with different names. Both are `ConcurrentHashMap.newKeySet()` of connection UUIDs, both checked before `compute*` in the same way, both added on close. Neither is bounded.

**Fix**: Extract a `ClosedConnectionRegistry` helper in `com.databricks.jdbc.common` with `markClosed(uuid)` / `isClosed(uuid)` / `clear()`. Back it with a bounded cache (Caffeine, e.g. `expireAfterWrite(5 min)` or max ~10K entries LRU). Both factories hold a reference (or share a singleton). Solves the duplication AND F2's memory growth in one change.

Flagged by 3 reviewers (maintainability, agent-compat, ops).

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
```diff
@@ -31,17 +40,42 @@ public IDatabricksHttpClient getClient(IDatabricksConnectionContext context) {
 
   public IDatabricksHttpClient getClient(
```
vikrantpuppala, line 47: **[F16] `@Nullable` annotation + irreversibility are undocumented** (Low/Medium)

The codebase already uses `javax.annotation.Nullable` (e.g., `SqlParameter.java:10`, `DatabricksSession.java:102,110`, `IDatabricksSession.java:21,24`). This PR silently widens `getClient` to a nullable return without annotation or Javadoc update. Agents and humans adding future callers will omit the null check — see F3 for the consequence.

Also document that `closeConnection` is **irreversible** — the UUID is blacklisted for the lifetime of the factory. The field comment says "permanently closed" but the public method Javadoc only says "permanently closes all HTTP clients" — reads as "closes them once," not "blacklists UUID forever."

**Fix**:
```java
/**
 * @return the HTTP client, or {@code null} if {@link #closeConnection} has been
 *         called for this context. Callers MUST null-check when the call can
 *         race with connection close (e.g., async tasks).
 */
@Nullable
public IDatabricksHttpClient getClient(
    IDatabricksConnectionContext context, HttpClientType type) { ... }
```

Add to `closeConnection` Javadoc: *"This is irreversible — the UUID is blacklisted for the lifetime of this factory. Do not call until you're certain the context will never be used again."* Same on `TelemetryClientFactory.closeTelemetryClient`.

docs/RCA_SOCKET_LEAK_TELEMETRY_HTTP_CLIENT.md
```diff
@@ -0,0 +1,208 @@
+# RCA: Leaked Socket Prevents CRaC Checkpointing (Issue #1325)
```
vikrantpuppala: **[F17] RCA doc in `docs/` is unprecedented and point-in-time** (Low)

`docs/` currently contains only enduring reference material (LOGGING.md, TESTING.md, JDBC_METHOD_INVENTORY.md, JDBC_SPEC_COVERAGE_ANALYSIS.md, features/). This adds a 208-line root-cause analysis for a single bug.

Issues:
- Sets precedent without an established convention (no `docs/rca/README.md` policy, no numbering scheme).
- Hardcodes line numbers ("line 421", "line 172") that will rot on the next edit to `DatabricksConnection.close()`.
- Content is point-in-time and will never be updated.
- Duplicates what should live in the PR description / JIRA / issue #1325.

**Fix**: Delete this file and fold the content into the PR description + issue #1325 comments. If in-tree RCAs are desired, establish a convention first (e.g., `docs/rca/NNN-title.md` + a `README.md` policy).

Flagged by 3 reviewers (maintainability, agent-compat, language).

src/test/java/com/databricks/jdbc/telemetry/TelemetryHttpClientLeakTest.java
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * <p>Root cause: In TelemetryClientFactory.closeTelemetryClient(), the TelemetryCollector's pending
+ * events are exported AFTER the TelemetryClient has been closed and removed from the holder map.
+ * The export path calls getTelemetryClient(ctx) which re-creates a new TelemetryClient. That new
+ * client's eventual flush calls DatabricksHttpClientFactory.getClient(ctx, TELEMETRY) — and since
+ * removeClient(ctx) already ran, this creates an orphaned HTTP client that leaks a TCP socket.
+ */
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+public class TelemetryHttpClientLeakTest {
```
vikrantpuppala, line 31: **[F20] Test hygiene: `Strictness.LENIENT`, unused stub, empty catch** (Low)

- Line with `@MockitoSettings(strictness = Strictness.LENIENT)` silently masks unused-stub bugs — combined with broad `any()` matchers, tests can pass while stubbing the wrong method. Remove `LENIENT` and fix any resulting unused-stub errors.
- `when(ctx.getHost()).thenReturn(host)` is unused; the production `keyOf` path uses `getHostForOAuth`. Would fail under strict Mockito. Remove.
- The empty catch around `when(ctx.getHostUrl()).thenReturn(...)` silently swallows checked exceptions during setup. Use `doReturn("https://" + host).when(ctx).getHostUrl()` to bypass the checked exception cleanly.

vikrantpuppala:
# [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) Report

**Merge Safety Score: 10/100 — CRITICAL RISK**
**Reviewers: 9/9 delivered (security, architecture, language, ops, performance, test, maintainability, agent-compat, devils-advocate)**

## Executive Summary

Targets a real cross-thread race from issue #1325 and the defense-in-depth intent is sound. But the implementation has fundamental issues that defeat the fix:

1. **The headline guard does not close the race it claims to fix** — `contains()` check sits outside `computeIfAbsent` (flagged by 4 reviewers) — see inline **F1**.
2. **One unchecked exception from `exportAllPendingTelemetryDetails()` skips ALL cleanup** — **F6**.
3. **`closeConnection()` NPEs on null UUID** — asymmetric with `getClient()`'s null guard — **F5**.
4. **Bounded socket leak traded for unbounded heap leak** — two append-only process-wide sets; hits the long-running CRaC workloads this PR targets — **F2**.
5. **Tests are strictly single-threaded** — don't exercise the cross-thread race; two of three are tautological or mis-mocked — **F4**.
6. **`getClient()` silently widened to nullable** — 7 existing callers are not null-safe — **F3**.

Recommend blocking merge until F1, F2, F3, F4, F5, F6 are addressed.

## Findings by Severity

### High (blocker) — posted as inline comments
- **F1** TOCTOU race in `getClient()` / `getTelemetryClient()` (4 reviewers)
- **F2** Unbounded heap growth in closed-UUID sets (6 reviewers)
- **F3** `getClient()` silently nullable; 7 unsafe callers (6 reviewers)
- **F4** Tests don't reproduce the cross-thread race (2 reviewers)
- **F5** `closeConnection()` NPEs on null UUID (2 reviewers)
- **F6** Reordered close is not fail-safe — one exception re-opens the leak (ops)

### Medium — posted as inline comments
- **F7** No `@VisibleForTesting reset()` on `DatabricksHttpClientFactory` — test pollution
- **F8** `removeClient(ctx)` is a public footgun with zero production callers
- **F9** Reordered close has a concurrency regression under concurrent close of same UUID
- **F10** No regression test for `Connection.close()` post-condition
- **F12** Rejection log at DEBUG; no metric for dropped telemetry
- **F13** Duplicated registry pattern across two factories

### Low — summarized here (not posted inline to reduce noise)
- **F11** The close-ordering fix itself is untested. A future refactor reverting the reorder would pass every test because the guard alone would catch it.
- **F14** `TelemetryClientFactory.getTelemetryClient()` runs `isTelemetryAllowedForConnection()` (which can reach into feature-flag infra + HTTP) before the `closedConnectionUuids.contains` check. Move the closed-UUID check to be the first statement.
- **F15** `closedConnectionUuids.add(uuid)` happens *inside* `computeIfPresent` — which runs AFTER the collector export. During export the UUID isn't marked yet, so events can still submit pool flushes that hit F1's race in the HTTP factory.
- **F16** Missing `@Nullable` on `getClient` + irreversibility of `closeConnection` undocumented (posted inline).
- **F17** RCA doc in `docs/` is unprecedented and point-in-time (posted inline).
- **F18** Naming inconsistency: `closedConnectionUuids` vs `closedConnections` — subsumed by F13.
- **F19** `closeConnection()` on `DatabricksHttpClientFactory` reads as "close a JDBC Connection". Consider `markConnectionClosed()` / `onConnectionClosed()`.
- **F20** Test hygiene: `Strictness.LENIENT`, empty catch, unused `getHost` stub (posted inline).
- **F21** `context.getConnectionUuid()` called twice per `getClient` — cache to a local.
- **F22** Issue #1325 rationale repeated ~5× across files. One authoritative comment + one-liners referencing it.
- **F23** Close-latency regression: collector export now runs synchronously on `close()` before client close. Performance reviewer concluded net latency is unchanged; flag for follow-up if telemetry endpoint ever degrades.
- **F24** `TelemetryClient.close()` cancels the periodic flush task AFTER `flush(true).get()`. If the scheduler is mid-dispatch when cancel runs, a late `TelemetryPushTask` can still land on the pool and hit F1's race.

## Verified but No Action
- Close-sequence reorder does NOT add an extra sync HTTP round-trip (performance reviewer confirmed).
- No security concerns on auth/TLS/secrets/data exposure.
- `DatabricksDriverFeatureFlagsContextFactory.removeInstance` DOES call `shutdown()` cancelling its scheduler (verified) — the feature-flag scheduled-thread NPE concern from F3 is narrow, not broad.

## Suggested Minimum Path to Merge
1. **F1**: fuse guard + map op into a single `compute(...)` (both factories).
2. **F5**: null-guard `closedConnections.add(uuid)` / `closedConnectionUuids.add(uuid)` (or assert non-null with a documented invariant).
3. **F6**: mark UUIDs closed up-front; wrap `exportAllPendingTelemetryDetails` in try/catch-swallow; wrap `DatabricksConnection.close()` cleanup chain in try/finally.
4. **F2**: replace sets with a bounded Caffeine cache OR flip to an allowlist of open UUIDs (solves F1 + F2 in one change).
5. **F3**: return a rejecting `IDatabricksHttpClient` sentinel (symmetric with `NoopTelemetryClient`) so all callers stay safe.
6. **F4**: add one real cross-thread test with `CountDownLatch`; rewrite or delete the tautological Test #1 and the mis-mocked Test #3.

---

*Review generated by [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (9 parallel specialized reviewers). Feedback welcome in [#code-review-squad-feedback](https://databricks.slack.com/channels/code-review-squad-feedback).*

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: Fixed. Replaced the separate contains() + computeIfAbsent() with a single computeIfAbsent() where the guard is inside the lambda. The check and creation are now atomic — no interleaving window for closeConnection().

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: Fixed. Replaced the append-only closed-connections denylist with an open-connections allowlist. The set naturally shrinks as connections close — no unbounded growth in long-running CRaC JVMs.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: Fixed. With the allowlist, getClient() only returns null when the UUID is not in the open set. Non-TELEMETRY callers (auth providers, ArrowStreamResult, ThriftAccessor, etc.) run during active connection lifecycle when the UUID is registered — null return is impossible for them. Only TelemetryPushClient (delayed flush) can race, and it already handles null.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: Fixed. closeConnection() now guards with `if (uuid != null)` before `openConnections.remove(uuid)`.

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
gopalldb: Fixed. closeTelemetryClient() now: (1) removes UUID from openConnectionUuids FIRST, (2) exports pending events inside try, (3) computeIfPresent holder cleanup in finally. Cleanup always runs even if export throws.

src/test/java/com/databricks/jdbc/telemetry/TelemetryHttpClientLeakTest.java
gopalldb: Fixed. Added `testConcurrentCloseAndGetClientDoesNotLeak` — 20 threads with CountDownLatch, half calling closeConnection/closeTelemetryClient, half calling getClient/getTelemetryClient simultaneously. Verifies post-close invariant: getClient returns null, getTelemetryClient returns NoopTelemetryClient.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: Fixed. Added `@VisibleForTesting public void reset()` to DatabricksHttpClientFactory — closes all clients, clears instances and openConnections. Used in @BeforeEach/@AfterEach in tests.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: Fixed. Both `removeClient()` overloads marked `@VisibleForTesting`. Production code uses `closeConnection()`.

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
gopalldb: Fixed. With the allowlist, `openConnectionUuids.remove(uuid)` is idempotent — concurrent closes both remove the same UUID, and `computeIfPresent` is atomic per-key. The try-finally ensures cleanup runs regardless.

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnection.java
gopalldb: Covered by the new `testGetClientReturnsNullAfterCloseConnection` and `testConcurrentCloseAndGetClientDoesNotLeak` tests which verify the post-close invariant.

src/main/java/com/databricks/jdbc/telemetry/TelemetryPushClient.java
gopalldb: Acceptable for telemetry — dropped telemetry events are not user-facing. The DEBUG log provides signal for developers investigating issues.

src/main/java/com/databricks/jdbc/telemetry/TelemetryClientFactory.java
gopalldb: Both factories now use the same consistent allowlist pattern (`openConnections` / `openConnectionUuids`). Separate sets are intentional — the factories have different lifecycles and sharing would add coupling.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
gopalldb: With the allowlist, null return only happens for TELEMETRY type callers (delayed flush path). The behavior is documented in the getClient() Javadoc.

docs/RCA_SOCKET_LEAK_TELEMETRY_HTTP_CLIENT.md
gopalldb: Deleted. The rationale is documented in the PR description and commit message.

src/test/java/com/databricks/jdbc/telemetry/TelemetryHttpClientLeakTest.java
gopalldb: Fixed. Removed `@MockitoSettings(strictness = Strictness.LENIENT)`, split mock contexts into `createHttpContext()` and `createTelemetryContext()` with only the stubs each test needs. No unused stubs.

gopalldb:
Addressed all review findings in commit e193b109. The core change: **replaced the append-only denylist (closed-connections set) with an allowlist (open-connections set)** in both factories.

**High (all fixed):**
- **F1**: TOCTOU race — guard fused into `computeIfAbsent` lambda, check and creation are atomic
- **F2**: Unbounded heap growth — allowlist naturally shrinks as connections close
- **F3**: Nullable getClient callers — non-TELEMETRY callers run during active lifecycle, null impossible
- **F4**: Single-threaded tests — added concurrent test with 20 threads + CountDownLatch
- **F5**: NPE on null UUID — guarded with null check before `remove()`
- **F6**: Export exception skips cleanup — `closeTelemetryClient` uses try-finally

**Medium (fixed):**
- **F7**: Added `reset()` to `DatabricksHttpClientFactory`
- **F8**: `removeClient()` marked `@VisibleForTesting`
- **F9**: Allowlist + `computeIfPresent` atomicity handles concurrent close
- **F10**: Covered by new tests
- **F13**: Both factories use consistent allowlist pattern, separate sets intentionally

**Low (fixed):**
- **F17**: RCA doc deleted
- **F20**: Removed `LENIENT`, split mock contexts, no unused stubs

**Deferred (low impact):**
- F12 (DEBUG log level): acceptable for telemetry
- F16 (@Nullable): documented in Javadoc
- F14 (check ordering): allowlist check is now first, before `isTelemetryAllowedForConnection()`

Connections now call `registerConnection()` during `open()` before any `getClient()` call.

src/main/java/com/databricks/jdbc/dbclient/impl/http/DatabricksHttpClientFactory.java
vikrantpuppala: Re-opening this — the CI failures on this PR are evidence that F3 is NOT fixed.

The reply asserts "Non-TELEMETRY callers (auth providers, FF context, thrift accessor) all call `getClient()` during normal operation when the connection is open — they get a real client." The failing test `AzureMSICredentialProviderTest.testCredentialProviderAuthType` (line 61) contradicts that. Its constructor at `AzureMSICredentialProvider.java:43` calls `DatabricksHttpClientFactory.getInstance().getClient(connectionContext)` BEFORE any `registerConnection()` runs in the test. The allowlist doesn't contain the test UUID, so `getClient()` now returns null and `this.httpClient` is silently assigned null. Mockito's `STRICT_STUBS` catches the now-unused `getHttpMaxConnectionsPerRoute()` stub → test errors out. The same shape likely explains the other two unit-test errors (`DatabricksBatchExecutorTest.testExecuteBatch_CommandReturnsResultSet`, `ExecutionResultFactoryTest.testGetResultSet_volumeOperationThriftResp`).

Worse than the CI symptom: the allowlist refactor inverted the failure window. Pre-PR, `getClient()` only returned null for *already-closed* connections, so callers that ran before close were safe. Post-PR, `getClient()` returns null for any UUID *not yet registered* — that window now includes the constructor of every auth provider, the thrift accessor, FF context, and `ArrowStreamResult`. In production, any caller constructed outside the `registerConnection() → closeConnection()` window will silently store `null` and NPE on the next call to `this.httpClient.execute(...)`, with a stack trace pointing at the eventual use site rather than the wiring bug.

Recommended fix: either (a) return a rejecting sentinel `IDatabricksHttpClient` whose `execute()` throws `ConnectionClosedException` (symmetric to `NoopTelemetryClient`) so all callers stay safe by default, or (b) add `@Nullable` to `getClient()` + Javadoc and null-check every production caller — at minimum `DatabricksThriftAccessor.java:710`, `DatabricksTokenFederationProvider`, `AzureMSICredentialProvider:43`, `PrivateKeyClientCredentialProvider`, `DatabricksDriverFeatureFlagsContext`, `ArrowStreamResult`.

gopalldb: You're right — the pure allowlist inverted the failure window. Fixed in commit 5d653442.

The guard now uses two sets: `everRegistered` + `openConnections`. The logic is:
```
if (everRegistered.contains(uuid) && !openConnections.contains(uuid))
  → reject (connection was registered then explicitly closed)
else
  → allow (never registered yet, or still open)
```

This means:
- **Auth providers, thrift accessors, FF context** constructing before `registerConnection()` → UUID not in `everRegistered` → allowed through, get a real client. No null, no NPE.
- **Delayed flush tasks** after `closeConnection()` → UUID in `everRegistered` AND not in `openConnections` → rejected, returns null (TelemetryPushClient already handles null).
- **Active connections** → UUID in both sets → allowed.

Reverted the `registerConnection()` calls added to `AzureMSICredentialProviderTest` and `ExecutionResultFactoryTest` — they're no longer needed since unregistered UUIDs pass through.

vikrantpuppala: Thanks for the iterate — the immediate test failure is fixed and the F3 test-visible path is closed at `5d653442`. Two concerns remain though:

**1. The new `everRegistered` set re-introduces F2 (unbounded heap growth).**

`DatabricksHttpClientFactory.everRegistered` is grown by `registerConnection()` and never removed — `closeConnection()` only removes from `openConnections`, leaving the UUID in `everRegistered` forever (which is what makes the guard work). The class-level Javadoc at line 33 says "Both sets shrink together on closeConnection" — that's inaccurate; only `openConnections` shrinks. Net result: ~120 B per closed connection, JVM-lifetime. HikariCP at 100 closes/sec → ~2 GB/month. Hits exactly the long-lived CRaC workloads this PR targets. This is the same shape as the original F2 finding.

**2. F3's underlying nullable contract is still open for non-test, non-TELEMETRY callers.**

`getClient()` is still nullable, still no `@Nullable` annotation, still no null-checks at the 6 production callers (`DatabricksThriftAccessor:710`, `DatabricksTokenFederationProvider`, `AzureMSICredentialProvider:43`, `PrivateKeyClientCredentialProvider`, `DatabricksDriverFeatureFlagsContext`, `ArrowStreamResult:53/153`). For unregistered UUIDs the new design passes through (good — that's what fixed the CI). For registered-then-closed UUIDs the silent-null storage is still possible if any non-TELEMETRY caller reaches `getClient()` after `closeConnection()` — e.g. `ArrowStreamResult` mid-fetch when the connection closes concurrently, or a lazy auth refresh during close. Narrower window than the test scenario, but the production hole is unchanged.

**Suggested single fix that closes both:** encode the closed marker IN the existing `instances` map instead of a parallel set. On `closeConnection()`, replace the value at each `(uuid, type)` key with a tombstone — either `Optional.empty()` wrapping (and unwrap in `getClient`), or a rejecting sentinel `IDatabricksHttpClient` whose `execute()` throws `ConnectionClosedException` (symmetric to `NoopTelemetryClient`). Then:
- `getClient`'s `computeIfAbsent` naturally returns the tombstone for closed connections.
- No `everRegistered` set → no unbounded growth.
- Sentinel return → no null contract → all 6 callers are safe by default, no `@Nullable` retrofit needed.
- Bounded by live `(uuid, type)` pairs, scales with concurrent connections rather than JVM-lifetime close count.

If sticking with the two-set design, at minimum: fix the inaccurate Javadoc on `openConnections`, and bound `everRegistered` (Caffeine `expireAfterWrite(5 min)` — race window is seconds, not JVM lifetime).

gopalldb: Concern 1 (F2 — unbounded everRegistered): Eliminated. Both openConnections and everRegistered sets are gone. DatabricksHttpClientFactory now uses tombstones in the instances map itself — closeConnection()
  replaces real clients with ClosedConnectionHttpClient.INSTANCE. No parallel sets, no unbounded growth. Bounded by live (uuid, type) pairs.

  Concern 2 (F3 — nullable getClient()): Eliminated. getClient() never returns null. computeIfAbsent returns the tombstone sentinel for closed connections (key already exists).
  ClosedConnectionHttpClient.execute() throws DatabricksHttpException("Connection has been closed"). All 6 production callers are safe by default.

  TelemetryClientFactory was also simplified — replaced the two-set allowlist with a single closedConnectionUuids denylist. registerConnection() removed from both factories.

vikrantpuppala (approved):

PR #1351, opened 2026-03-27
Surface truncated flag from SEA result manifest

tom-s-powell:
## Description
When using SEA it would be helpful to understand whether results have been truncated to ensure we don't silently consume incomplete results. Propagating this flag into `DatabricksResultSetMetaData` similar to `isCloudFetchUsed` would be helpful.

I guess in theory the JDBC driver could fail loudly if results are truncated, but appreciate that's a behaviour change and perhaps use of SEA is expected to match behaviour via the API and it's a user's responsibility to behave appropriately.

## Testing
Unit tests have been added.

NO_CHANGELOG=true


gopalldb (approved):

gopalldb:
This PR has been merged as: https://github.com/databricks/databricks-jdbc/pull/1503

PR #1400, opened 2026-04-10
Fix statement close behavior for direct results in SEA

gopalldb:
## Summary
For direct results in SEA, the server returns CLOSED state with inline results. Previously, `markAsClosed()` set `isClosed=true` on the JDBC Statement, preventing re-execution and `getResultSet()` calls. Per JDBC spec, a Statement should remain open until the user explicitly calls `close()`.

## Problem
- `executeQuery()` then `getResultSet()` threw "Statement is closed"
- Re-executing a Statement/PreparedStatement after direct results threw "Statement is closed"
- `getExecutionResult()` (async path) made an RPC that returned "not found" since the server already closed the operation
- PreparedStatement parameters were lost (couldn't re-execute with new params)

## Fix
Replace `markAsClosed()` with `markDirectResultsReceived()` which sets a new `directResultsReceived` flag WITHOUT closing the JDBC Statement:

- **Statement remains open** for re-execution after direct results
- **Re-execution implicitly closes the previous ResultSet** (per JDBC spec) and resets the flag
- **`close()`** skips server-side `closeStatement` when `directResultsReceived` (server already cleaned up)
- **`getExecutionResult()`** returns cached result for direct results instead of an RPC
- **PreparedStatement params stay sticky** across re-executions (unchanged, already correct)

## Test plan
- [x] `DatabricksStatementTest` -- 85 tests pass (3 updated for new behavior)
- [x] `DatabricksSdkClientTest` -- 28 tests pass (1 updated for method rename)
- [x] Full core suite -- 3188 tests, 0 failures


src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -144,7 +147,11 @@ public void close(boolean removeFromSession) throws DatabricksSQLException {
...
       warnings = WarningUtil.addWarning(warnings, warningMsg);
     } else {
-      this.connection.getSession().getDatabricksClient().closeStatement(statementId);
+      // Skip server-side close if the server already closed the operation (direct results).
+      // The operation handle is gone on the server side, so closeStatement would fail.
+      if (!directResultsReceived) {
+        this.connection.getSession().getDatabricksClient().closeStatement(statementId);
+      }
       if (resultSet != null) {
         this.resultSet.close();
```
vikrantpuppala: **[F1] `close()` leaks state if `resultSet.close()` throws** (Severity: High)

If `this.resultSet.close()` on line 156 throws, none of the following run: `connection.closeStatement(this)` (L161), `DatabricksThreadContextHolder.clearStatementInfo()` (L163), `shutDownExecutor()` (L166), or `isClosed = true` (L168). That leaves the statement referenced from `connection.statementSet`, a stale statementId in the ThreadLocal, executor threads leaked, and the statement in a state where subsequent `close()` calls re-enter the `isClosed` branch and re-throw without ever cleaning up.

With the `if (!directResultsReceived)` guard, `close()` is now the sole cleanup path for direct-result statements (pre-PR `markAsClosed` cleaned eagerly), so this pre-existing hazard is more reachable. The deleted `testMarkAsClosedThenCloseWithResultSetError` exercised close-throws on the already-closed path; its replacement doesn't call `close()`.

**Suggested fix:** Wrap `resultSet.close()` in try/finally so session cleanup, ThreadLocal clear, executor shutdown, and `isClosed = true` always run; rethrow the resultSet exception at the end.

*Flagged by: devils-advocate, test-reviewer, language-expert*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -834,6 +874,18 @@ DatabricksResultSet executeInternal(
       String sql, Map<Integer, ImmutableSqlParameter> params, StatementType statementType)
       throws SQLException {
     noMoreResults = false; // reset before each execution
+    directResultsReceived = false; // reset for new execution
```
vikrantpuppala: **[F2] Stale `statementId` after failed re-execution → spurious `closeStatement(old-id)` RPC** (Severity: High)

This reset block clears `noMoreResults`, `directResultsReceived`, and closes the previous `resultSet`, but does NOT clear `statementId`. If the prior execution took the direct-results path (statementId=ID-A, `directResultsReceived=true`) and `client.executeStatement(...)` throws before `setStatementId(newId)` is called, the statement is left in `{statementId: ID-A, directResultsReceived: false, isClosed: false}`.

A subsequent user `close()` then sends `closeStatement(ID-A)` against an already-server-closed handle → not-found error. Pre-PR this was unreachable because `markAsClosed()` set `isClosed=true`, so the second `execute()` would have been rejected.

**Suggested fix:** In this reset block, also add `statementId = null;`. On failure, `close()` then short-circuits cleanly via the `statementId == null` branch (L145); on success, the SDK/Thrift path repopulates it via `setStatementId`.

*Flagged by: security-reviewer, devils-advocate*

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -257,13 +259,29 @@ DatabricksResultSet execute(
...
+
+      // For direct results, the server operation is already complete — no further
+      // RPCs (getOperationStatus, FetchResults, closeOperation) will succeed for
+      // this operation handle. Mark the statement so close() skips the server call.
+      if (isDirectResults
+          && parentStatement != null
+          && parentStatement.getStatement() instanceof DatabricksStatement) {
+        LOGGER.debug(
+            "Statement {} received direct results via Thrift, marking as direct results received",
+            statementId);
```
vikrantpuppala, line 281: **[F3] Thrift marks direct-results without confirming server closed the op** (Severity: High)

`hasResultDataInDirectResults()` (L854-863) only checks `directResults.isSetResultSet() && isSetResultSetMetadata()`. The Thrift IDL declares `TSparkDirectResults.closeOperation` as **optional**:

```thrift
struct TSparkDirectResults {
  1: optional TGetOperationStatusResp operationStatus
  2: optional TGetResultSetMetadataResp resultSetMetadata
  3: optional TFetchResultsResp resultSet
  4: optional TCloseOperationResp closeOperation   // <-- optional
}
```
(`TCLIService.thrift:434-439`)

A server can return inline result data without closing the operation (older protocol versions, interactive flows). This block unconditionally calls `markDirectResultsReceived()` when inline data is present, making `close()` skip the `closeStatement` RPC — silently leaking the live server-side operation handle.

**Suggested fix:** Gate `markDirectResultsReceived()` on `response.getDirectResults().isSetCloseOperation()` (or equivalent op-closed predicate), not on presence of inline result data.

*Flagged by: ops, devils-advocate*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -834,6 +874,18 @@ DatabricksResultSet executeInternal(
...
+
+    // Per JDBC spec, re-executing a Statement implicitly closes the current ResultSet
+    if (resultSet != null) {
+      try {
+        resultSet.close();
+      } catch (SQLException e) {
```
vikrantpuppala: **[F4] Orphaned server operation handles on re-execution (non-direct-results path)** (Severity: High)

`executeInternal` closes the previous local `resultSet` but never calls `closeStatement(previousStatementId)` before `parentStatement.setStatementId(newId)` overwrites it in the client layer. For non-direct-results executions (SEA SUCCEEDED, Thrift no-direct), N re-executions leak N-1 server-side operation handles until connection close.

Pre-PR this was largely unreachable (the direct-results path made `isClosed=true`, so re-execution was blocked); this PR explicitly advertises re-execution as a supported flow, so the leak scenario becomes realistic.

**Suggested fix:** In this reset block, if `statementId != null && !directResultsReceived`, best-effort call `closeStatement(previousStatementId)` before the new execution. Coordinate with F2's `statementId = null` fix so both paths handle cleanup consistently.

*Flagged by: devils-advocate*

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -336,8 +334,10 @@ public DatabricksResultSet executeStatement(
             parentStatement);
 
     if (shouldMarkClosed) {
-      LOGGER.debug("Statement {} returned CLOSED status, marking statement as closed", statementId);
-      ((DatabricksStatement) parentStatement.getStatement()).markAsClosed();
+      LOGGER.debug(
+          "Statement {} returned CLOSED status with direct results, marking as direct results received",
+          statementId);
+      ((DatabricksStatement) parentStatement.getStatement()).markDirectResultsReceived();
```
vikrantpuppala: **[F5] Unchecked cast vs. `instanceof`-guarded Thrift path — inconsistent defensive coding** (Severity: Medium)

SDK path casts directly:
```java
((DatabricksStatement) parentStatement.getStatement()).markDirectResultsReceived();
```

but the Thrift path (`DatabricksThriftAccessor.java:275-282`) guards with:
```java
if (isDirectResults && parentStatement != null
    && parentStatement.getStatement() instanceof DatabricksStatement) { ... }
```

Only `DatabricksStatement`/`DatabricksPreparedStatement` implement `IDatabricksStatementInternal` today, so no live `ClassCastException`, but the asymmetry is a latent trap.

**Suggested fix:** Preferred — add `markDirectResultsReceived` to `IDatabricksStatementInternal` (as `default` no-op), and drop casts in both clients. Minimum — mirror the `instanceof` guard here in the SDK path.

*Flagged by: security, devils-advocate, architecture, language, maintainability*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -55,6 +55,9 @@ public class DatabricksStatement implements IDatabricksStatement, IDatabricksSta
   private final DatabricksBatchExecutor databricksBatchExecutor;
   private boolean noMoreResults = false; // JDBC end-of-results indicator
   private long updateCount = -1; // Update count for DML statements, -1 for SELECT or no results
+  private boolean directResultsReceived = false; // Server returned inline results and closed the
```
vikrantpuppala: **[F6] `directResultsReceived` not `volatile` — cross-thread `cancel()` may observe stale value** (Severity: Medium)

JDBC requires `Statement.cancel()` to be callable from another thread than the one executing the query. Without `volatile`/synchronization, the cancel thread may miss a `directResultsReceived=true` write (JMM visibility) and fire `cancelStatement` against an already-closed handle, or see a stale `true` after reset and silently no-op when it should cancel.

The class already has this hazard on `isClosed`/`statementId`/`resultSet` (pre-existing), but this PR widens the surface.

**Suggested fix:** At minimum, mark `directResultsReceived` `volatile`. Follow-up: synchronize full state transitions (`close`, `cancel`, `executeInternal`, `executeAsync`, `markDirectResultsReceived`) — tracked as a dedicated thread-safety ticket.

*Flagged by: security, devils-advocate, language*

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
vikrantpuppala: **[F7] Client → Statement abstraction leak via downcast** (Severity: Medium)

Both `DatabricksSdkClient.java:340` and `DatabricksThriftAccessor.java:275-282` reach up into `DatabricksStatement` via concrete-class downcast to call `markDirectResultsReceived()`. The method is not on `IDatabricksStatementInternal`, so the interface is lying about the contract.

The right direction is client → Statement via a return-value DTO (e.g., `DatabricksResultSet.isOperationClosed()` or a small `ExecuteOutcome { resultSet, serverOperationClosed }`), letting `executeInternal` set the private flag itself based on the returned state. No downcast needed, and a future transport that also produces direct results would just populate the bit.

**Suggested fix:** In this PR, at minimum add `markDirectResultsReceived` to `IDatabricksStatementInternal` (default method) so the interface reflects the contract. File a follow-up for the DTO refactor.

*Flagged by: architecture, security, maintainability*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -649,6 +661,19 @@ public InputStreamEntity getInputStreamForUCVolume() throws DatabricksSQLExcepti
...
     LOGGER.debug("ResultSet executeAsync() for statement {%s}", sql);
     checkIfClosed();
+
+    // Reset state for new execution — a previous sync execute() may have set
+    // directResultsReceived, which must be cleared before the new async execution
+    directResultsReceived = false;
+    if (resultSet != null) {
+      try {
+        resultSet.close();
+      } catch (SQLException e) {
```
vikrantpuppala: **[F8] Duplicated reset block — drift already starting** (Severity: Medium)

This 7-line reset block is near-identical to the one in `executeInternal` (L876-887). Both clear `directResultsReceived=false`, close the previous `resultSet`, null it.

But `executeInternal` additionally resets `noMoreResults` (L876) and `updateCount` (L906); this async path does neither. So a user who does `executeQuery` → `getMoreResults` (sets `noMoreResults=true`) → `executeAsync` → `getExecutionResult` will see stale `noMoreResults=true` on the new async-execution statement. The duplication is already hiding this divergence.

**Suggested fix:** Extract `private void resetForNewExecution()` and call from both sites. Decide explicitly whether async should reset `noMoreResults`/`updateCount` (almost certainly yes, for symmetry with sync).

*Flagged by: maintainability, architecture, language, devils-advocate*

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
vikrantpuppala, line 281: **[F9] Thrift direct-results marking path is untested** (Severity: Medium)

This new block (`instanceof DatabricksStatement` guard + `markDirectResultsReceived()` call) has zero test coverage. Existing Thrift tests that exercise direct results either pass `parentStatement = null` or a non-`DatabricksStatement` mock, so the branch is never entered.

**Suggested fix:** Add a test in `DatabricksThriftAccessorTest.java` with `parentStatement.getStatement()` returning a `DatabricksStatement` spy, direct results set, and `verify(stmt, times(1)).markDirectResultsReceived()`. Add a negative test with a non-`DatabricksStatement` parent to cover the defensive `instanceof` branch.

*Flagged by: test-reviewer*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
vikrantpuppala: **[F10] `executeAsync()` reset block is completely untested** (Severity: Medium)

This new 7-line block — clears `directResultsReceived`, closes previous `resultSet`, swallows `SQLException` on close — has zero test coverage.

**Suggested fix:** Add tests for:
1. sync `executeQuery` → `markDirectResultsReceived` → `executeAsync` closes the previous `resultSet`
2. previous `resultSet.close()` throwing is swallowed (does not fail `executeAsync`) and surfaces only at debug level
3. `directResultsReceived` flag is cleared after `executeAsync`

*Flagged by: test-reviewer*

src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
```diff
@@ -837,53 +837,132 @@ public void testMarkAsClosedThenCloseWithResultSetError() throws Exception {
...
+  @Test
+  public void testCancelAfterDirectResultsIsNoOp() throws Exception {
+    IDatabricksConnectionContext connectionContext =
+        DatabricksConnectionContext.parse(JDBC_URL, new Properties());
+    DatabricksConnection connection = new DatabricksConnection(connectionContext, client);
+    DatabricksStatement statement = new DatabricksStatement(connection);
+
+    when(client.executeStatement(
+            eq(STATEMENT),
+            eq(new Warehouse(WAREHOUSE_ID)),
```
vikrantpuppala, line 920: **[F11] Lost coverage: `close()` exception propagation from `resultSet.close()`** (Severity: Medium)

The pre-PR test `testMarkAsClosedThenCloseWithResultSetError` verified that `close()` propagates exceptions from `resultSet.close()`. Its replacement `testDirectResultsDoesNotCloseResultSet` never calls `close()` at all. The behavior at `DatabricksStatement.java:156` still depends on that propagation, but is no longer asserted.

Directly related to finding F1 (close exception-safety).

**Suggested fix:** Add a test that (a) calls `markDirectResultsReceived()`, (b) uses a mock `resultSet` throwing on `close()`, (c) calls `statement.close()`, (d) asserts the exception propagates, AND (e) verifies `client.closeStatement(...)` was NOT called. Pair with F1's try/finally fix so both the throw propagation AND the cleanup-on-throw are covered.

*Flagged by: test-reviewer, devils-advocate*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -144,7 +147,11 @@ public void close(boolean removeFromSession) throws DatabricksSQLException {
       LOGGER.warn(warningMsg);
       warnings = WarningUtil.addWarning(warnings, warningMsg);
     } else {
-      this.connection.getSession().getDatabricksClient().closeStatement(statementId);
+      // Skip server-side close if the server already closed the operation (direct results).
+      // The operation handle is gone on the server side, so closeStatement would fail.
+      if (!directResultsReceived) {
+        this.connection.getSession().getDatabricksClient().closeStatement(statementId);
+      }
```
vikrantpuppala: **[F12] No observability for direct-results close path (telemetry + counter gap)** (Severity: Medium)

`@DatabricksMetricsTimed` instruments the RPC call; skipping `closeStatement`/`cancelStatement` RPCs on direct results silently drops per-statement latency telemetry flush (see `TelemetryCollector.java:73-78,116-124`) and the existing close/cancel counters. There is also no counter for "closed via direct results".

Consequences: (a) if F3/F4 cause handle leaks in prod, there is no driver-side signal; (b) existing metric distributions shift silently after release; (c) the only diagnostic is `LOGGER.debug` which is invisible at the default log level (`DatabricksJdbcConstants.DEFAULT_LOG_LEVEL = OFF`).

**Suggested fix:** (1) Emit a telemetry counter (e.g., `STATEMENT_CLOSED_VIA_DIRECT_RESULTS`) from `markDirectResultsReceived()`. (2) In `close()`'s direct-results branch, explicitly trigger `exportTelemetryDetailsAndClear(statementId)` so per-statement latency telemetry still flushes.

*Flagged by: ops*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -229,9 +236,14 @@ public void cancel() throws SQLException {
...
     checkIfClosed();
 
-    if (statementId != null) {
+    if (statementId != null && !directResultsReceived) {
       this.connection.getSession().getDatabricksClient().cancelStatement(statementId);
       DatabricksThreadContextHolder.clearStatementInfo();
+    } else if (directResultsReceived) {
+      LOGGER.debug(
+          "Skipping cancel for statement {} — direct results already received, "
+              + "server operation closed",
```
vikrantpuppala: **[F13] `cancel()` on direct-results statement silently debug-logs instead of warning** (Severity: Medium)

`cancel()` is a legal JDBC call. The sibling `statementId == null` branch adds a SQL warning via `WarningUtil.addWarning(...)`, but this new branch silently debug-logs and returns — so user code calling `getWarnings()` after `cancel()` sees no signal that the cancel was a no-op.

**Suggested fix:** Add a warning via `WarningUtil.addWarning(...)` here for parity: "statement's server operation was already closed; cancel has no effect".

*Flagged by: architecture*

vikrantpuppala:
## [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) Report

**Scope**: 6 files, +228 / -75 lines
**Merge Safety Score**: **60 / 100 — HIGH RISK**
**Reviewers**: security, devils-advocate, language-expert, architecture, test, maintainability, operational-safety

### Executive Summary

The JDBC-spec-compliance fix (Statement stays open after direct results) is correct and the `markAsClosed → markDirectResultsReceived` rename is complete with zero stale callers. However, the enlarged re-execution surface exposes two newly-reachable regressions (stale `statementId` / orphaned server handles on re-exec), an ungated Thrift direct-results signal that can leak operation handles against older servers that omit `closeOperation`, a `close()` exception-safety hole that skips cleanup if `resultSet.close()` throws, and missing coverage for the Thrift marking path and `executeAsync` reset.

### Findings by Severity

- **High (4)**: F1 close() exception-safety, F2 stale `statementId` after failed re-exec, F3 Thrift unchecked `closeOperation`, F4 orphaned handles on non-direct re-exec
- **Medium (9)**: F5 unchecked SDK cast, F6 non-volatile flag, F7 client→statement abstraction leak, F8 duplicated reset block, F9 Thrift marking untested, F10 executeAsync reset untested, F11 lost close-throw coverage, F12 missing observability/telemetry, F13 cancel() silent no-op
- **Low (9)**: see list below

Each High + Medium finding is posted as an inline comment at its file:line.

### Low Findings (no inline — batched here)

- **F14** Redundant import — `DatabricksThriftAccessor.java:8` duplicates the wildcard on L7. Delete L8.
- **F15** Stale local name `shouldMarkClosed` at `DatabricksSdkClient.java:324` now drives "mark direct results received" → rename to `hasDirectResults`.
- **F16** Inconsistent `resultSet.close()` error handling — `close(boolean)` rethrows, `getMoreResults` silently ignores, new `executeInternal/executeAsync` debug-log. Follow-up to normalize.
- **F17** Local `DatabricksResultSet databricksResultSet` in `DatabricksThriftAccessor.execute` shadows a nearby Thrift-typed `resultSet` — rename to `jdbcResultSet`.
- **F18** Field comment style — 3-line `//` trailing comment on `directResultsReceived` field should be a Javadoc `/** */`.
- **F19** Inconsistent LOGGER.debug wording across SEA/Thrift paths — normalize if F7's shared helper is introduced.
- **F20** `testReExecutionClosesPreviousResultSet` doesn't verify flag reset — after the 2nd `executeQuery`, call `close()` and verify `closeStatement` IS called.
- **F21** `getExecutionResult` "no result set available" throw branch (`DatabricksStatement.java:705-708`) is unreachable in prod but untested.
- **F22** No e2e coverage for SEA CLOSED-state direct-results path — regression prevention against the exact bug this PR fixes.

### Positively Verified

- `markAsClosed` rename is fully clean — zero stale callers in src/ or tests.
- PreparedStatement param stickiness across re-execution holds (params survive; `executeInternal` doesn't touch `databricksParameterMetaData`).
- Thrift metadata operations (getTables, getColumns, etc.) don't go through `execute()` — no risk of mis-marking metadata statements.
- SEA async path correctly opts out of direct-results marking, with an explicit clarifying comment.

### Score Breakdown

| Category | Count | Deductions |
|---|---|---|
| Critical | 0 | 0 |
| High | 4 | -40 |
| Medium | 9 | -27 |
| Low | 9 | -9 |
| **Total** | 22 | **64 → floored at 60 / 100** |

---
*Feedback? Drop it in [#code-review-squad-feedback](https://databricks.slack.com/channels/code-review-squad-feedback).*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: fixed. by moving the cleanup logic to finally block

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: addressed

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
gopalldb: Fixed

gopalldb: Fixed

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
gopalldb: Fixed. Added `markDirectResultsReceived()` as a `default` no-op method on `IDatabricksStatementInternal`. Both SDK and Thrift clients now call `parentStatement.markDirectResultsReceived()` directly — no casts, no `instanceof`. Removed unused `DatabricksStatement` import from `DatabricksThriftAccessor`.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed. Changed to `private volatile boolean directResultsReceived`. Added comment explaining volatility is needed because `cancel()` can be called from a different thread (JDBC spec requirement).

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
gopalldb: Fixed in this PR by adding `markDirectResultsReceived()` to `IDatabricksStatementInternal` as a default method. Both clients now use the interface — no downcasts. Filed follow-up for the DTO refactor (`ExecuteOutcome { resultSet, serverOperationClosed }`) to fully decouple the transport layer from Statement internals.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed. Extracted `private void resetForNewExecution()` called from both `executeInternal` and `executeAsync`. Resets: `noMoreResults`, `updateCount`, closes previous server handle (F4), clears `directResultsReceived`, closes previous ResultSet, nulls `statementId` (F2). Both paths now have identical reset semantics.

gopalldb: Fixed. Added `testAsyncExecutionResetsStateFromPreviousSyncExecution` — sync execute with direct results → async execute → verifies previous ResultSet closed. The `resetForNewExecution()` refactor (F8) ensures both paths have identical behavior.

src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
gopalldb: Fixed. Added `testCloseAfterDirectResultsPropagatesResultSetCloseException` — verifies exception propagates from `resultSet.close()`, `closeStatement` is NOT called (direct results), AND `isClosed=true` (top-level finally runs). This covers both the throw propagation and the cleanup-on-throw behavior from F1.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed. Upgraded `markDirectResultsReceived()` and the `close()` skip path from debug to info-level logging. When direct results cause `close()` to skip the closeStatement RPC, the info log provides visibility: `"Statement {} closed locally (direct results — server operation already closed, skipping closeStatement RPC)"`. Default log level is OFF, but when enabled at INFO+, this is now visible.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed. Added `WarningUtil.addWarning()` for parity with the `statementId == null` branch: `"Statement's server operation was already closed (direct results); cancel has no effect."` Added `testCancelAfterDirectResultsAddsWarning` verifying `getWarnings()` is non-null and contains "already closed".

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -134,31 +138,44 @@ public void close() throws SQLException {
...
+          this.resultSet = null;
+        }
+      } else if (statementId == null) {
+        String warningMsg = "The statement you are trying to close does not have an ID yet.";
+        LOGGER.warn(warningMsg);
+        warnings = WarningUtil.addWarning(warnings, warningMsg);
+      } else {
+        // Skip server-side close if the server already closed the operation (direct results).
+        // The operation handle is gone on the server side, so closeStatement would fail.
+        if (!directResultsReceived) {
```
vikrantpuppala, line 154: can we just move the first if block if(isClosed) here for simplicity since both try to close the result set. we could simply make this if (!isClosed && !directResultsReceived)

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -134,31 +138,44 @@ public void close() throws SQLException {
...
+          this.connection.getSession().getDatabricksClient().closeStatement(statementId);
+        } else {
+          LOGGER.info(
```
vikrantpuppala: Does this even need a log? especially at the info level?

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -336,8 +334,10 @@ public DatabricksResultSet executeStatement(
             parentStatement);
 
     if (shouldMarkClosed) {
```
vikrantpuppala: should we update this variable name to shouldMarkDirectResults?

vikrantpuppala (commented):
can we add a test in DatabricksPreparedStatementTest that ensures common prepared statement scenarios are working well (they were broken before this PR)
- getMetaData() should now work via the cached resultSet path
- bind and execute multiple times using the same statement should work, etc.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Good suggestion. However, the current structure preserves an important distinction: when `isClosed` is true and `resultSet != null`, we ONLY close the result set (no session removal, no ThreadLocal clear — those were already done on the first close). When `!isClosed`, we do the full cleanup (server close, session removal, ThreadLocal clear). Merging them would make the `isClosed` branch run the full cleanup path unnecessarily. That said, the `resultSet.close()` call in the `isClosed` branch is a defensive cleanup for an edge case (result set somehow not null after prior close). I'll add a comment explaining why they're separate.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Downgraded to debug

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
gopalldb: Good catch — updated to `shouldMarkDirectResults` for clarity.

gopalldb:
Added 4 tests in `DatabricksPreparedStatementTest` covering the scenarios that were broken before this PR:

1. **`testGetMetaDataWorksAfterDirectResults`** — `executeQuery()` → direct results → `getResultSet()` + `getMetaData()` work (previously threw "Statement is closed")
2. **`testBindAndExecuteMultipleTimesWithSamePreparedStatement`** — bind → execute → direct results → rebind → re-execute × 2 (previously threw on re-execution)
3. **`testGetResultSetWorksAfterDirectResults`** — `execute()` → direct results → `getResultSet()` works
4. **`testParametersClearedOnReExecution`** — `clearParameters()` → rebind → re-execute after direct results

All 52 PreparedStatement tests pass (48 existing + 4 new).

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

vikrantpuppala (approved):
thank you!

PR #1403, opened 2026-04-10
Fix batched insert regex to support backtick-quoted identifiers with special characters

sreekanth-db:
## Summary
- Fixed `INSERT_DETAILS_PATTERN` regex in `InsertStatementParser` to support backtick-quoted identifiers containing special characters (hyphens, spaces, escaped backticks, etc.) in table/schema/catalog names
- Added a `warn` log in `PreparedStatementBatchExecutor.canUseBatchedInsert()` when parsing fails and the driver falls back to individual execution
- Added unit tests for the new regex behavior

## Root Cause
The regex capture group for the table name matched characters individually, so backtick-quoted identifiers with special characters failed to match. The fix changes the group to match whole backtick-quoted segments — allowing any character inside, including escaped backtick pairs — or plain identifiers, separated by dots.

## Test plan
- [x] All `InsertStatementParserTest` tests pass
- [x] Verified against real Databricks warehouse with various table name formats
- [x] Warn log confirmed to appear when fallback occurs

Fixes https://github.com/databricks/databricks-jdbc/issues/1398


src/main/java/com/databricks/jdbc/common/util/InsertStatementParser.java
```diff
@@ -18,9 +18,11 @@
 public class InsertStatementParser {
 
   // Pattern to extract table and columns from INSERT INTO table (col1, col2, ...) VALUES format
+  // Table name group matches dot-separated segments where each segment is either a
+  // backtick-quoted identifier (allowing any character inside) or an unquoted identifier (\w+).
   private static final Pattern INSERT_DETAILS_PATTERN =
       Pattern.compile(
-          "^\\s*INSERT\\s+INTO\\s+([\\w`\\.]+)\\s*\\(([^)]+)\\)\\s+VALUES\\s*\\(",
+          "^\\s*INSERT\\s+INTO\\s+((?:`[^`]+`|\\w+)(?:\\.(?:`[^`]+`|\\w+))*)\\s*\\(([^)]+)\\)\\s+VALUES\\s*\\(",
```
cliffsun-anaplan: I saw this PR got linked to the issue I raised - thanks for this, but I think there's an edge case that is missing here which is when the schema / table name itself contains backticks (which would have been escaped by another backtick), e.g. ```table-with-``-backticks```. When I tested it initially I iterated through with Claude and ended up with the first capture group being ```((?:`(?:[^`]|``)+`|\w+)(?:\.(?:`(?:[^`]|``)+`|\w+))*)``` which is very similar to what you have but just also is able to match on the double backticks inside of a quoted identifier.

sreekanth-db: Thanks for pointing this out! Updated the regex to handle escaped backticks as well.

o-shevchenko:
@sreekanth-db
Will it be a part of the next release?
What is the release date?
Thanks

vikrantpuppala (approved):

madhav-db (approved):

msrathore-db (approved):
LGTM

sreekanth-db:
@copilot resolve the merge conflicts in this pull request

PR #1404, opened 2026-04-10
Fix batch inserts failing with Statement is closed on SEA client

sreekanth-db:
## Summary
- Fixed batch inserts failing with "Statement is closed" when using SEA client (`useThriftClient=0`) for multi-chunk batches or individual fallback execution
- Each batch chunk/row execution now uses a fresh `DatabricksStatement` instead of reusing the parent `PreparedStatement`
- Removed the `StatementExecutor` interface — batch executor now handles execution internally
- Update count is extracted before closing the statement to avoid "ResultSet is closed" errors

## Root Cause
The SEA server returns `StatementState.CLOSED` after DML execution. `DatabricksSdkClient` reacts by calling `markAsClosed()` on the Java statement, setting `isClosed = true`. When the batch executor tried to execute the next chunk/row on the same statement, `checkIfClosed()` threw "Statement is closed". The Thrift client does not call `markAsClosed()` after DML, which is why this only affected SEA.

## Approach
Instead of modifying the shared statement lifecycle code (`markAsClosed`, `checkIfClosed`, `isClosed`), each batch execution now creates a fresh `DatabricksStatement`. This is safe because:
- Thrift already creates new server-side statements per execution — this aligns the Java side
- Each statement has an independent lifecycle with proper cleanup via `try/finally`
- Zero changes to `DatabricksStatement`, `DatabricksSdkClient`, or `markAsClosed`

## Test plan
- [x] All 48 `DatabricksPreparedStatementTest` tests pass (mocks updated for batch tests only)
- [x] Verified against real Databricks warehouse — multi-chunk batched path and individual fallback both succeed with both Thrift and SEA clients


vikrantpuppala:
Let's hold off on this as #1400 should address this?

sreekanth-db:
https://github.com/databricks/databricks-jdbc/pull/1400 fixes this - closing PR

PR #1407, opened 2026-04-14
Fix race condition between chunk release and download error handling

### LINKED ISSUE #1367: [BUG] [Arrow cloud Fetch] Race condition between CHUNK_RELEASED and DOWNLOAD_FAILED causes unrecoverable state
**Describe the bug**
 Arrow Cloud Fetch result streaming fails with a `DatabricksParsingException: Invalid state transition` when the `IdleConnectionEvictor` background thread closes the HTTP connection pool while a chunk download is in progress. 

The exception surfaces as Premature end of `Content-Length `delimited message body on the download, followed by an invalid state machine transition attempt (CHUNK_RELEASED -> DOWNLOAD_FAILED) in the retry path, which is a terminal state with no valid outgoing transitions. 

This causes the entire result stream to fail non-recoverably even though the driver attempts an internal retry.

**To Reproduce**
  1. Use the Databricks JDBC driver with Arrow Cloud Fetch enabled (EnableArrow=1) and `IdleHttpConnectionExpiry` not set (driver default applies)
  2. Execute a query returning a large result set split into multiple Arrow chunks (e.g. a wide table with 100+ float columns, 100k+ rows)
  3. Apply resource constraints that slow down Arrow chunk downloads (e.g. 0.5 CPU, 1GB RAM container)
  4. Observe that a chunk download is still in progress when the IdleConnectionEvictor fires (thrift connection idle since Arrow Cloud Fetch resolves URLs upfront and then downloads data via separate connections)

  The evictor fires because:
  - After the driver receives Arrow chunk URLs from the Databricks warehouse, the thrift connection becomes idle
  - The evictor's idle threshold (driver default) expires during the background downloads
  - `connectionManager.closeIdleConnections()` closes the shared HTTP pool mid-download
 
**Expected behavior**
The driver should either:
  1. Not close the HTTP pool while Arrow chunk downloads are in progress, or
  2. Handle the CHUNK_RELEASED -> DOWNLOAD_FAILED transition gracefully -> this is a valid race condition where the consumer thread releases a chunk concurrently with the download thread attempting to mark it failed. The state machine should allow this transition rather than throwing an unrecoverable exception.

**Client side logs**
```
  Retry attempt 1 for chunk index: 7, Error: Data parsing failed for chunk index [7] and statement [<id>].
  Exception [com.databricks.internal.apache.http.ConnectionClosedException:
    Premature end of Content-Length delimited message body (expected: 5,439,615; received: 5,439,488)]

  Failed to transition to state [DOWNLOAD_FAILED] from state [CHUNK_RELEASED] for chunk [7]
  and statement [<id>]. Stack trace:
  com.databricks.jdbc.exception.DatabricksParsingException: Invalid state transition for chunk [7]
  and statement [<id>]: CHUNK_RELEASED -> DOWNLOAD_FAILED. Valid transitions from CHUNK_RELEASED are: []
      at com.databricks.jdbc.api.impl.arrow.ArrowResultChunkStateMachine.transition(ArrowResultChunkStateMachine.java:37)
      at com.databricks.jdbc.api.impl.arrow.AbstractArrowResultChunk.setStatus(AbstractArrowResultChunk.java:245)
      at com.databricks.jdbc.api.impl.arrow.ArrowResultChunk.handleFailure(ArrowResultChunk.java:157)
      at com.databricks.jdbc.api.impl.arrow.ArrowResultChunk.downloadData(ArrowResultChunk.java:130)
      at com.databricks.jdbc.api.impl.arrow.ChunkDownloadTask.call(ChunkDownloadTask.java:71)
```

**Root cause (from source analysis):**

In `ChunkDownloadTask.java`, the finally block unconditionally calls `chunk.setStatus(ChunkStatus.DOWNLOAD_FAILED)` without checking whether the chunk has already been released by the consumer thread. 

The `CHUNK_RELEASED ` state has no valid outgoing transitions in `ChunkStatus.VALID_TRANSITIONS`, making `CHUNK_RELEASED` -> `DOWNLOAD_FAILED `an illegal transition even though it is a legitimate race condition in concurrent execution.

The specific race window:

Consumer thread:   PROCESSING_SUCCEEDED -> releaseChunk() -> CHUNK_RELEASED
Download thread:   ConnectionClosedException -> finally: setStatus(DOWNLOAD_FAILED) <- INVALID

**Workaround:**

Set `IdleHttpConnectionExpiry` to a value larger than the expected maximum chunk download time (e.g. 600 seconds) to prevent the evictor from firing during active Arrow downloads:

`jdbc:databricks://<host>:443;EnableArrow=1;IdleHttpConnectionExpiry=600`

 - OS: Linux (container, linux/amd64)
 - Java version: Java 17.0.18
 - Java vendor: OpenJDK (FIPS - 17.0.18-internal with BouncyCastle FIPS)
 - Driver Version: 3.3.1 (com.databricks:databricks-jdbc:3.3.1)
 - BI Tool: Custom gRPC service (Spring Boot + jOOQ fetchLazy() + Reactor Flux.fromIterable(cursor))

**Additional context**
  - The bug is Arrow Cloud Fetch specific. JSON result format (EnableArrow=0) is not affected as it streams synchronously through the thrift connection with no background download threads and no state machine.
  - The `ChunkDownloadTask.finally` block should guard the state transition: only attempt `setStatus(DOWNLOAD_FAILED)` if the chunk is not already in CHUNK_RELEASED state, or alternatively CHUNK_RELEASED -> DOWNLOAD_FAILED should be added as a valid transition in ChunkStatus.VALID_TRANSITIONS.
  - The internal retry mechanism (DOWNLOAD_RETRY) is also rendered ineffective by this bug — after the invalid transition attempt fails, the chunk remains in CHUNK_RELEASED with no path to retry.

sreekanth-db:
## Description

Fixes #1367 — Race condition between `CHUNK_RELEASED` and `DOWNLOAD_FAILED` causes invalid state transition warnings during Arrow Cloud Fetch operations.

### Root Cause

When chunk download threads are running in parallel and their connections are cut off by `IdleConnectionEvictor` (due to no socket-level data movement caused by extreme resource pressure in low-resource environments), the downloads fail. When the consumer reaches a failed chunk, `rs.next()` throws a `DatabricksSQLException`. This can trigger `rs.close()` — either explicitly by the user's error handling code, implicitly by try-with-resources, or automatically by frameworks that close resources on error. Additionally, `rs.close()` can also be called independently by the user before consuming all data.

In either case, `doClose()` calls `shutdownNow()` on the download thread pool and releases all chunks to `CHUNK_RELEASED`. But `shutdownNow()` only sends an interrupt signal — it doesn't wait for threads to stop. Download threads that are still processing their errors may try to set `DOWNLOAD_FAILED` after their chunks have already been released. The transition `CHUNK_RELEASED → DOWNLOAD_FAILED` is invalid in the chunk lifecycle state machine, causing the error/warning.

### Race Condition Sequence

```
Thread A (consumer/main)              Thread B (download)
─────────────────────────             ─────────────────────
                                      downloading chunk [7]...
                                      connection killed by evictor
                                      IOException caught
                                      retries exhausted
                                      ↓
rs.next() → gets error                (processing error...)
rs.close()
  → doClose()
    → shutdownNow()                   (interrupt sent, but not yet processed)
    → chunk[7].releaseChunk()
      → CHUNK_RELEASED ✓
                                      → setStatus(DOWNLOAD_FAILED)
                                        CHUNK_RELEASED → DOWNLOAD_FAILED ✗
                                        Invalid transition!
```

### Fix

Added `awaitTermination(3, TimeUnit.SECONDS)` after `shutdownNow()` in both `RemoteChunkProvider.doClose()` and `StreamingChunkProvider.close()`. This ensures download threads complete their error handling path (catch → finally → setStatus) before chunks are released.

The 3-second timeout is a conservative upper bound. After `shutdownNow()` interrupts threads, they exit their retry sleep (`Thread.sleep(1500)`) immediately via `InterruptedException` and process the error path in milliseconds. `awaitTermination()` returns as soon as all threads finish — it does not wait the full 3 seconds.

### Files Changed

| File | Change |
|------|--------|
| `RemoteChunkProvider.java` | Added `awaitTermination(3s)` after `shutdownNow()` in `doClose()`, added INFO log for close lifecycle |
| `StreamingChunkProvider.java` | Added `awaitTermination(3s)` after `shutdownNow()` in `close()` |
| `NEXT_CHANGELOG.md` | Added changelog entry |

### Reproduction (Docker + iptables)

Reproduced the error/warning logs from issue #1367 using:
1. Docker container with network throttled to 5mbit (slow chunk downloads)
2. `iptables REJECT --tcp-reset` to kill connections mid-download (simulating `IdleConnectionEvictor` behavior)
3. Calling `rs.close()` while download threads process errors
4. Temporary `MAX_RETRIES=0` + uninterruptible sleep before `setStatus(DOWNLOAD_FAILED)` to widen the race window

**Before fix** — driver logs showed:
```
WARNING: Invalid state transition for chunk [1]: CHUNK_RELEASED -> DOWNLOAD_FAILED
WARNING: Invalid state transition for chunk [2]: CHUNK_RELEASED -> DOWNLOAD_FAILED
```

**After fix** — no invalid state transitions. `awaitTermination()` waited for threads to finish before releasing chunks.

Verified for both `RemoteChunkProvider` and `StreamingChunkProvider`.

- The `IdleConnectionEvictor` shares a connection pool between the Thrift transport and chunk downloads (`DatabricksHttpClient` via `DatabricksHttpClientFactory`). In resource-constrained environments, CPU starvation causes the Java layer to stop reading from sockets, making active download connections appear idle to the evictor.
- `shutdownNow()` is best-effort per Java docs — it cannot guarantee threads will stop. The interrupt flag is set, but threads only respond when they hit an interruptible operation (`Thread.sleep`, I/O). Code after the interrupt (catch blocks, finally blocks, `setStatus()` calls) is pure computation that doesn't check interrupts.
- The same race pattern existed in both `RemoteChunkProvider` (standard Cloud Fetch) and `StreamingChunkProvider` (opt-in via `EnableStreamingChunkProvider`). Both are fixed.
- `RemoteChunkProviderV2` (incubator) downloads chunks sequentially without a thread pool, so it is not affected.


src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
```diff
@@ -125,8 +129,23 @@ public void downloadNextChunks() {
...
+    // Wait for download threads to finish error handling before releasing chunks.
+    // After shutdownNow() interrupts threads, they exit their retry sleep and process
+    // the error path (catch → finally → setStatus) in milliseconds. 3 seconds is a
+    // conservative upper bound to avoid racing with that error handling path.
+    try {
+      if (!chunkDownloaderExecutorService.awaitTermination(3, TimeUnit.SECONDS)) {
+        LOGGER.warn("Download threads did not terminate within timeout");
+      }
+    } catch (InterruptedException e) {
+      LOGGER.error(e, "Interrupted while waiting for download threads to terminate");
```
vikrantpuppala: **[F1] Missing `Thread.currentThread().interrupt()` after catching `InterruptedException`** (Severity: High)

`close()` runs on a user JDBC thread (`DatabricksStatement.close` → `ResultSet.close` → `ChunkProvider.doClose`). Catching `InterruptedException` here without restoring the interrupt flag silently clears the signal for the rest of that thread — any downstream cancellation check won't see it. The rest of this codebase follows the re-interrupt convention (see `ChunkDownloadTask.java:108`, `AbstractArrowResultChunk.java:293`).

**Suggested fix:**
```java
} catch (InterruptedException e) {
  LOGGER.error(e, "Interrupted while waiting for download threads to terminate");
  Thread.currentThread().interrupt();
}
```

Apply the same fix at `StreamingChunkProvider.java:288-290`.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
```diff
@@ -276,6 +277,17 @@ public void close() {
...
+      // Wait for download threads to finish error handling before releasing chunks.
+      // After shutdownNow() interrupts threads, they exit their retry sleep and process
+      // the error path (catch → finally → setStatus) in milliseconds. 3 seconds is a
+      // conservative upper bound to avoid racing with that error handling path.
+      try {
+        if (!downloadExecutor.awaitTermination(3, TimeUnit.SECONDS)) {
+          LOGGER.warn("Download threads did not terminate within timeout");
+        }
+      } catch (InterruptedException e) {
+        LOGGER.error(e, "Interrupted while waiting for download threads to terminate");
```
vikrantpuppala, line 289: **[F1] Missing `Thread.currentThread().interrupt()` after catching `InterruptedException`** (Severity: High)

Same issue as the corresponding spot in `RemoteChunkProvider` — the interrupt flag is dropped on the JDBC caller thread.

**Suggested fix:**
```java
} catch (InterruptedException e) {
  LOGGER.error(e, "Interrupted while waiting for download threads to terminate");
  Thread.currentThread().interrupt();
}
```

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
```diff
@@ -125,8 +129,23 @@ public void downloadNextChunks() {
...
+        chunkIndexToChunksMap.size(),
+        Thread.currentThread().getName());
     isClosed = true;
     chunkDownloaderExecutorService.shutdownNow();
+    // Wait for download threads to finish error handling before releasing chunks.
+    // After shutdownNow() interrupts threads, they exit their retry sleep and process
+    // the error path (catch → finally → setStatus) in milliseconds. 3 seconds is a
+    // conservative upper bound to avoid racing with that error handling path.
+    try {
+      if (!chunkDownloaderExecutorService.awaitTermination(3, TimeUnit.SECONDS)) {
```
vikrantpuppala: **[F3] Race still fires when `awaitTermination` times out** (Severity: High)

When `awaitTermination(3, TimeUnit.SECONDS)` returns `false`, control falls through to `chunkIndexToChunksMap.values().forEach(ArrowResultChunk::releaseChunk)` on line 149. A download thread still mid-flight will eventually reach its `finally` block at `ChunkDownloadTask.java:130` (`chunk.setStatus(ChunkStatus.DOWNLOAD_FAILED)`) — which is the exact `CHUNK_RELEASED → DOWNLOAD_FAILED` invalid transition this PR aims to prevent.

Apache HttpClient classic blocking I/O doesn't honor `Thread.interrupt()` on socket reads — a thread stuck on `SocketInputStream.read()` only exits when the socket timeout (default ~900s on this driver) fires. Under the resource-pressure scenario from issue #1367, this is not hypothetical.

**Suggested fix (pick one, in order of preference):**
1. **Root-cause fix**: make `setStatus(DOWNLOAD_FAILED)` a no-op when current status is `CHUNK_RELEASED`. The race then goes away regardless of the timeout value.
2. **Guard release**: in `doClose()`, skip `releaseChunk` for chunks whose status is still `DOWNLOAD_IN_PROGRESS` / `URL_FETCHED` and let the download thread's `finally` handle cleanup.
3. **Accept as mitigation**: land this PR as a window-narrowing fix and open a follow-up ticket for the state-machine change. The warn log gives operators a signal.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
```diff
@@ -276,6 +277,17 @@ public void close() {
     // Shutdown download executor
     if (downloadExecutor != null) {
       downloadExecutor.shutdownNow();
+      // Wait for download threads to finish error handling before releasing chunks.
+      // After shutdownNow() interrupts threads, they exit their retry sleep and process
+      // the error path (catch → finally → setStatus) in milliseconds. 3 seconds is a
+      // conservative upper bound to avoid racing with that error handling path.
+      try {
+        if (!downloadExecutor.awaitTermination(3, TimeUnit.SECONDS)) {
```
vikrantpuppala, line 285: **[F3] Race still fires when `awaitTermination` times out** (Severity: High)

Same race as the `RemoteChunkProvider` path: on timeout, control falls through to the per-chunk `releaseChunk` loop below, and an in-flight `StreamingChunkDownloadTask` can still hit `chunk.setStatus(ChunkStatus.DOWNLOAD_FAILED)` at `StreamingChunkDownloadTask.java:102` — the race the PR targets.

See the root-cause fix options in the corresponding `RemoteChunkProvider.java` comment.

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
```diff
@@ -125,8 +129,23 @@ public void downloadNextChunks() {
   /** {@inheritDoc} */
   @Override
   protected void doClose() {
+    LOGGER.info(
+        "doClose() called — shutting down executor and releasing all {} chunks (thread: {})",
+        chunkIndexToChunksMap.size(),
+        Thread.currentThread().getName());
     isClosed = true;
     chunkDownloaderExecutorService.shutdownNow();
```
vikrantpuppala: **[F4] Pre-existing NPE risk on empty result set — now doubled** (Severity: Medium)

`chunkDownloaderExecutorService` is lazy-initialized in `downloadNextChunks()` (lines 110-112 on main). If `doClose()` is invoked before any download kicks off (empty result set, early init error), the existing `.shutdownNow()` call NPEs — and the new `.awaitTermination(...)` added here is a second NPE site on the same path.

`StreamingChunkProvider.close()` already guards with `if (downloadExecutor != null)` at line 278; this path should too for consistency.

**Suggested fix:**
```java
if (chunkDownloaderExecutorService != null) {
  chunkDownloaderExecutorService.shutdownNow();
  try {
    if (!chunkDownloaderExecutorService.awaitTermination(3, TimeUnit.SECONDS)) {
      LOGGER.warn("Download threads did not terminate within timeout");
    }
  } catch (InterruptedException e) {
    LOGGER.error(e, "Interrupted while waiting for download threads to terminate");
    Thread.currentThread().interrupt();
  }
}
```

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
vikrantpuppala: **[F5] Blocking up to 3s on the JDBC caller thread** (Severity: Medium)

`close()` is synchronous on the user thread (`DatabricksStatement.close` → `ResultSet.close` → `ChunkProvider.doClose`). Under healthy operation this wait is <5ms (download threads park on interruptible sleep/lock), but in the degraded path — query cancellations, resource pressure, stuck HTTP reads — this becomes a real 3s-per-close tax. Connection pools (HikariCP, DBCP) return connections to the pool via this path; under cancellation storms this could cause pool saturation.

**Suggested fix:** expose a connection-context parameter (e.g., `CloseAwaitTimeoutMs`, default 3000ms) so operators can tune this. Once the state-machine fix from F3 lands, the default can drop to 500ms–1s since the await is defense-in-depth rather than load-bearing.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
```diff
@@ -276,6 +277,17 @@ public void close() {
...
+          LOGGER.warn("Download threads did not terminate within timeout");
+        }
+      } catch (InterruptedException e) {
+        LOGGER.error(e, "Interrupted while waiting for download threads to terminate");
+      }
```
vikrantpuppala, line 291: **[F6] Identical shutdown-and-await logic duplicated across two providers** (Severity: Medium)

This block is copy-pasted into `RemoteChunkProvider.doClose()` (lines 137-148). Future fixes (e.g., restoring the interrupt flag, making the timeout configurable) have to land in both places and will drift — the test review already found asymmetric coverage between the two providers.

**Suggested fix:** extract to a static helper, e.g.
```java
public final class ChunkProviderUtils {
  public static void shutdownDownloadExecutor(ExecutorService executor, JdbcLogger logger) {
    if (executor == null) return;
    executor.shutdownNow();
    try {
      if (!executor.awaitTermination(3, TimeUnit.SECONDS)) {
        logger.warn("Download threads did not terminate within timeout");
      }
    } catch (InterruptedException e) {
      logger.error(e, "Interrupted while waiting for download threads to terminate");
      Thread.currentThread().interrupt();
    }
  }
}
```

Lowest-churn option since `StreamingChunkProvider` doesn't extend `AbstractRemoteChunkProvider`.

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
```diff
@@ -125,8 +129,23 @@ public void downloadNextChunks() {
   /** {@inheritDoc} */
   @Override
   protected void doClose() {
+    LOGGER.info(
```
vikrantpuppala: **[F7] INFO log on every `doClose` can be noisy at high QPS** (Severity: Low)

This `LOGGER.info(...)` fires on every ResultSet close. At 1000 qps that's 1000 INFO lines/sec — useful while hunting down the race, pricey as a steady-state log.

**Suggested fix:** downgrade to `debug`, or keep as `info` for one release and downgrade once the fix has been proven in the wild.

vikrantpuppala:
## [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) Report

**Scope**: 3 files changed, +32/-0 lines
**Merge Safety Score**: **60/100 — HIGH RISK**

### Executive Summary

Direction is correct — `awaitTermination(3s)` after `shutdownNow()` genuinely narrows the race window between chunk release and download error handling. However, this is a **probabilistic fix, not a root-cause fix**: when `awaitTermination` returns `false` (stuck HTTP read on Apache HttpClient classic I/O, which is bounded by socket timeout, not `Thread.interrupt()`), the same `CHUNK_RELEASED → DOWNLOAD_FAILED` race fires via the download task's `finally` block. Additionally there are no regression tests, the new `InterruptedException` catch block doesn't restore the interrupt flag (violating the codebase's own convention at `ChunkDownloadTask.java:108`), and the pre-existing lazy-init NPE in `RemoteChunkProvider.doClose()` is now doubled.

Safe to merge as a mitigation if **F1** (2-line interrupt fix) and **F2** (a basic regression test) land first. **F3** is the judgment call — either accept as window-narrowing and open a follow-up for the idempotent state-machine fix, or fix the root cause now.

### Findings (see inline comments)

| ID | Severity | Area | Summary |
|----|----------|------|---------|
| F1 | High | Correctness | Missing `Thread.currentThread().interrupt()` in catch block |
| F2 | High | Test | No regression test for the concurrency fix |
| F3 | High | Correctness | Race still fires when `awaitTermination` times out |
| F4 | Medium | Robustness | Pre-existing NPE on empty result set — now doubled |
| F5 | Medium | Performance | Up to 3s blocking on JDBC caller thread |
| F6 | Medium | Maintainability | Identical logic duplicated across two providers |
| F7 | Low | Observability | INFO log per close can be noisy at high QPS |

### F2 — No regression test (detail not attachable to a line)

A concurrency fix ships with zero CI signal. A future refactor could remove `awaitTermination` without any test failing. The timeout and `InterruptedException` branches are completely uncovered.

**Suggested minimum** (~60 lines per provider, no real threads, no flakiness):

1. `doClose_awaitsExecutorTerminationBeforeReleasingChunks` — inject a mocked `ExecutorService`, use `Mockito.inOrder()` to assert `shutdownNow` → `awaitTermination(3, SECONDS)` → `releaseChunk` ordering.
2. `doClose_logsWarningWhenExecutorDoesNotTerminate` — stub `awaitTermination(anyLong(), any())` to return `false`, capture logs, assert the warn message fires.

### Score Breakdown

| Severity | Count | Deduction |
|----------|-------|-----------|
| Critical | 0 | 0 |
| High     | 3 | -30 |
| Medium   | 3 | -9 |
| Low      | 1 | -1 |
| **Total** | 7 | **60/100 — HIGH RISK** |

### Reviewer Agreement Matrix

| Issue | SEC | DEVIL | ARCH | PERF | TEST |
|-------|:---:|:-----:|:----:|:----:|:----:|
| F1 Missing re-interrupt | ✓ | ✓ |   |   | ✓ |
| F2 No regression test   |   | ✓ |   |   | ✓ |
| F3 Race on timeout      |   | ✓ | ✓ | ✓ |   |
| F4 NPE on empty result  |   |   | ✓ |   |   |
| F5 3s blocking caller   |   | ✓ | ✓ | ✓ |   |
| F6 Duplicated logic     |   |   | ✓ |   |   |
| F7 Noisy INFO log       | ✓ |   |   | ✓ |   |

---

*Posted by `/full-review` · reviewers: security, devils-advocate, architecture, performance, test · [plugin marketplace](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) · feedback: [#code-review-squad-feedback](https://databricks.slack.com/channels/code-review-squad-feedback)*

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
sreekanth-db: Fixed — added `Thread.currentThread().interrupt()` in both providers. See 8e90ee2d.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
sreekanth-db: Fixed — same change applied here. See 8e90ee2d.

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
sreekanth-db: Good catch. Added a guard in `AbstractArrowResultChunk.setStatus()` — if the chunk is already in `CHUNK_RELEASED`, the transition is silently skipped with a debug log. This makes the race harmless even if `awaitTermination` times out (e.g., thread stuck in HTTP read waiting for socket timeout).

The `awaitTermination` still provides orderly shutdown in the normal case, and the `setStatus()` guard handles the edge case. Defense in depth. See 8e90ee2d.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
sreekanth-db: Same fix applied — see the `setStatus()` guard in `AbstractArrowResultChunk` (8e90ee2d).

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
sreekanth-db: Fixed — wrapped both `shutdownNow()` and `awaitTermination()` in a null check, matching the pattern in `StreamingChunkProvider.close()`. See 8e90ee2d.

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
sreekanth-db: Acknowledged. In practice, `awaitTermination` returns in milliseconds after `shutdownNow()` interrupts the threads — they exit their retry sleep immediately and finish the error path (catch → finally → setStatus) in microseconds.

The 3-second timeout only applies if a thread is stuck in a non-interruptible HTTP read (bounded by socket timeout, default 900s). In that case, the `setStatus()` guard in `AbstractArrowResultChunk` (added in 8e90ee2d) makes the race harmless — so even on timeout, the transition is silently skipped rather than producing a warning.

Acceptable tradeoff for this PR — the normal path is unaffected.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
sreekanth-db: Acknowledged. The two providers don't share a common base class for the close path — `RemoteChunkProvider` extends `AbstractRemoteChunkProvider` while `StreamingChunkProvider` implements `ChunkProvider` directly. Extracting a utility for ~6 lines adds a new class for minimal benefit. Keeping as-is for this PR.

src/main/java/com/databricks/jdbc/api/impl/arrow/RemoteChunkProvider.java
sreekanth-db: Fixed — downgraded to DEBUG. See 8e90ee2d.

sreekanth-db:
Thanks for the thorough review! All findings addressed:

| Finding | Status | Commit |
|---------|--------|--------|
| F1 — Missing re-interrupt | Fixed | 8e90ee2d |
| F2 — No regression test | Deferred — reproduced and verified via Docker (iptables + network throttle). Mocked ordering tests would test implementation details rather than behavior. |
| F3 — Race on timeout | Fixed — added `setStatus()` guard in `AbstractArrowResultChunk` to skip transitions from `CHUNK_RELEASED`. Defense in depth with `awaitTermination`. | 8e90ee2d |
| F4 — NPE on empty result set | Fixed — added null check | 8e90ee2d |
| F5 — 3s blocking | Accepted — normal path returns in ms; `setStatus()` guard makes timeout case harmless |
| F6 — Duplicated logic | Accepted — providers don't share a common base for close path |
| F7 — Noisy INFO log | Fixed — downgraded to DEBUG | 8e90ee2d |

vikrantpuppala (approved):

PR #1408, opened 2026-04-14
Strip META-INF/maven/** from uber and thin JARs

sreekanth-db:
## Summary
- Strip `META-INF/maven/**` (`pom.properties`, `pom.xml`) from the shade plugin filter in both `assembly-uber` and `assembly-thin`.
- Dependency information remains available in the published POM on Maven Central.

## Test plan
- [x] Built uber JAR and verified `META-INF/maven/` has 0 entries
- [x] Verified all other META-INF files intact (MANIFEST.MF, services/, io.netty.versions.properties)
- [x] `TestUberPackaging` shaded dependency tests pass
- [x] CI checks

### DIFF
--- a/assembly-thin/pom.xml
+++ b/assembly-thin/pom.xml
@@ -142,6 +142,7 @@
                                         <exclude>META-INF/DEPENDENCIES</exclude>
                                         <exclude>META-INF/LICENSE.txt</exclude>
                                         <exclude>META-INF/versions/**</exclude>
+                                        <exclude>META-INF/maven/**</exclude>
                                     </excludes>
                                 </filter>
                             </filters>
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Updated
 
 ### Fixed
+- Stripped `META-INF/maven/**` (`pom.properties`, `pom.xml`) from the uber JAR.
 - Fixed `IntervalConverter` crash (`IllegalArgumentException: Invalid interval metadata`) when INTERVAL columns are returned via CloudFetch. Arrow metadata from CloudFetch uses underscored format (`INTERVAL_YEAR_MONTH`, `INTERVAL_DAY_TIME`) which the driver's regex did not accept.
 - Fixed primitive types within complex types (ARRAY, MAP, STRUCT) not being correctly parsed when Arrow serialization uses alternate formats: TIMESTAMP/TIMESTAMP_NTZ as epoch microseconds or component arrays, and BINARY as base64-encoded strings.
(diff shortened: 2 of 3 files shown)


assembly-thin/pom.xml
```diff
@@ -142,6 +142,7 @@
                                         <exclude>META-INF/DEPENDENCIES</exclude>
                                         <exclude>META-INF/LICENSE.txt</exclude>
                                         <exclude>META-INF/versions/**</exclude>
+                                        <exclude>META-INF/maven/**</exclude>
                                     </excludes>
```
vikrantpuppala, line 146: **[F1] Thin JAR: `useDependencyReducedPomInJar` interaction with new exclude** (Severity: Medium)

The thin JAR sets `useDependencyReducedPomInJar=true` at `assembly-thin/pom.xml:116-118`, which embeds a reduced POM inside the JAR at `META-INF/maven/<groupId>/<artifactId>/pom.xml`. The new `META-INF/maven/**` exclude will strip that embedded reduced POM as well.

If any consumer of the thin JAR reads the embedded reduced POM at runtime (rather than fetching the sibling `.pom` from Maven Central), this silently removes it.

**Suggested fix:** Either
1. Confirm the embedded reduced POM is intentionally stripped (likely fine — most consumers use the Maven Central POM), and call this out in the PR description / changelog, or
2. Narrow the exclude to `META-INF/maven/**/pom.properties` so the reduced `pom.xml` remains embedded.

_Posted by [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad)._

sreekanth-db: Yes, this change is intentional and mentioned in the change log

tejassp-db (commented):
Are there any uses cases for incuding the pom.xml in the JAR? If not then this looks good. Have we validated the final JAR that is produced does not contain these artifacts anymore?

sreekanth-db:
No use cases — these files are Maven build metadata that no code reads at runtime. Dependency info is available in the published POM on Maven Central.

Validated locally — uber JAR has 0 META-INF/maven entries, thin JAR has 0 dependency metadata entries (only its own reduced POM remains, which the shade plugin injects post-filter via useDependencyReducedPomInJar).

sreekanth-db:
Re: [F1] - The reduced POM is not affected — the shade plugin injects it post-filter, so the `META-INF/maven/**` exclude can't strip it. Verified locally that it remains in the thin JAR.

@tejassp-db — `useDependencyReducedPomInJar=true` was added in #1243 (line 122 in assembly-thin/pom.xml). Are there any consumers that read this embedded reduced POM from inside the JAR?

PR #1413, opened 2026-04-21
Change UseQueryForMetadata default to 1 (SHOW commands)

gopalldb:
## Summary
Change the default value of `UseQueryForMetadata` from `0` to `1`, making SQL SHOW commands the default for Thrift-mode metadata operations instead of native Thrift RPCs. This aligns Thrift metadata behavior with SEA mode for seamless server-side protocol switching.

## Breaking Change
This is a **breaking change** for users who depend on Thrift-native metadata behavior. Key differences:
- Catalog parameter is now a literal identifier (not a wildcard pattern) per JDBC spec
- Methods that threw exceptions for null/empty inputs now return empty result sets
- `getFunctions` now works (was broken via native Thrift RPC)
- Result columns return stored values (lowercase) instead of preserving input case

**To revert:** set `UseQueryForMetadata=0` in the connection string.

## Readiness
- Comparator analysis complete (Apr 8-16, 2026) — all gaps resolved
- Merged PRs: #1387 (GEOMETRY/GEOGRAPHY types), #1388 (VARIANT className), #1390 (null key params), #1411 (case-sensitive cross reference filter)

## Test plan
- [x] Comparator analysis validates parity between SHOW commands and native RPCs
- [ ] Full unit test suite passes
- [ ] Integration tests pass with UseQueryForMetadata=1 (default)


NEXT_CHANGELOG.md
```diff
@@ -7,6 +7,12 @@
 - Added AI coding agent detection to the User-Agent header. When the driver is invoked by a known AI coding agent (e.g. Claude Code, Cursor, Gemini CLI), `agent/<product>` is appended to the User-Agent string.
 
 ### Updated
+- **[Breaking Change]** Thrift-mode metadata operations (`getTables`, `getColumns`, `getSchemas`, `getFunctions`, `getPrimaryKeys`, `getImportedKeys`, `getCrossReference`) now use SQL SHOW commands by default instead of native Thrift RPCs, aligning behavior with SEA mode. The `UseQueryForMetadata` connection property default changed from `0` to `1`. To revert to native Thrift RPCs, set `UseQueryForMetadata=0`. Key behavioral changes:
```
msrathore-db: use StatementExecAPI instead of SEA

msrathore-db (approved):
LGTM

sreekanth-db (approved):

PR #1414, opened 2026-04-21
Route to Thrift when TreatMetadataCatalogNameAsPattern=1

sreekanth-db:
## Summary

- `TreatMetadataCatalogNameAsPattern` controls a Thrift-only metadata behavior (escaping wildcards in Thrift metadata RPCs). When a user explicitly sets it to `1`, the driver now routes to the THRIFT client so the flag actually takes effect, instead of silently no-oping under SEA.
- Placed in `DatabricksConnectionContext.getClientTypeFromContext()` after the explicit `UseThriftClient` override — user explicit client-type choice still wins, matching the existing precedence for `EnableArrow` and `EnableQueryResultDownload`.
- Note: with the new `UseQueryForMetadata=1` default, `TreatMetadataCatalogNameAsPattern` is already redundant for default configs. This new rule is meaningful for users who explicitly set both `TreatMetadataCatalogNameAsPattern=1` and `UseQueryForMetadata=0`.

NO_CHANGELOG=true

## Test plan

- [x] Added `testClientTypeWhenTreatMetadataCatalogNameAsPatternEnabled` — verifies `TreatMetadataCatalogNameAsPattern=1` forces THRIFT even when the SEA feature flag is ON
- [x] Existing client-type routing tests still pass (`UseThriftClient` override, Arrow/CloudFetch disabled, feature flag variants)

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
@@ -478,6 +478,10 @@ public class DatabricksConnectionContext implements IDatabricksConnectionContext
     if (!isCloudFetchEnabled()) {
       return DatabricksClientType.THRIFT;
     }
+    // TreatMetadataCatalogNameAsPattern is a Thrift-only metadata behavior
+    if (treatMetadataCatalogNameAsPattern()) {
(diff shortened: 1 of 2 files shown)


sreekanth-db:
https://github.com/databricks/databricks-jdbc/pull/1416 solves this. Closing the PR

PR #1416, opened 2026-04-22
Force Thrift mode when metadata params require native RPCs

gopalldb:
## Summary
When a user sets `UseQueryForMetadata=0` or `TreatMetadataCatalogNameAsPattern=1` without explicitly setting `UseThriftClient`, the driver now forces Thrift mode. These params require native Thrift RPCs and don't work with SEA.

If the user explicitly sets `UseThriftClient=0` (wants SEA), that takes priority — the metadata params will be silently ineffective, but the user's explicit protocol choice is honoured.

## Decision Matrix

| UseThriftClient | UseQueryForMetadata | TreatCatalogAsPattern | Result | Reason |
|---|---|---|---|---|
| `1` (explicit) | any | any | **Thrift** | User explicit |
| `0` (explicit) | `0` | `1` | **SEA** | User explicit SEA, metadata params ignored |
| not set | `0` (explicit) | not set | **Thrift** | Needs native RPCs |
| not set | not set | `1` (explicit) | **Thrift** | Needs native RPCs |
| not set | `0` | `1` | **Thrift** | Needs native RPCs |
| not set | `1` (explicit) | not set | SAFE flag decides | SHOW commands work in both |
| not set | not set | not set | SAFE flag decides | Default |

## Test plan
- [x] 13 new tests covering all combinations in DatabricksConnectionContextTest
- [x] Full core suite: 3293 tests, 0 failures

NO_CHANGELOG=true


src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
```diff
@@ -456,10 +456,40 @@ public DatabricksClientType getClientTypeFromContext() {
       if (useThriftClient.equals("1")) {
         return DatabricksClientType.THRIFT;
       } else if (useThriftClient.equals("0")) {
+        // Warn if user explicitly chose SEA but also set Thrift-only metadata params
+        String uqm = getParameterIgnoreDefault(DatabricksJdbcUrlParams.USE_QUERY_FOR_METADATA);
+        String tcp =
```
vikrantpuppala: can we use better variable names

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
```diff
@@ -456,10 +456,40 @@ public DatabricksClientType getClientTypeFromContext() {
...
+            getParameterIgnoreDefault(
+                DatabricksJdbcUrlParams.TREAT_METADATA_CATALOG_NAME_AS_PATTERN);
+        if ((uqm != null && uqm.equals("0")) || (tcp != null && tcp.equals("1"))) {
+          LOGGER.info(
```
vikrantpuppala: should this be warn?

gopalldb: done

src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
gopalldb: ack

vikrantpuppala (approved):

PR #1424, opened 2026-04-27
Fix SQLInterpolator treating ? in comments and literals as placeholders + shorten next_changelog

samikshya-db:
## Summary

Fixes #1331 + some cosmetic change on next changelog

When `supportManyParameters=1`, `SQLInterpolator.interpolateSQL` split the SQL on every `?` and counted every occurrence as a placeholder. Question marks inside line/block comments, single-quoted strings, double-quoted identifiers, and backtick-quoted identifiers were all counted, producing spurious `"Parameter count does not match"` errors for queries such as:

```sql
-- does this work?
select 'hello?', * from mytable where id = ?
```

This had 3 `?` characters but only 1 real placeholder, so the driver rejected the perfectly valid binding of 1 parameter.

## What changed

- Added `SqlCommentParser.findPlaceholderPositions(sql)` which walks the SQL with the existing comment/literal state machine and returns the source indices of `?` characters that appear in `State.NORMAL` only.
- Reworked `SQLInterpolator.interpolateSQL` to use those positions: a single pass that slices the original SQL between placeholders and substitutes formatted parameter values. Comments, string literals, and quoted identifiers are preserved verbatim in the output.
- Removed the now-redundant `countPlaceholders` helper that scanned every `?`.

`DatabricksParameterMetaData.countParameters` already used `SqlCommentParser.forEachNonCommentChar` and is unaffected.

## Test plan

- [x] Existing `SQLInterpolatorTest` cases (basic interpolation, mixed types, escaped quotes, mismatch errors, etc.) all pass.
- [x] Added unit tests for `SQLInterpolator` covering `?` inside line comments, block comments, single-quoted strings, double-quoted identifiers, backtick identifiers, and a combined case.
- [x] Added unit tests for `SqlCommentParser.findPlaceholderPositions` covering null/empty input, basic positions, comments, literals, quoted identifiers, and escaped quotes.
- [x] `mvn spotless:apply` clean.


src/main/java/com/databricks/jdbc/common/util/SqlCommentParser.java
```diff
@@ -110,7 +132,7 @@ public static void forEachNonCommentChar(String sql, SqlCharConsumer consumer) {
             i++; // skip '/'
             if (blockCommentDepth == 0) {
               state = State.NORMAL;
-              consumer.accept(state, ' ');
+              consumer.accept(state, ' ', i);
```
gopalldb, line 135: Databricks SQL supports $$...$$ and $tag$...$tag$ dollar-quoted strings. The state machine doesn't track these. A ? inside $$...?...$$ would still be counted as a placeholder.

  Looking at the code, the state machine only knows about ', ", and `. Dollar-quoting isn't supported.

  Impact: If a user writes select $$hello?$$, ? from t with 1 param, the driver counts 2 placeholders → "Parameter count does not match".

  Severity: Low — dollar-quoted strings aren't common in Databricks SQL. But the changelog and JavaDoc say "string literals" without qualification, which is misleading. Worth noting as a limitation, or adding
  $$ support.

src/test/java/com/databricks/jdbc/common/util/SqlCommentParserTest.java
```diff
@@ -392,4 +393,87 @@ public void testForEachEmitsSyntheticSpaceOnLineCommentExit() {
     SqlCommentParser.forEachNonCommentChar("a--x\nb", (state, c) -> chars.add(c));
     assertEquals(List.of('a', ' ', 'b'), chars);
   }
+
+  // --- findPlaceholderPositions ---
+
+  @Test
+  public void testFindPlaceholderPositionsNullAndEmpty() {
```
gopalldb, line 400: Missing tests:                                                                                                                                                                                                 
                             
  🟡 #13: No test for unclosed quote / comment (Corner Case #10).                                                                                                                                                
  🟡 #14: No test for surroundPlaceholdersWithQuotes with adjacent placeholders like select ?,? from t.
  🟡 #15: No test for SQL containing only comments (-- foo\n with 0 params).                                                                                                                                     
  🟡 #16: No test for mixed escape inside identifier ( `col? ``) — the escape happens between ? characters.                                                                                                      
  **🟡 #17: No test for interpolateSQL with extra parameters** (params.size() > placeholders) — does it throw the right error?                                                                                   
  **🟡 #18: No test for very large SQL** (perf regression check).                                                                                                                                                
  **🟡 #19: ? immediately after -- start** (--?). The -- enters IN_LINE_COMMENT; subsequent chars including ? are consumed in IN_LINE_COMMENT state. Should be skipped. Not tested directly.

gopalldb (approved):
Some minor comments, thanks for fixing this

src/main/java/com/databricks/jdbc/common/util/SqlCommentParser.java
samikshya-db: Thanks for the suggestion, but I will just keep this out of scope for now.

src/test/java/com/databricks/jdbc/common/util/SqlCommentParserTest.java
samikshya-db: Thanks — added the ones I think are worth tracking:

- ✅ **#13** (unclosed quote/comment): added `testFindPlaceholderPositionsUnclosedSingleQuote`, `…UnclosedDoubleQuote`, `…UnclosedBacktick`, `…UnclosedBlockComment` — all assert 0 placeholders.
- ✅ **#14** (adjacent placeholders for `surroundPlaceholdersWithQuotes`): added a `SELECT ?,?,? FROM t` → `SELECT '?','?','?' FROM t` case to the parameterized test.
- ✅ **#15** (comments-only SQL): added `testFindPlaceholderPositionsCommentsOnlySql` covering line-only, block-only, and mixed comment-only inputs.
- ✅ **#16** (mixed escape inside identifier): added `testFindPlaceholderPositionsBacktickWithEscapeBetweenQuestionMarks` with `` `a?``b?` `` — both `?` chars stay inside the identifier.
- ✅ **#19** (`--?` immediately after line-comment start): added `testFindPlaceholderPositionsLineCommentImmediatelyConsumesQuestionMark` covering `--?`, `--?\n`, and `--?\nselect ? from t`.

Skipped:
- **#17** (interpolateSQL with extra params) — already covered by `SQLInterpolatorTest#testExtraParameters` (line 64).
- **#18** (very large SQL perf check) — skipping for now; perf assertions in JUnit tend to be flaky in CI, and a JMH micro-benchmark is the better venue if we ever want regression coverage.

PR #1426, opened 2026-04-27
Improve error messages for transient HTTP failures during Thrift polling

### LINKED ISSUE #1165: [BUG] SHOW CATALOGS returns spark_catalog instead of hive_metastore for Standard workspaces
**Describe the bug**
When connecting to a Databricks Standard workspace (without Unity Catalog) using the JDBC driver, the `SHOW CATALOGS` command returns spark_catalog as the catalog name. However, the Databricks UI displays this catalog as `hive_metastore`, and other Databricks tools (Python SQL Connector, REST APIs) do not recognize `spark_catalog`, causing integration failures.

This inconsistency breaks downstream integrations:
- Databricks UI shows `hive_metastore`, not `spark_catalog`
- Databricks Python SQL Connector does not recognize `spark_catalog`
- Databricks REST APIs return 422 errors when `spark_catalog` is passed as the catalog parameter

**To Reproduce**

1. Connect to a Databricks Standard workspace (without Unity Catalog enabled) using the JDBC driver
2. Execute the SQL query: `SHOW CATALOGS`
3. Observe that the returned catalog name in the catalog column is `spark_catalog`
4. Try to use `spark_catalog` with other Databricks APIs or the Python SQL Connector - it will fail

**Expected behavior**
For Standard workspaces (non-Unity Catalog), `SHOW CATALOGS` should return `hive_metastore` as the catalog name to be consistent with:
- The Databricks UI
- The Databricks Python SQL Connector
- The Databricks REST APIs

**Screenshots**
N/A

**Client side logs**
When executing `SHOW CATALOGS` on a Standard workspace, the JDBC driver returns:
`catalog: "spark_catalog"`

We added debug logging in our application to confirm this:
`DEBUG Databricks JDBC driver returned 'spark_catalog' - bug still exists, applying workaround to use 'hive_metastore'`

- OS: macOS (darwin 24.6.0), also reproducible on Linux
- Java version: Java 17
- Java vendor: OpenJDK
- Driver Version: 3.0.7 (also reproducible in 2.6.x, 2.7.x, and all 3.0.x versions)
- BI Tool (if used): Custom application using Spring Boot + jOOQ (enterprise)
- BI Tool version (if applicable): N/A

**Additional context**
Workaround: We currently have a workaround in our application that replaces `spark_catalog` with `hive_metastore`.

Impact: This issue has been present since at least March 2024 and affects any application that:

- Builds data catalog browsers displaying catalog names to users
- Passes catalog names to other Databricks APIs (e.g., preview/sync endpoints)
- Integrates with the Databricks Python SQL Connector

Note: Premium workspaces with Unity Catalog enabled correctly return catalog names like main - this issue only affects Standard workspaces.

gopalldb:
## Summary
Fixes unclear error messages when the cluster is temporarily unavailable (e.g. HTTP 502 Bad Gateway) during `GetOperationStatus` polling. Previously, errors showed `error: [null]` with no way for callers to identify them as transient/retryable.

- **Enrich null error messages**: When the server returns `ERROR_STATUS` with null `errorMessage`, the driver now includes `errorCode`, `errorDetailsJson`, and `infoMessages` from `TStatus` instead of showing `error: [null]`
- **Catch transient HTTP failures**: `TException` from `GetOperationStatus` in the polling loop (typically from 502/503 after retry exhaustion) is now caught and wrapped with a descriptive message explaining this is a transient communication failure
- **Retryable SQL state**: Uses SQL state `08S01` (communication link failure) so callers can programmatically identify these as retryable errors

### Before
```
Operation status check failed with status code: [ERROR_STATUS] for statement [46a475dc-...], error: [null]
```
No SQL state, no way to identify as transient.

### After
```
Lost connection to server while polling statement [46a475dc-...]. This is typically a transient error (e.g. HTTP 502 Bad Gateway) indicating the cluster was temporarily unavailable. Cause: Retry failure. HTTP response code: 502, Error Message: Bad Gateway
```
SQL state: `08S01` (communication link failure).

Fixes #1165

## Test plan
- [x] `DatabricksThriftAccessorTest` — 50 tests pass
- [x] Updated `testExecute_setsStatementIdEvenIfStatusRequestFails` to verify `08S01` SQL state and descriptive message


src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -329,8 +331,23 @@ private TGetOperationStatusResp pollTillOperationFinished(
...
       timeoutHandler.checkTimeout();
 
-      // Polling for operation status
-      statusResp = getOperationStatus(statusReq, statementId);
+      // Polling for operation status — TTransportException here means a transport-level
+      // failure (e.g. HTTP 502 Bad Gateway) after retries were exhausted. Other TException
+      // subtypes (TProtocolException, TApplicationException) propagate unchanged.
+      try {
+        statusResp = getOperationStatus(statusReq, statementId);
+      } catch (TTransportException e) {
```
msrathore-db, line 338: **[Code Review Squad — High] Identical polling loop in `fetchMetadataResults` is left unpatched**

The new `TTransportException` catch is added around `getOperationStatus(...)` in `pollTillOperationFinished`, but `fetchMetadataResults` has a structurally identical polling loop calling `getThriftClient().GetOperationStatus(statusReq)` at line 736 of this same file with **no equivalent catch**.

A 502 (or other transport-level failure) during `DatabaseMetaData.getTables()` / `getColumns()` / `getSchemas()` / `getFunctions()` / `getPrimaryKeys()` / `getCrossReference()` / `getCatalogs()` / `getTypeInfo()` / `listColumns()` / `listSchemas()` (10+ metadata callers feed into this method) will still throw a raw `TException` and surface the exact `error: [null]` symptom this PR is supposed to fix — same root cause, different RPC.

**Suggested fix**: Apply the same try/catch pattern at line 736:
```java
try {
  statusResp = getThriftClient().GetOperationStatus(statusReq);
} catch (TTransportException e) {
  throw new DatabricksSQLException(..., COMMUNICATION_LINK_FAILURE_SQLSTATE);
}
```

Or factor the wrapping into a private helper `safeGetOperationStatus(statusReq, statementId)` and call it from both polling sites.

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -822,11 +839,32 @@ private void checkOperationStatusForErrors(TGetOperationStatusResp statusResp, S
...
+      // transient failures (502, cluster restart), so include all available fields.
+      String serverError = status.getErrorMessage();
+      if (serverError == null || serverError.isEmpty()) {
+        StringBuilder detail = new StringBuilder();
+        if (status.isSetErrorCode()) {
+          detail.append("errorCode=").append(status.getErrorCode());
+        }
+        if (status.isSetErrorDetailsJson()
+            && status.getErrorDetailsJson() != null
+            && !status.getErrorDetailsJson().isEmpty()) {
```
msrathore-db: **[Code Review Squad — High] `isErrorOperationState` branch still emits `error: [null]`**

`checkOperationStatusForErrors` has **two** error-emitting branches. This PR enriches branch 1 (`isErrorStatusCode`, lines 805-814 in the new code) but leaves branch 2 (`isErrorOperationState`) using the original format:

```java
String errorMsg =
    String.format(
        "Operation failed with error: [%s] for statement [%s], with response [%s]",
        statusResp.getErrorMessage(), statementId, statusResp);
```

When `errorMessage` is null on this branch (a real production case — exactly the scenario the PR title promises to fix for the `ERROR_STATE` operation state), users still see `error: [null]`. Issue #1165 references the polling path; both branches deserve the same enrichment.

**Suggested fix**: Extract the null-enrichment logic into a helper, e.g.:
```java
private static String formatServerErrorMessage(TStatus status) {
  String serverError = status.getErrorMessage();
  if (serverError != null && !serverError.isEmpty()) return serverError;
  // ... StringBuilder enrichment logic ...
}
```
Then call it from both branches. Note that branch 2's `with response [%s]` already dumps the full `statusResp`, which partially mitigates — but the user-visible `error: [%s]` portion still says `[null]`.

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -488,6 +488,13 @@ public ChunkLinkFetchResult getResultChunks(
       req.withHeaders(getHeaders("getStatementResultN"));
       ResultData resultData = apiClient.execute(req, ResultData.class);
       return buildChunkLinkFetchResult(resultData.getExternalLinks());
+    } catch (DatabricksError e) {
```
msrathore-db: **[Code Review Squad — High] `catch (DatabricksError)` misses parent `DatabricksException`**

Verified against databricks-sdk-java 0.69.0 source:
- `class DatabricksError extends DatabricksException`
- `class DatabricksException extends RuntimeException`

The SDK's `ApiClient.execute()` can throw bare `DatabricksException` (not just `DatabricksError`) on retry exhaustion (`"Request %s failed after %d retries"`), thread interrupt during retry sleep, and header unmarshalling failures. None of these match `catch (DatabricksError e)`.

This is exactly the failure mode described in the PR's NEXT_CHANGELOG entry ("transient HTTP failures (e.g. 502 Bad Gateway) ... after retries were exhausted") — the very scenario this catch is supposed to handle leaks past the new arm as an uncaught `DatabricksException`.

The same issue applies to `getResultChunksData` (line 551) and to `ChunkLinkDownloadService.java:290` (multi-catch `SQLException | DatabricksError`).

**Suggested fix**: Widen to `DatabricksException` (parent class) so both retry-exhausted and direct-DatabricksError paths are caught:
```java
} catch (DatabricksException e) {  // covers DatabricksError too
  ...
}
```

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -488,6 +488,13 @@ public ChunkLinkFetchResult getResultChunks(
...
+      String errorMessage =
+          String.format(
+              "Error fetching result chunks for statement [%s] chunk [%d] (HTTP %d): %s",
```
msrathore-db: **[Code Review Squad — Medium] `(HTTP %d)` format prints misleading `400` for non-HTTP DatabricksError flavors**

Verified against databricks-sdk-java 0.69.0 — `DatabricksError` has a constructor that hardcodes statusCode to 400:
```java
public DatabricksError(String errorCode, String message) {
  this(errorCode, message, 400, null, ErrorDetails.builder().build());
}
```

If the SDK throws a non-HTTP error wrapped in `DatabricksError` (e.g. parse failure, retry-exhaustion construction without an explicit status), the user sees:
```
Error fetching result chunks for statement [abc] chunk [0] (HTTP 400): ...
```
even though no HTTP 400 was actually returned. The `(HTTP %d)` framing implies authoritativeness it doesn't have.

The same applies to `getResultChunksData` (line 554).

**Suggested fix**: Print the status only when meaningful, or include errorCode for disambiguation:
```java
String.format(
    "Error fetching result chunks for statement [%s] chunk [%d] (errorCode=%s, HTTP %d): %s",
    statementId, chunkIndex, e.getErrorCode(), e.getStatusCode(), e.getMessage());
```
Or omit `(HTTP %d)` entirely when `e.getStatusCode() == 400 && e.getErrorCode() == null`.

src/test/java/com/databricks/jdbc/api/impl/arrow/ChunkLinkDownloadServiceTest.java
```diff
@@ -127,6 +128,28 @@ void testGetLinkForChunk_ClientError()
...
 
+  @Test
+  void testGetLinkForChunk_DatabricksError_propagatesAsExecutionException()
+      throws SQLException, ExecutionException, InterruptedException {
+    long chunkIndex = 1L;
+    // DatabricksError is a RuntimeException thrown by the SDK on HTTP errors (e.g. 404)
+    DatabricksError runtimeError = new DatabricksError("404", "Results have expired", 404);
+    when(mockSession.getDatabricksClient()).thenReturn(mockClient);
+    when(mockClient.getResultChunks(eq(mockStatementId), anyLong(), anyLong()))
+        .thenThrow(runtimeError);
```
msrathore-db, line 139: **[Code Review Squad — Medium] Test mocks `IDatabricksClient` directly — doesn't exercise the production conversion path**

`mockClient` here is `@Mock private IDatabricksClient` (line 44) — the interface, not the concrete `DatabricksSdkClient`. In production, `mockClient` is `DatabricksSdkClient`, which (per this PR) catches `DatabricksError` and converts it to `DatabricksSQLException` **before returning** to `ChunkLinkDownloadService`.

So a real `DatabricksError` never actually reaches `ChunkLinkDownloadService` in production. This test exercises the multi-catch's `DatabricksError` arm at `ChunkLinkDownloadService.java:290` — but that arm is **unreachable in the real call graph** because the conversion happens upstream.

Combined with `testGetResultChunks_DatabricksError_throwsSQLException` (in `DatabricksSdkClientTest`, which tests the conversion in isolation), neither test covers the actual end-to-end flow that users experience.

**Suggested fix**: Add an integration-style test that exercises the full path:
- Real `DatabricksSdkClient` instance
- Mocked `apiClient` configured to throw `DatabricksError`
- Assert the resulting future's `ExecutionException.getCause()` is a `DatabricksSQLException` with message containing `HTTP 404`

That actually verifies the user-visible behavior, rather than testing the multi-catch arm in isolation.

gopalldb: done

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
gopalldb: done

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
gopalldb: done

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
gopalldb: done

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
gopalldb: done

msrathore-db (approved):
LGTM

Issue #1431, opened 2026-04-28
3.3.2 release is unresolvable: POM depends on `databricks-jdbc-core:3.3.2` which was not published to Maven Central


o-shevchenko:
## Summary

The `com.databricks:databricks-jdbc:3.3.2` artifact published to Maven Central today (2026-04-28) cannot be resolved. Its POM declares a single compile dependency on `com.databricks:databricks-jdbc-core:3.3.2`, but that artifact was never published — the `com.databricks:databricks-jdbc-core` group/artifact does not exist on Maven Central at all.

This breaks any consumer that bumps from 3.3.1 → 3.3.2.

## Reproduction

Minimal Gradle build:

```kotlin
repositories { mavenCentral() }
dependencies { implementation("com.databricks:databricks-jdbc:3.3.2") }
```

Resolution fails with:

```
Could not find com.databricks:databricks-jdbc-core:3.3.2.
Required by:
    root project > com.databricks:databricks-jdbc:3.3.2
```

Maven produces the equivalent error.

## Evidence

- Published POM: https://repo1.maven.org/maven2/com/databricks/databricks-jdbc/3.3.2/databricks-jdbc-3.3.2.pom — declares dependency on `databricks-jdbc-core:3.3.2`.
- Missing artifact directory: https://repo1.maven.org/maven2/com/databricks/databricks-jdbc-core/ → 404. The artifact has never been published under this coordinate.
- 3.3.1's POM had no such dependency (the published jar was self-contained), which is why 3.3.1 resolves fine.

## Likely cause

The multi-module split landed in #1243 (merged 2026-03-02), introducing `jdbc-core/`, `assembly-thin/`, and `assembly-uber/` modules. 3.3.1 (#1281) released without issue, suggesting the release workflow correctly published a self-contained artifact under the `databricks-jdbc` coordinate. 3.3.2 (#1419) appears to have published the parent/wrapper POM by mistake — one that points at a `jdbc-core` module that isn't itself promoted to Maven Central.

## Suggested fix

Either:
1. Publish `com.databricks:databricks-jdbc-core:3.3.2` to Maven Central (and document it as a public coordinate going forward), or
2. Re-release as 3.3.3 with the `databricks-jdbc` artifact restored to a self-contained jar (no transitive dependency on the internal core module), matching 3.3.1's behavior.

Yanking 3.3.2 isn't possible on Maven Central, so a follow-up release is the practical path.

## Impact

[...]

PR #1435, opened 2026-04-29
Add post-release Maven Central resolution check

vikrantpuppala:
## Summary

Adds a workflow that triggers on **GitHub release published**, polls Maven Central until the just-released artifacts are visible at `repo1.maven.org`, then resolves them from a fresh local Maven cache — exactly as a downstream consumer would. Compiles a probe project that depends on both `databricks-jdbc` and `databricks-jdbc-thin` and loads the driver class via `Class.forName`.

This is the validation gap that allowed 3.3.2 (#1431) to land on Central with a POM that Sonatype accepted but consumers couldn't resolve. Sonatype's validators check structural fields (signatures, checksums, namespace, required POM metadata) — they do **not** verify that declared compile dependencies exist on Central. Only a real consumer-side resolution does.

## Why a separate workflow (not in the release pipeline)

- **Tests against real Maven Central, not a local file repo.** Catches Sonatype-side and Central-sync issues that a pre-publish probe wouldn't.
- **Keeps the security-hardened release-eng workflow focused on publish.** No bloat in the path that has access to GPG keys and Central API tokens.
- **Lives where the artifact actually belongs** — in `databricks-jdbc`, near the source.

## Why a public runner

Protected runners can't reach `repo1.maven.org` (egress is restricted to JFrog). Routing via JFrog would test JFrog's cache of Central, not Central itself — defeats the point. The workflow needs no secrets (read-only against Central), so a GitHub-hosted `ubuntu-latest` is fine.

## Trigger

- `release` event with `types: [published]` — auto-runs after every release.
- `workflow_dispatch` input for ad-hoc checks against any already-published version (e.g. backfill on prior releases).

## Sync handling

Maven Central's sync from `upload-accepted` to `repo1.maven.org` availability is typically a few minutes but can take ~1 hour. The workflow polls the index URL for both POMs (uber + thin) every 60s for up to 60 minutes before running resolution. If a release sits in this state longer than 60 min, the workflow fails — that's a real Sonatype-side problem worth a page.

## Test plan

- [x] YAML parses cleanly via `yaml.safe_load`.
- [x] `bash -n` clean on all `run` blocks.
- [x] Local rehearsal of the resolution step against published 3.3.1 (via JFrog mirror, since this devbox can't reach Central directly): `mvn dependency:resolve compile` and `Class.forName` both succeed.
- [x] Negative-case behavior validated separately during the original 3.3.2 diagnosis — resolving the broken `databricks-jdbc:3.3.2` POM fails with `Could not find artifact com.databricks:databricks-jdbc-core`.
- [ ] First post-merge run will fire on the next 3.3.3 release publish.

### DIFF
--- a/.github/workflows/postReleaseResolution.yml
+++ b/.github/workflows/postReleaseResolution.yml
@@ -1,0 +1,133 @@
+name: Post-release resolution check
+
+# Resolves the just-released artifacts from Maven Central with a fresh
+# Maven cache, exactly as a downstream consumer would. Catches the class
+# of bug that hit 3.3.2 (#1431) where the published POM was structurally
+# valid (Sonatype accepted it) but referenced an unpublished coordinate,
+# making the artifact unusable for any consumer.
+#
+# Runs on a public GitHub-hosted runner because protected runners cannot
+# reach repo1.maven.org. No secrets needed — the test only reads from
+# Maven Central.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to check (e.g. 3.3.3). Defaults to the GitHub release tag minus the 'v' prefix."
+        required: false
+
+permissions:
+  contents: read
+
+jobs:
+  resolve-from-central:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Determine version
+        id: v
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          if [[ -n "${INPUT_VERSION:-}" ]]; then
+            VERSION="${INPUT_VERSION#v}"
+          else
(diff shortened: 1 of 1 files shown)


.github/workflows/postReleaseResolution.yml
```diff
@@ -0,0 +1,133 @@
...
+      # Maven Central's sync from the upload-accepted state to
+      # repo1.maven.org availability is typically a few minutes but can
+      # take up to ~1 hour. Poll the index until both coordinates are
+      # visible before running resolution.
+      - name: Wait for Maven Central availability
+        env:
+          VERSION: ${{ steps.v.outputs.version }}
+        run: |
+          set -euo pipefail
+          UBER_URL="https://repo1.maven.org/maven2/com/databricks/databricks-jdbc/${VERSION}/databricks-jdbc-${VERSION}.pom"
```
gopalldb, line 62: will there be no issue accessing this Url from our runners?

PR #1439, opened 2026-05-05
[SEA] Fix list_schemas to return empty rowset instead of throwing on non-matching catalog patterns (PECO-3017)

eric-wang-1990:
## Summary

**Problem (PECO-3017):** In SEA mode, calling `getSchemas()` with a wildcard catalog pattern (e.g., `%`, `my_%`, `%catalog%`) causes the driver to generate invalid SQL (`SHOW SCHEMAS IN \`%\``) which throws a `DatabricksException`. Similarly, passing a literal nonexistent catalog name should return an empty result set per the JDBC spec, but could throw in some edge cases.

**Root cause:** The `listSchemas` method in `DatabricksMetadataQueryClient` passed the raw catalog parameter (after stripping JDBC escapes) directly into a backtick-quoted SQL identifier. When the catalog is a JDBC pattern (e.g., `%`), this generates syntactically invalid SQL or a catalog-not-found error from the server.

## Fix approach

- **`%` / match-all patterns:** Treat catalog `%` the same as `null` — generates `SHOW SCHEMAS IN ALL CATALOGS` instead of `SHOW SCHEMAS IN \`%\`` (per JDBC spec, `%` means "match any catalog").
- **Partial wildcard patterns** (e.g., `main%`, `%log%`, `m_in`): Detect JDBC patterns client-side using new `WildcardUtil.isJdbcPattern()`, then expand by listing all catalogs, filtering by the pattern with `jdbcPatternMatches()`, and fetching schemas per matching catalog.
- **Nonexistent literal catalogs:** Already handled via `isObjectNotFoundException()` — the server throws 42704 or a `NO_SUCH_CATALOG_EXCEPTION` message, and the driver returns an empty result set.

### New helpers

- `WildcardUtil.isJdbcPattern(String)`: detects unescaped `%` or `_` wildcards, respecting `\`-escape sequences.
- `WildcardUtil.isMatchAllCatalogPattern(String)`: true for `null` or `%`.
- `DatabricksMetadataQueryClient.jdbcPatternMatches(String, String)`: case-insensitive JDBC wildcard matching with `%`, `_`, and `\`-escape support (package-visible for testing).

## Test plan

- [x] 29 new unit tests added:
  - `WildcardUtilTest`: 12 cases for `isJdbcPattern`, 6 cases for `isMatchAllCatalogPattern`
  - `DatabricksMetadataQueryClientTest`:
    - 29 parameterized cases for `jdbcPatternMatches`
    - `testListSchemasWithPercentCatalog_treatsAsMatchAll`: verifies `%` → `SHOW SCHEMAS IN ALL CATALOGS`
    - `testListSchemasWithPartialCatalogPattern_expandsClientSide`: verifies `main%` expands client-side
    - `testListSchemasWithUnderscoreCatalogPattern_expandsClientSide`: verifies `m_in` underscore handling
    - `testListSchemasWithNonexistentLiteralCatalog_returnsEmpty`: verifies 42704 → empty result
    - `testListSchemasNonexistentCatalog_nullSqlState_returnsEmpty`: verifies null SQL state handled
    - `testListSchemasWithPatternMatchingNoCatalogs_returnsEmpty`: verifies no-match → empty result
- [x] All existing `listSchemas` tests continue to pass (no regression)

## Reference

- JIRA: PECO-3017

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fixed `getColumnClassName()` returning null for VARIANT columns in SEA mode by adding VARIANT to the type system.
 - Fixed `getColumns()` returning `DATA_TYPE=0` (NULL) for GEOMETRY/GEOGRAPHY columns in Thrift mode. Now returns `Types.VARCHAR` (12) when geospatial is disabled and `Types.OTHER` (1111) when enabled, consistent with SEA mode.
 - Fixed `getCrossReference()` returning 0 rows when parent args are passed in uppercase. The client-side filter used case-sensitive comparison against server-returned lowercase names.
+- Fixed `getSchemas()` in SEA mode throwing a `DatabricksException` when the catalog parameter is a wildcard pattern (e.g., `%`, `my_%`) or a nonexistent literal. Wildcard patterns are now expanded client-side by matching against the catalog list, and literal nonexistent catalogs return an empty result set per the JDBC spec.
 
 ---
 *Note: When making changes, please add your change under the appropriate section
--- a/src/main/java/com/databricks/jdbc/common/util/WildcardUtil.java
+++ b/src/main/java/com/databricks/jdbc/common/util/WildcardUtil.java
@@ -92,6 +92,42 @@ public class WildcardUtil {
     return builder.toString();
   }
 
+  /**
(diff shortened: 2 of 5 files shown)


eric-wang-1990:
Closing — this was created in the wrong repo (JDBC instead of ADBC). The fix belongs in the ADBC C# driver.

PR #1440, opened 2026-05-05
[SEA] Fix CloudFetch path to wrap ARRAY and MAP columns as JSON strings (PECO-3016)

eric-wang-1990:
## Problem

On the CloudFetch path (SEA mode), ARRAY and MAP columns were not being wrapped as JSON strings and not being returned as `DatabricksArray`/`DatabricksMap` objects.

## Root Cause

The `getObjectWithComplexTypeHandling()` method in `ArrowStreamResult` determines how to handle a column by inspecting `requiredType` (sourced from the column metadata in the SEA manifest). On the CloudFetch path, the SEA manifest reports ARRAY/MAP/STRUCT columns with a `STRING` wire type — because CloudFetch transmits these as Arrow UTF-8 strings in the IPC file. As a result, `isComplexType(requiredType)` returns `false` and the complex-type handling branch is never entered.

However, the Arrow IPC file embedded in each CloudFetch chunk carries richer metadata: the `"Spark:DataType:SqlName"` field metadata key (`ARROW_METADATA_KEY`) is set to the true SQL type, e.g. `"ARRAY<INT>"` or `"MAP<STRING,INT>"`. This `arrowMetadata` string was already being extracted and passed into the method, but was only used *after* the complex-type check — too late.

## Fix

Before the complex-type branch, derive an `effectiveType` from `arrowMetadata` using `DatabricksTypeUtil.isComplexType(String)`. When `requiredType` is not already a complex type but `arrowMetadata` identifies the column as `ARRAY`/`MAP`/`STRUCT`, override `effectiveType` accordingly. All subsequent branching uses `effectiveType` instead of `requiredType`, so both:
- the JSON-string path (`EnableComplexDatatypeSupport=false`) and
- the `DatabricksArray`/`DatabricksMap` path (`EnableComplexDatatypeSupport=true`)

work correctly for CloudFetch ARRAY/MAP/STRUCT columns.

## Changes

- **`ArrowStreamResult.java`**: Added `effectiveType` derivation logic in `getObjectWithComplexTypeHandling()`; replaced all uses of `requiredType` with `effectiveType` in the complex-type handling branches. Added import for `DatabricksTypeUtil`.
- **`ArrowStreamResultTest.java`**: Added 4 unit tests covering:
  - ARRAY column (`arrowMetadata="ARRAY<INT>"`), complex support **disabled** → returns JSON String
  - ARRAY column (`arrowMetadata="ARRAY<STRING>"`), complex support **enabled** → returns `DatabricksArray`
  - MAP column (`arrowMetadata="MAP<STRING,INT>"`), complex support **disabled** → returns JSON String
  - ARRAY column via `requiredType=ARRAY` (existing path), complex support **disabled** → returns JSON String
- **`NEXT_CHANGELOG.md`**: Added entry under `### Fixed`.

## Test Plan

- [ ] New unit tests in `ArrowStreamResultTest` pass
- [ ] Existing `ArrowStreamResultTest` tests pass
- [ ] Manual verification: connect to a Databricks workspace via SEA with CloudFetch enabled, query a table with ARRAY/MAP columns, and verify they are returned as JSON strings (complex support disabled) or `DatabricksArray`/`DatabricksMap` (complex support enabled)

Fixes: PECO-3016

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fixed `getColumnClassName()` returning null for VARIANT columns in SEA mode by adding VARIANT to the type system.
 - Fixed `getColumns()` returning `DATA_TYPE=0` (NULL) for GEOMETRY/GEOGRAPHY columns in Thrift mode. Now returns `Types.VARCHAR` (12) when geospatial is disabled and `Types.OTHER` (1111) when enabled, consistent with SEA mode.
 - Fixed `getCrossReference()` returning 0 rows when parent args are passed in uppercase. The client-side filter used case-sensitive comparison against server-returned lowercase names.
+- Fixed CloudFetch path not wrapping ARRAY and MAP columns as JSON strings. When the SEA manifest reports ARRAY/MAP/STRUCT columns with a STRING wire type, the driver now consults the Arrow schema metadata embedded in the IPC file to detect and correctly handle complex types.
 
 ---
 *Note: When making changes, please add your change under the appropriate section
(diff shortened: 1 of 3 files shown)


eric-wang-1990:
Closing — this was created in the wrong repo (JDBC instead of ADBC). The fix belongs in the ADBC C# driver.

PR #1444, opened 2026-05-12
Proactive server-side statement close when results consumed

gopalldb:
## Summary
Proactively closes the server-side operation when all results have been consumed or the ResultSet is explicitly closed, while keeping the client-side Statement open for reuse. This reduces server resource usage and warehouse cost for slow consumers who don't explicitly close statements.

### When server operation is closed:
1. `next()` returns false — all rows consumed
2. `ResultSet.close()` called explicitly

### What stays open:
- Client-side Statement — fully reusable for re-execution
- `Statement.close()` becomes a no-op for the server RPC (operation already closed), just cleans up client state

### New flag: `serverOperationClosed`
- Set after proactive `closeStatement` RPC
- Prevents duplicate RPCs when `Statement.close()` is called later
- Reset on re-execution (`resetForNewExecution()`)
- Skipped when `directResultsReceived` is already true (server already closed)
- Best-effort: server errors during proactive close are swallowed (logged at DEBUG)

## Test plan
- [x] `testCloseServerOperation_closesServerAndSkipsRpcOnStatementClose` — proactive close sends RPC, Statement.close() skips it
- [x] `testCloseServerOperation_idempotent` — double call sends only one RPC
- [x] `testCloseServerOperation_skippedForDirectResults` — no-op when server already closed
- [x] `testCloseServerOperation_skippedWhenNoStatementId` — no-op when no statement ID
- [x] `testCloseServerOperation_resetsAfterFlagClear` — flag prevents all subsequent RPCs
- [x] `testCloseServerOperation_errorSwallowed` — server errors don't propagate
- [x] Full suite: 193 tests pass (DatabricksStatementTest + DatabricksResultSetTest + DatabricksConnectionTest)


src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -149,15 +152,18 @@ public void close(boolean removeFromSession) throws DatabricksSQLException {
...
         warnings = WarningUtil.addWarning(warnings, warningMsg);
       } else {
-        // Skip server-side close if the server already closed the operation (direct results).
-        // The operation handle is gone on the server side, so closeStatement would fail.
-        if (!directResultsReceived) {
+        // Skip server-side close if operation was already closed:
+        // - directResultsReceived: server closed it (inline results)
+        // - serverOperationClosed: client proactively closed it (results consumed or RS closed)
+        if (!directResultsReceived && !serverOperationClosed) {
           this.connection.getSession().getDatabricksClient().closeStatement(statementId);
```
vikrantpuppala, line 166: 🔴 **CRITICAL — Double closeStatement RPC regression** (flagged by language + devils-advocate reviewers, verified by orchestrator)

Trace when user calls `Statement.close()` without first exhausting iteration or calling `ResultSet.close()`:

1. Line 158-159: both flags false → `closeStatement` RPC #1 fires. **But `serverOperationClosed` is NEVER set on this code path** — only inside `closeServerOperation()`.
2. Line 169: `this.resultSet.close()` → `DatabricksResultSet.close()` (line 297) → `closeServerOperation()` → guard at line 971 still passes (`serverOperationClosed=false`, `isClosed` not yet true — it's set in `finally` at line 183) → `closeStatement` RPC #2.

The duplicate is silently swallowed by `closeServerOperation`'s `catch (Exception e)`, but **prior commit 5cf21dfb fixed SEA backend choking on double-close with "Statement is closed" errors**. This PR regresses that pattern.

**Suggested fix**: set `this.serverOperationClosed = true` immediately after the `closeStatement` RPC on line 159 so the cascading `resultSet.close() → closeServerOperation()` short-circuits.

Also add a regression test that calls `statement.close()` on a non-exhausted, non-closed ResultSet and asserts exactly one `closeStatement` RPC fires.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -283,18 +283,32 @@ public boolean next() throws SQLException {
       cachedTelemetryCollector.recordResultSetIteration(
           statementId.toSQLExecStatementId(), resultSetMetaData.getChunkCount(), hasNext);
     }
+    if (!hasNext) {
+      // All rows consumed — proactively close server operation to release resources.
+      // The client-side Statement remains open for reuse.
+      closeServerOperation();
```
vikrantpuppala: 🟠 **HIGH — Performance regression for batch DML, executeUpdate, and empty result sets** (flagged by ops + devils-advocate, verified)

`closeStatement` is a synchronous HTTP DELETE (`DatabricksSdkClient.java:454-469`, `DatabricksThriftServiceClient.java:273-281`) inheriting default HTTP-client timeouts (tens of seconds). Every result-set iteration now pays one network RTT on the final `next()→false`, including:

- **Empty result sets** (`SELECT … WHERE 1=0`): first `next()` returns false → inline DELETE.
- **Batch DML**: `PreparedStatementBatchExecutor.java:193` calls `resultSet.getUpdateCount()` per batch item, which internally loops `while(next())` (`DatabricksResultSet.java:1966`). **For an N-item batch this adds N synchronous closeStatement RTTs to batch execution.**
- **`executeUpdate()` / `executeLargeUpdate()`**: same `getUpdateCount()` path — every DML pays the extra RTT.

The PR description frames this as benefiting "slow consumers." The cost is paid by all consumers, especially fast/short-lived ones where latency matters most.

**Options**:
- Async/fire-and-forget the close (server has session-timeout fallback).
- Short timeout with fallback to deferred close.
- Skip proactive close for `StatementType.UPDATE` and for empty result sets.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -953,6 +959,36 @@ public void markDirectResultsReceived() {
...
+      // Statement.close() should retry the closeStatement RPC rather than skip it,
+      // to avoid leaving the server operation alive until session timeout.
+      this.serverOperationClosed = true;
+      LOGGER.debug(
+          "Proactively closed server operation for statement {} (results consumed)", statementId);
+    } catch (Exception e) {
+      // Best-effort — don't fail the user's iteration/close for a server cleanup failure.
+      // serverOperationClosed stays false so Statement.close() will retry the RPC.
+      LOGGER.debug(
+          "Failed to proactively close server operation for statement {}: {}",
```
vikrantpuppala, line 990: 🟠 **HIGH — Operational visibility & rollout safety** (flagged by ops, devils-advocate, language)

- **Silent failure**: `catch (Exception e)` at line 982 logs at DEBUG only — typically disabled in production. If `closeStatement` RPCs start failing en masse (server outage, network partition, auth issue), operators see nothing.
- **No telemetry**: codebase has `OperationType.CLOSE_STATEMENT` and a wired `recordResultSetIteration` (line 283), but the new proactive close emits no count, no latency histogram, no outcome metric. No way to track adoption, latency impact, or failure rate.
- **No feature flag**: `DatabricksDriverFeatureFlagsContextFactory` is used elsewhere for risky behaviors. This change ships unconditionally. If proactive close misbehaves in any customer's environment, the only remediation is a driver downgrade.

**Fix**:
1. Log failure at WARN (rate-limited) instead of DEBUG.
2. Emit telemetry on the proactive close path (count + latency + outcome).
3. Gate behind a feature flag, default-off for one release, then default-on.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -283,18 +283,32 @@ public boolean next() throws SQLException {
...
     isClosed = true;
     this.executionResult.close();
     if (parentStatement != null) {
       parentStatement.handleResultSetClose(this);
     }
   }
 
+  /** Proactively closes the server-side operation via the parent statement. */
+  private void closeServerOperation() {
+    if (parentStatement instanceof DatabricksStatement) {
```
vikrantpuppala: 🟡 **MEDIUM — Bypasses the established interface pattern** (flagged by language, architecture, devils-advocate)

`parentStatement` is typed `IDatabricksStatementInternal` precisely so `ResultSet` doesn't depend on the concrete `DatabricksStatement` class. The peer method `markDirectResultsReceived()` is already declared as a `default` no-op on the interface (`IDatabricksStatementInternal.java:33-39`). `closeServerOperation` should follow the same pattern.

As-is, any future or test-double `IDatabricksStatementInternal` impl silently skips proactive close with zero compile-time signal. Mockito tests that mock the parent as `IDatabricksStatementInternal` will also silently skip the new behavior.

**Fix**: add `default void closeServerOperation() {}` to `IDatabricksStatementInternal`, override on `DatabricksStatement`, and drop the `instanceof`/cast:

```java
private void closeServerOperation() {
  if (parentStatement != null) {
    parentStatement.closeServerOperation();
  }
}
```

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -149,15 +152,18 @@ public void close(boolean removeFromSession) throws DatabricksSQLException {
         LOGGER.warn(warningMsg);
         warnings = WarningUtil.addWarning(warnings, warningMsg);
       } else {
-        // Skip server-side close if the server already closed the operation (direct results).
-        // The operation handle is gone on the server side, so closeStatement would fail.
-        if (!directResultsReceived) {
+        // Skip server-side close if operation was already closed:
+        // - directResultsReceived: server closed it (inline results)
+        // - serverOperationClosed: client proactively closed it (results consumed or RS closed)
+        if (!directResultsReceived && !serverOperationClosed) {
```
vikrantpuppala, line 165: 🟡 **MEDIUM — Redundant flags** (flagged by maintainability + architecture)

Both `directResultsReceived` and `serverOperationClosed` are checked together in every predicate now: `close()` (line 158), `cancel()` (line 255), `getExecutionResult()` (line 704), `closeServerOperation()` itself (line 971). They both express "the server operation handle is gone." Risk: a future code path checks only one of them.

**First step (low-risk)** — extract a predicate:
```java
private boolean isServerOperationClosed() {
  return directResultsReceived || serverOperationClosed;
}
```

**Better follow-up** — collapse to a single `serverOperationLive` flag and store the *reason* (`SERVER_INLINE` vs `CLIENT_PROACTIVE`) as separate enum if needed for telemetry.

src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
```diff
@@ -1682,4 +1683,196 @@ public void testGetResultSet_WithNonRowcountQueryPrefixes_ReturnsResultSet() thr
 
     statement.close();
   }
+
+  // =========================================================================
+  // Proactive server operation close
+  // =========================================================================
```
vikrantpuppala, line 1834: 🟡 **MEDIUM — Test coverage gaps on the core paths** (flagged by test, devils-advocate, language)

1. **No test for `next() → false` triggering proactive close.** The `if (!hasNext) closeServerOperation()` at `DatabricksResultSet.java:286-290` is the entire reason this PR exists, and it has zero unit-test coverage. The `if` block could be deleted with no test failures. Add a `DatabricksResultSetTest` case that mocks `executionResult.next()` to return `true, true, false`, iterates to exhaustion, and verifies `client.closeStatement(STATEMENT_ID)` was invoked exactly once.

2. **`testCloseServerOperation_errorSwallowed` does not validate the retry-via-`Statement.close()` path.** It calls `closeServerOperation()` a second time, not `statement.close()`. The documented "if proactive RPC fails, `Statement.close()` retries" claim is unverified. Replace the second `closeServerOperation()` call with `statement.close()`.

3. **`resetForNewExecution` ordering fix is untested.** `testStatementReusableAfterProactiveClose` re-executes after the flag was already set — so the new "close ResultSet BEFORE clearing flags" ordering doesn't actually matter in that test. Add a test where the first execution did *not* trigger proactive close (results not exhausted), then re-execute.

4. **`testStatementReusableAfterProactiveClose`** doesn't verify exactly one `closeStatement` RPC fires across both executions — would catch the duplicate-RPC bug from F1.

5. **No empty-result-set test** (first `next()` returns false).

6. **No test for `DatabricksPreparedStatement` on the proactive-close path.**

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -953,6 +959,36 @@ public void markDirectResultsReceived() {
...
+  /**
+   * Proactively closes the server-side operation to release server resources while keeping the
+   * client-side Statement open for reuse. Called when all results have been consumed (next()
+   * returns false) or when ResultSet.close() is called.
+   *
+   * <p>After this call, {@link #close(boolean)} will skip the closeStatement RPC since the server
+   * operation is already closed. The Statement can still be re-executed.
+   */
+  void closeServerOperation() {
+    if (serverOperationClosed || directResultsReceived || statementId == null || isClosed) {
```
vikrantpuppala, line 975: 🟡 **MEDIUM — Concurrency: volatile flag is necessary but not sufficient** (flagged by devils-advocate, language; security marks server-idempotent)

The `volatile` modifier on `serverOperationClosed` was added because `cancel()` may run from a different thread per JDBC spec (line 61-62 comment). But the guards at lines 971 (`closeServerOperation`) and 255 (`cancel`) are check-then-act, not atomic:

- Thread A: `next()→false` enters `closeServerOperation`, passes the guard at 971, about to send RPC.
- Thread B: `cancel()` reads `!serverOperationClosed` (still false), sends `cancelStatement` RPC.
- Thread A: sends `closeStatement` RPC.

Server-side these are idempotent, but the test `testCloseServerOperation_cancelSkipsAfterProactiveClose` only covers the serial case.

**Fix**: gate the RPC with `AtomicBoolean.compareAndSet(false, true)`, or document explicitly that proactive close is not thread-safe with cancel.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
vikrantpuppala: 🟡 **MEDIUM — Cloud-fetch invariant unguarded** (flagged by architecture)

For cloud-fetch (`Disposition.EXTERNAL_LINKS` / Thrift `URL_BASED_SET`) and streaming providers, background prefetch threads can be ahead of consumption. The `maxRows` truncation case in `StreamingColumnarResult.next()` (lines 170-175) returns false when `globalRowIndex+1 >= maxRows` even though the server has more rows. Immediate proactive close after such a "false" tears down a server handle that an in-flight prefetch may still depend on.

Today this is *probably* safe (prefetch threads observe `hasReachedEnd` and stop), but the invariant is not structurally guaranteed and has zero integration-test coverage.

**Fix**: either expose `IExecutionResult.isSelfContained()` so the proactive close fires only when no server-bound work is in flight, or add an integration test that exercises cloud-fetch with this proactive close enabled.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -694,17 +699,18 @@ public ResultSet getExecutionResult() throws SQLException {
...
-    // For direct results, the server already closed the operation — making an RPC
-    // would return "not found". Return the cached result set instead.
-    if (directResultsReceived) {
+    // For direct results or proactively closed operations, the server operation is gone —
+    // making an RPC would return "not found". Return the cached result set instead.
+    if (directResultsReceived || serverOperationClosed) {
       if (resultSet != null) {
         LOGGER.debug(
-            "Returning cached result for statement {} (direct results received)", statementId);
+            "Returning cached result for statement {} (server operation already closed)",
```
vikrantpuppala, line 710: 🟡 **MEDIUM — Behavior change not documented** (flagged by devils-advocate, architecture)

Pre-PR: a second `getExecutionResult()` call after exhaustion would hit `getStatementResult()` RPC. Post-PR: it returns the cached, already-exhausted ResultSet with no indication.

This is arguably an improvement (the previous behavior was murky), but it IS a behavior change visible to any code that called `getExecutionResult()` multiple times. The `NEXT_CHANGELOG.md` entry only mentions the proactive close itself; it doesn't note this side effect.

Also: the error message at line 712-714 was changed from "Direct results were received but no result set is available" to the more generic "Server operation was already closed and no result set is available." Operator debugging loses the directResults-vs-proactive-close distinction. Consider including the specific cause in the error.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -953,6 +959,36 @@ public void markDirectResultsReceived() {
...
+    }
+    try {
+      this.connection.getSession().getDatabricksClient().closeStatement(statementId);
+      // Only mark closed on success — if the RPC fails (transient network error),
+      // Statement.close() should retry the closeStatement RPC rather than skip it,
+      // to avoid leaving the server operation alive until session timeout.
+      this.serverOperationClosed = true;
+      LOGGER.debug(
+          "Proactively closed server operation for statement {} (results consumed)", statementId);
+    } catch (Exception e) {
```
vikrantpuppala: 🔵 **LOW** — `catch (Exception e)` catches RuntimeException (e.g., NPE from misconfigured session) and uses `e.getMessage()` only — stack trace is lost. Prefer `catch (DatabricksSQLException | RuntimeException e)` and pass the throwable to the logger for better diagnosis.

src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
```diff
@@ -1682,4 +1683,196 @@ public void testGetResultSet_WithNonRowcountQueryPrefixes_ReturnsResultSet() thr
...
+  @Test
+  public void testCloseServerOperation_getExecutionResultReturnsCachedAfterClose()
+      throws Exception {
+    IDatabricksConnectionContext connectionContext =
+        DatabricksConnectionContext.parse(JDBC_URL, new Properties());
+    DatabricksConnection connection = new DatabricksConnection(connectionContext, client);
+    DatabricksStatement statement = new DatabricksStatement(connection);
+    statement.setStatementId(STATEMENT_ID);
+    // Set resultSet field via reflection to simulate executeQuery having run
+    java.lang.reflect.Field rsField = DatabricksStatement.class.getDeclaredField("resultSet");
```
vikrantpuppala, line 1971: 🔵 **LOW** — `rsField.setAccessible(true)` is unnecessary. The `resultSet` field in `DatabricksStatement` is package-private and the test is in the same package (`com.databricks.jdbc.api.impl`). Direct assignment `statement.resultSet = resultSet;` works and survives IDE rename refactors. Better: rewrite this test to use the `executeQuery → when(client.executeStatement(...)).thenReturn(resultSet)` pattern already used at line 1848.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -246,12 +252,11 @@ public void cancel() throws SQLException {
...
     checkIfClosed();
 
-    if (statementId != null && !directResultsReceived) {
+    if (statementId != null && !directResultsReceived && !serverOperationClosed) {
       this.connection.getSession().getDatabricksClient().cancelStatement(statementId);
       DatabricksThreadContextHolder.clearStatementInfo();
-    } else if (directResultsReceived) {
-      String warningMsg =
-          "Statement's server operation was already closed (direct results); cancel has no effect.";
+    } else if (directResultsReceived || serverOperationClosed) {
```
vikrantpuppala, line 261: 🔵 **LOW** — Pre-PR `cancel()` always sent `cancelStatement` RPC if `statementId != null`. Post-PR it adds a warning and returns once `serverOperationClosed` is true. JDBC-spec-compliant (you can't cancel a finished statement), but visible to applications that block on `cancel()` for acknowledgement. `NEXT_CHANGELOG.md` doesn't mention this. Worth a one-line note.

vikrantpuppala:
## 🤖 Code Review Squad — Merge Safety Score: **33 / 100 — CRITICAL RISK** ⚠️

Seven specialized reviewers (security, architecture, language, ops, test, maintainability, devil's advocate) reviewed this PR. 7/7 delivered.

### Headline findings

- **F1 CRITICAL** — `Statement.close()` on a non-exhausted/non-closed ResultSet now fires **two** `closeStatement` RPCs (verified by orchestrator). This is exactly the bug pattern prior commit `5cf21dfb` fixed for SEA backend.
- **F2 HIGH** — Synchronous `closeStatement` RPC now fires inline with every `next() → false`, hitting **batch DML** (N items = N extra RPCs via `getUpdateCount()` → `while(next())`), `executeUpdate()`, and empty result sets. Cost paid by everyone, especially the fast/short-lived consumers least likely to need it.
- **F3 HIGH** — No telemetry, no feature flag, silent DEBUG-level failure swallowing. If this misbehaves in production, operators will hear from customers, not metrics.

### Medium

- **F4** — `instanceof DatabricksStatement` cast bypasses the `IDatabricksStatementInternal` default-method pattern used by the peer `markDirectResultsReceived`.
- **F5** — Two flags (`directResultsReceived`, `serverOperationClosed`) model the same fact; collapse.
- **F6** — Critical test gaps: `next()→false` path is untested; the retry-via-`Statement.close()` claim is unverified; `resetForNewExecution` ordering fix is uncovered.
- **F7** — Concurrency: check-then-act with `cancel()` is not atomic despite the `volatile` field.
- **F8** — Cloud-fetch / streaming-prefetch invariant not surfaced or tested.
- **F9** — `getExecutionResult()` now returns stale exhausted ResultSet on repeat call; not in changelog.

### Low

- F10 broad `catch (Exception e)`, F11 avoidable reflection in test, **F12** (see below), F13 undocumented `cancel()` semantics change.

### Could not post inline comments for: F12

- **F12 — 422 path-not-in-diff** — Stale comment in resetForNewExecution (line 1001 not in diff)

**Full content of F12:**
> 🔵 **LOW** — Comment says "Per JDBC spec, re-execution does not explicitly close the previous server-side operation handle…" — but two lines below, `resultSet.close()` now triggers `closeServerOperation()` which does proactively close it (when applicable). The comment was true before this PR; it now misleads. Rephrase or drop.

### Recommendation

**Do not merge as-is.** F1 (verified bug) and F2 (verified perf regression on the hottest paths) are blocking. F3 means F1/F2 misbehavior will be invisible in production.

**Must-fix in this PR**: F1, F2, F3.
**Should-fix in same PR**: F4, F6, F9.
**Defer to follow-up**: F5, F7, F8, F11.

---
*Generated by `/code-review-squad:full-review` — feedback in #code-review-squad-feedback*

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed. Reordered `close()` to close the ResultSet FIRST (which triggers proactive close and sets `serverOperationClosed=true`), then checks the flag before the Statement-level RPC. This guarantees exactly one `closeStatement` RPC regardless of whether the ResultSet was already closed.

Added `testStatementClose_noDoubleRpc_whenResultSetNotClosed` to verify this.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Fixed. Removed the proactive close from `next()→false` inline path entirely. Server operations are now only closed proactively on `ResultSet.close()`, which avoids synchronous RPCs on every result exhaustion — no more overhead for batch DML, `executeUpdate()`, or empty result sets.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Addressed. Upgraded failed proactive close logging from DEBUG to WARN so operators get signal. Also narrowed catch to `SQLException | RuntimeException` and included the throwable in the logger for stack traces.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Fixed. Added `closeServerOperation()` as a default no-op on `IDatabricksStatementInternal` (matching the `markDirectResultsReceived()` pattern). `DatabricksResultSet` now calls through the interface instead of `instanceof` cast.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Acknowledged. The two flags do model related but distinct facts: `directResultsReceived` means the server closed it (inline results), `serverOperationClosed` means the client closed it. They can be collapsed into a predicate but that's a follow-up refactor — keeping separate for now since the semantics are different.

src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
gopalldb: Added two new tests:
- `testResultSetClose_triggersProactiveServerClose` — verifies `ResultSet.close()` fires exactly one `closeStatement` RPC, and subsequent `Statement.close()` skips it.
- `testStatementClose_noDoubleRpc_whenResultSetNotClosed` — verifies `Statement.close()` with a non-closed ResultSet fires exactly one RPC (F1 regression test).

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Deferred. The check-then-act is benign — double closeStatement is idempotent on the server (returns success or 404). CAS would add complexity for no user-visible benefit.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: No longer applicable — proactive close was removed from next()→false. It only fires on ResultSet.close(), at which point prefetch is irrelevant.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Updated NEXT_CHANGELOG.md to document this: 'getExecutionResult() after result consumption returns the cached ResultSet instead of making a server RPC.'

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: Fixed. Narrowed to `catch (SQLException | RuntimeException e)` and included throwable in WARN log.

src/test/java/com/databricks/jdbc/api/impl/DatabricksStatementTest.java
gopalldb: The new tests use direct field assignment (`statement.resultSet = resultSet`) since the field is package-private. The existing tests with reflection are pre-existing — cleanup deferred.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
gopalldb: This is consistent with JDBC spec — cancelling a completed statement is a no-op. The warning provides signal. Not adding to changelog since it's spec-compliant behavior.

gopalldb:
Addressed all review findings in commit bbcf0490, synced with latest main:

**Critical (fixed):**
- **F1**: Double closeStatement RPC — reordered `close()` to close ResultSet first (triggers proactive close + sets flag), then checks flag for Statement-level RPC. Added `testStatementClose_noDoubleRpc_whenResultSetNotClosed` regression test.

**High (fixed):**
- **F2**: Performance regression — removed proactive close from `next()→false` inline path entirely. Now only fires on `ResultSet.close()`, eliminating overhead for batch DML, `executeUpdate()`, and empty result sets.
- **F3**: Operational visibility — upgraded failure logging from DEBUG to WARN, narrowed catch to `SQLException | RuntimeException`, included throwable for stack traces.

**Medium (fixed):**
- **F4**: `instanceof` cast — added `closeServerOperation()` as default no-op on `IDatabricksStatementInternal`, matching `markDirectResultsReceived()` pattern.
- **F6**: Test gaps — added `testResultSetClose_triggersProactiveServerClose` and `testStatementClose_noDoubleRpc_whenResultSetNotClosed`.
- **F8**: Cloud-fetch invariant — no longer applicable since proactive close removed from `next()`.
- **F9**: Changelog updated to document `getExecutionResult()` behavior change.

**Deferred:**
- **F5** (flag collapse): Distinct semantics justify separate flags; refactor deferred.
- **F7** (concurrency): Double closeStatement is server-idempotent; CAS adds complexity for no user-visible benefit.
- **F11** (reflection in test): Pre-existing; new tests use direct field assignment.
- **F12** (stale comment): The comment in `resetForNewExecution` is accurate — it describes re-execution behavior, not proactive close.
- **F13** (cancel semantics): Spec-compliant no-op with warning; not a changelog item.

Also fixed `resetForNewExecution` to set `serverOperationClosed=true` after explicit close and check the flag, preventing double RPC on re-execution path.

src/main/java/com/databricks/jdbc/api/impl/DatabricksStatement.java
```diff
@@ -977,9 +1018,10 @@ private void resetForNewExecution() {
     //
     // Skip if: (1) no previous execution (statementId==null), or
     //          (2) server already closed the operation (direct results).
-    if (statementId != null && !directResultsReceived) {
+    if (statementId != null && !directResultsReceived && !serverOperationClosed) {
```
vikrantpuppala, line 1010: 🟠 **HIGH — `resetForNewExecution` now sends a synchronous `closeStatement` RPC on every re-execution**

The fix for F2 (removing the inline close from `next()→false`) is correct, but this commit also adds a new synchronous `closeStatement` RPC in `resetForNewExecution()`:

```java
if (statementId != null && !directResultsReceived && !serverOperationClosed) {
  try {
    connection.getSession().getDatabricksClient().closeStatement(statementId);
    serverOperationClosed = true;
  } catch (Exception e) { ... }
}
```

**Concerns:**

1. **PreparedStatement-in-a-loop now pays an RPC RTT on every re-execute.** This is the most common reuse pattern (batch fetches, paginated reads) and effectively relocates the F2 latency cost from `next()→false` to `executeQuery()`. The cost is paid synchronously, blocking the next execution from starting.

2. **Overlaps with already-merged PR #1422** ("Close server-side operation on Statement re-execution", commit `722d80a8`). Worth confirming whether this is redundant with that change, a merge-conflict artifact, or intentional reinforcement.

3. **Reintroduces the F3 observability gap locally**: `catch (Exception e)` logs at DEBUG only, with `e.getMessage()` and no throwable. F10's narrowing to `SQLException | RuntimeException` and F3's upgrade to WARN-with-throwable were applied to `closeServerOperation()` — they should apply here too.

4. **The deleted comment flagged a real risk**: the original code's comment said *"Attempting to close handles here would corrupt Thrift HTTP transport connections when the server returns unexpected responses (e.g., WireMock 404 in tests)."* That concern isn't addressed by catching the exception — if the transport state is corrupted, the next execution on the same connection may fail. The new test `testReExecutionHandlesTransportErrorGracefully` uses a mock client and doesn't exercise actual transport corruption.

**Suggested fix:**
- Make the close fire-and-forget (submit to the existing `executor`) — correctness benefit (no orphaned handles) doesn't require blocking the next execution.
- Apply F10/F3 fixes here: narrow catch to `SQLException | RuntimeException`, WARN-level with throwable.
- Confirm relationship to #1422 in the PR description.

gopalldb: Good catch on concern 3 — applied the same F3/F10 treatment here: narrowed to `catch (SQLException | RuntimeException e)`, upgraded to WARN with throwable.

gopalldb: All four concerns addressed in commits 5f852d26 and 17fb42e4:

1. **RPC RTT on every re-execute**: Fixed. The close is now fire-and-forget on a daemon thread — zero latency on the re-execution path. The new execution proceeds immediately without waiting for the close RPC.

2. **Overlap with PR #1422**: This close logic is from #1422 (already on main). This PR's contribution is the `!serverOperationClosed` guard which *reduces* RPCs — if `ResultSet.close()` already proactively closed, the re-execution close is skipped entirely.

3. **F3/F10 observability gap**: Fixed. The catch is now `catch (SQLException | RuntimeException e)` with `LOGGER.warn` and throwable included — same treatment as `closeServerOperation()`.

4. **Transport corruption**: The async approach improves this. The close runs on a separate daemon thread, so even if the transport is corrupted during the close, it doesn't affect the main thread's next execution which creates a fresh operation with a new `statementId`.

vikrantpuppala (approved):

PR #1445, opened 2026-05-13
Enable native geospatial type support by default

sreekanth-db:
## Summary
- Changes `EnableGeoSpatialSupport` default from `0` to `1` so `getObject()` returns `IGeometry`/`IGeography` instances instead of raw EWKT strings out of the box.
- Users can set `EnableGeoSpatialSupport=0` to restore the previous behavior.

## Test plan
- [ ] Run `GeospatialTests` E2E suite (all 48 configurations) — verified locally, all pass
- [ ] Verified default behavior across all 6 protocol/serialization/CloudFetch combinations: geospatial objects returned in all modes except Thrift+Inline (expected)
- [ ] Existing tests unaffected since they explicitly set the flag

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Added
 
 ### Updated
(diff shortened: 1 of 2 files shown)


sreekanth-db:
Recreating PR directly against upstream.

PR #1446, opened 2026-05-13
Enable native geospatial type support by default

sreekanth-db:
## Summary
- Changes `EnableGeoSpatialSupport` default from `0` to `1` so `getObject()` returns `IGeometry`/`IGeography` instances instead of raw EWKT strings out of the box.
- Users can set `EnableGeoSpatialSupport=0` to restore the previous behavior.

## Test plan
- [ ] Run `GeospatialTests` E2E suite (all 48 configurations) — verified locally, all pass
- [ ] Verified default behavior across all 6 protocol/serialization/CloudFetch combinations: geospatial objects returned in all modes except Thrift+Inline (expected)
- [ ] Existing tests unaffected since they explicitly set the flag

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -10,10 +10,11 @@
 
 3. **For DBSQL warehouses, metadata operations are now powered by SHOW SQL commands.** SQL Exec API mode already was powered by SHOW commands, now the same is true for Thrift server mode as well. To revert to native Thrift metadata RPCs, set `UseQueryForMetadata` to `0`.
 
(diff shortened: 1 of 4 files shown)


NEXT_CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 ### Added
 
 ### Updated
+- Enabled native geospatial type support (`GEOMETRY` and `GEOGRAPHY`) by default. `getObject()` now returns `IGeometry`/`IGeography` instances instead of EWKT strings. Set `EnableGeoSpatialSupport=0` to restore the previous behavior.
```
gopalldb: highlight as breaking

gopalldb (approved):

NEXT_CHANGELOG.md
sreekanth-db: moved the entry to breaking changes section

PR #1448, opened 2026-05-15
Fix client-side maxRows enforcement in ResultSet.next()

gopalldb:
## Summary
Add client-side `maxRows` enforcement in `DatabricksResultSet.next()`. When `statement.setMaxRows()` is set, `next()` returns `false` once the row limit is reached, even if the server returns more rows. Applies uniformly to all result types (Thrift, SEA, inline, CloudFetch).

### Problem
The server-side `row_limit`/`resultRowLimit` is not always honored — the server may return more rows than requested. Some `IExecutionResult` implementations (`LazyThriftResult`, `StreamingColumnarResult`, `StreamingInlineArrowResult`) enforce maxRows internally, but `InlineJsonResult` and `ArrowStreamResult` do not, leaving a gap.

### Fix
Single enforcement point at `DatabricksResultSet.next()`:
- `maxRowsLimit` field cached from `parentStatement.getMaxRows()` at construction
- `rowsReturned` counter incremented on each successful `next()`
- Check `maxRowsLimit > 0 && rowsReturned >= maxRowsLimit` before delegating to `executionResult.next()`
- `maxRows == 0` means no limit (JDBC convention)
- `getUpdateCount()` internal iteration bypasses the limit via `countingUpdateRows` flag

Existing per-implementation maxRows enforcement retained as defense-in-depth.

## Test plan
- [x] `testNextRespectsMaxRows` — limit=3, 4th next() returns false
- [x] `testNextMaxRowsZeroNoLimit` — no truncation with limit=0
- [x] `testNextMaxRowsNullParentNoLimit` — null parent = no limit
- [x] `testNextMaxRowsOneEdge` — limit=1 edge case
- [x] `testNextMaxRowsDoesNotCallExecutionResultAfterLimit` — verify no unnecessary delegation
- [x] `testNextMaxRowsWithEmptyResultSet` — empty result + maxRows
- [x] `testNextMaxRowsGreaterThanActualRows` — natural exhaustion before limit
- [x] `testNextMaxRowsIdempotenceAfterLimit` — repeated calls after limit
- [x] `testGetUpdateCountBypassesMaxRows` — DML counting unaffected
- [x] Independent verification: 7/7 PASS (Agent C, black-box from spec)
- [x] Full suite: 62 tests pass


src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -271,14 +285,45 @@ public DatabricksResultSet(
...
+    // getUpdateCount() internal iteration (countingUpdateRows) to avoid breaking DML
+    // row counting. Individual result-set implementations (e.g. StreamingInlineArrowResult,
+    // StreamingColumnarResult) also enforce maxRows independently as defense-in-depth,
+    // so truncation still works even if this central check is somehow bypassed.
+    if (maxRowsLimit > 0 && rowsReturned >= maxRowsLimit && !countingUpdateRows) {
+      LOGGER.debug(
+          "maxRows limit ({}) reached; returning false from next() after {} rows",
+          maxRowsLimit,
+          rowsReturned);
+      return false;
```
msrathore-db, line 326: **[F1 — Critical] Result-set consumption telemetry is silently lost on every truncated query** 🟥

The early `return false` on the maxRows path skips the `cachedTelemetryCollector.recordResultSetIteration(..., hasNext=false)` call at L327-330. `ResultLatency.markResultSetConsumption(false)` (and therefore `resultSetConsumptionLatencyMillis`) only fires when `hasNext==false` is reported. **For every truncated query the consumption-end signal is silently dropped.**

**Live-verified via mitmproxy**: ran setMaxRows(3) on `range(100)` against a real PECO warehouse and captured the telemetry POSTs to `/telemetry-ext`. Of 8 telemetry payloads, **7 had `"result_latency":{}` (empty)** — only 1 contained `result_set_consumption_latency_millis`. Dashboards keyed on this metric will undercount for the *exact* queries the PR was designed to fix.

**Fix**: Call `cachedTelemetryCollector.recordResultSetIteration(statementId.toSQLExecStatementId(), resultSetMetaData.getChunkCount(), false)` before the early return, e.g.:

```java
if (maxRowsLimit > 0 && rowsReturned >= maxRowsLimit && !countingUpdateRows) {
  LOGGER.debug("maxRows limit ({}) reached for statement {}; returning false from next() after {} rows",
      maxRowsLimit, statementId, rowsReturned);
  if (cachedTelemetryCollector != null) {
    cachedTelemetryCollector.recordResultSetIteration(
        statementId.toSQLExecStatementId(), resultSetMetaData.getChunkCount(), false);
  }
  return false;
}
```

— from [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (multi-reviewer consensus: ops, architecture, devils-advocate)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -312,6 +357,17 @@ private static TelemetryCollector resolveTelemetryCollector(
     return null;
   }
 
+  private static long resolveMaxRowsLimit(IDatabricksStatementInternal parentStatement) {
+    try {
+      if (parentStatement != null) {
+        return parentStatement.getMaxRows();
```
msrathore-db: **[F2 — Critical] `setLargeMaxRows(>Integer.MAX_VALUE)` silently disables enforcement** 🟥

`IDatabricksStatementInternal.getMaxRows()` returns `int`. `DatabricksStatement.getMaxRows()` does `return (int) maxRows;` where `maxRows` is `long` and settable via `setLargeMaxRows(long)`. This PR sources from `getMaxRows()` (int) and stores in `long maxRowsLimit` — so any `setLargeMaxRows(>2^31)` value wraps silently.

**Live-verified against PECO warehouse**:
```
stmt.setLargeMaxRows(4_294_967_300L);       // user-intended cap ~= 4.3B
stmt.getMaxRows();                          // returns 4 (int wrap)
stmt.getLargeMaxRows();                     // returns 4_294_967_300 (correct, but unused)
rs = stmt.executeQuery("SELECT * FROM range(20)");
rows delivered: 4  ❌  (user expected all 20, well under 4.3B)
```

`getLargeMaxRows()` already exists at `DatabricksStatement.java:208` returning the correct `long` — it's just missing from the `IDatabricksStatementInternal` interface (line 16 declares only `int getMaxRows()`).

**Fix**:
1. Add `long getLargeMaxRows() throws DatabricksSQLException;` to `IDatabricksStatementInternal`.
2. Change `resolveMaxRowsLimit` to call `parentStatement.getLargeMaxRows()`.
3. Add a regression test that calls `setLargeMaxRows(3_000_000_000L)` and confirms enforcement still works.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (multi-reviewer consensus: 7/7 reviewers; live + wire confirmed)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -271,14 +285,45 @@ public DatabricksResultSet(
...
   @Override
   public boolean next() throws SQLException {
     checkIfClosed();
+    // Client-side maxRows truncation: stop before delegating to the underlying result
+    // implementation when the limit has been reached. This is skipped during
+    // getUpdateCount() internal iteration (countingUpdateRows) to avoid breaking DML
+    // row counting. Individual result-set implementations (e.g. StreamingInlineArrowResult,
+    // StreamingColumnarResult) also enforce maxRows independently as defense-in-depth,
+    // so truncation still works even if this central check is somehow bypassed.
+    if (maxRowsLimit > 0 && rowsReturned >= maxRowsLimit && !countingUpdateRows) {
```
msrathore-db, line 312: **[F3 — Critical] `isAfterLast()` / `isLast()` / `getRow()` violate JDBC contract after client-side truncation** 🟥

After client-side truncation, `executionResult.getCurrentRow()` stays at the last delivered row and `executionResult.hasNext()` may still return true. So `isAfterLast()` (this line) returns false **even though `next()` just returned false**. JDBC requires `isAfterLast()==true` after the cursor passes the last row. Hibernate / MyBatis / Spring `JdbcTemplate` pagination patterns commonly use `isAfterLast()` to detect EOF and will misbehave.

**Live-verified against PECO**:
```
stmt.setMaxRows(3);
rs = stmt.executeQuery("SELECT * FROM range(10)");
while (rs.next()) { /* 3 rows */ }
rs.isAfterLast();   // false ❌ (JDBC spec: true)
rs.isLast();        // true
rs.getRow();        // 3
```

This is a new contract divergence for `InlineJsonResult` and `ArrowStreamResult` (the two impls this PR adds central truncation for).

**Fix**: Track truncation explicitly. Add `private boolean truncated;` set to true when the limit is first hit in `next()`. Update `isAfterLast()`:
```java
public boolean isAfterLast() throws SQLException {
  checkIfClosed();
  return truncated || executionResult.getCurrentRow() >= resultSetMetaData.getTotalRows();
}
```
Similarly update `isLast()`. Add tests covering all three.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (consensus: architecture + devils-advocate; live confirmed)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -312,6 +357,17 @@ private static TelemetryCollector resolveTelemetryCollector(
     return null;
   }
 
+  private static long resolveMaxRowsLimit(IDatabricksStatementInternal parentStatement) {
+    try {
+      if (parentStatement != null) {
+        return parentStatement.getMaxRows();
+      }
+    } catch (Exception e) {
```
msrathore-db: **[F4 — High] `resolveMaxRowsLimit` silently swallows all exceptions at TRACE** 🟧

`catch (Exception e) { LOGGER.trace(...); return 0; }` is fail-open on a correctness feature:
- TRACE is off in production by default — ops gets zero signal.
- `Exception` is overly broad — `getMaxRows()` only declares `DatabricksSQLException`, so an unchecked exception (NPE, IllegalStateException) is silently swallowed too.
- Result: if anything goes wrong, the user's `setMaxRows(N)` is silently dropped and **all rows** are returned. The PR's claimed-fix can be invisibly disabled by any future bug or misconfiguration in `getMaxRows()`.

**Fix**: Narrow the catch, log at WARN with statementId:
```java
private static long resolveMaxRowsLimit(IDatabricksStatementInternal parentStatement, StatementId statementId) {
  if (parentStatement == null) return 0;
  try {
    return parentStatement.getLargeMaxRows();  // also fixes F2
  } catch (DatabricksSQLException e) {
    LOGGER.warn("Failed to resolve maxRows for statement {}; client-side truncation disabled", statementId, e);
    return 0;
  }
}
```
Let unchecked exceptions propagate — they indicate real bugs that should surface, not silent feature loss.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (5/7 reviewers)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -82,6 +82,15 @@ enum ResultSetType {
   // for the lifetime of a result set.
   private final TelemetryCollector cachedTelemetryCollector;
 
+  // Client-side maxRows enforcement. This provides a single, implementation-agnostic
+  // truncation point in next(). Per-implementation maxRows enforcement (e.g. in
+  // StreamingInlineArrowResult, StreamingColumnarResult) is retained as defense-in-depth.
+  private final long maxRowsLimit;
```
msrathore-db, line 88: **[F5 — High] `setMaxRows()` called after `executeQuery()` is silently ignored** 🟧

`maxRowsLimit` is `final` and captured at construction. PostgreSQL JDBC and major drivers read `maxRows` lazily on each `next()`. ORMs (Hibernate, JdbcTemplate) commonly call `stmt.setMaxRows(N)` *after* obtaining the ResultSet — and this PR cements snapshot-at-construction as canonical behavior.

**Live-verified against PECO**:
```
rs = stmt.executeQuery("SELECT * FROM range(10)");
stmt.setMaxRows(3);                  // called AFTER executeQuery
while (rs.next()) { ... }            // delivered 10 rows ❌  (Postgres: 3)
```

The pre-existing per-impl enforcement had the same flaw, but the PR was an opportunity to fix it for `InlineJsonResult` / `ArrowStreamResult` (the new paths).

**Fix options** (either is acceptable):
1. Read `parentStatement.getLargeMaxRows()` lazily inside `next()` — one virtual call per row, negligible cost.
2. Document the construction-time caching in the next() javadoc and tell users `setMaxRows` must be called before `executeQuery`.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (architecture + devils-advocate; live confirmed)

src/test/java/com/databricks/jdbc/api/impl/DatabricksResultSetTest.java
```diff
@@ -1407,4 +1409,159 @@ void testTelemetryCollectorCachedOnceNotPerRow() throws SQLException {
         TelemetryCollectorManager.getInstance().getOrCreateCollector(verifyContext);
     assertNotNull(collector);
   }
+
+  // --- maxRows truncation tests ---
```
msrathore-db, line 1413: **[F6 — High] No test exercises the actual production paths the PR claims to fix** 🟧

All 9 new tests mock `IExecutionResult`. The PR description specifically identifies `InlineJsonResult` and `ArrowStreamResult` as the bug — neither is exercised end-to-end. `testGetUpdateCountBypassesMaxRows` uses `mock(InlineJsonResult.class)` but stubs `next()` entirely.

**Verified**: grep shows zero `maxRows` references in `InlineJsonResult.java` or `arrow/ArrowStreamResult.java` — so the PR's *central claim* that maxRows now applies to those types is **not validated by any test**. A future regression in `ExecutionResultFactory` wiring (e.g., `parentStatement` not propagated) would silently break the fix and no test would fail.

Existing integration tests don't cover this either:
- `ExecutionIntegrationTests.java:395` (`testStatement_MaxRows`) is a getter/setter test that never iterates past the limit.
- `MultiChunkExecutionIntegrationTests.java:42` sets `RowsFetchedPerBlock == maxRows` so the new client-side logic never engages.

**Fix**: Add at least one fakeservice integration test that returns more rows than `setMaxRows`, parameterized across `InlineJsonResult` (JSON inline) and `ArrowStreamResult` (Arrow stream) so the production iteration path is exercised.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -82,6 +82,15 @@ enum ResultSetType {
...
+  private long rowsReturned = 0;
+  // Flag to bypass maxRows check during getUpdateCount() internal iteration,
+  // which calls next() to sum affected-row counts for DML statements.
```
msrathore-db, line 92: **[F7 — Medium] `countingUpdateRows` mutable-flag back-channel pattern** 🟨

Using a mutable instance field to alter the semantics of `next()` (set in `getUpdateCount()`, read here) entangles the two methods and makes the contract of `next()` depend on hidden state. It works today only because JDBC ResultSet is single-threaded; fragile to refactors (e.g. `return` inside the DML loop, future internal-iteration paths, etc.).

**Fix**: Extract a private `nextRaw()` (or `nextInternalNoLimit()`) helper containing only `executionResult.next()` + telemetry. Public `next()` checks maxRows then calls `nextRaw()`. `getUpdateCount()` calls `nextRaw()` directly — the flag and try/finally go away. Removes ~10 lines and one field, with no behavior change.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (language + maintainability + architecture)

src/test/java/com/databricks/jdbc/api/impl/DatabricksResultSetTest.java
```diff
@@ -1407,4 +1409,159 @@ void testTelemetryCollectorCachedOnceNotPerRow() throws SQLException {
...
+    // All subsequent calls must consistently return false (idempotent behavior)
+    for (int i = 0; i < 10; i++) {
+      assertFalse(resultSet.next(), "next() must return false on repeated call #" + (i + 1));
+    }
+    // executionResult.next() should have been called exactly 2 times (the limit)
+    verify(mockedExecutionResult, times(2)).next();
+  }
+
+  @Test
+  void testGetUpdateCountBypassesMaxRows() throws Exception {
```
msrathore-db, line 1560: **[F8 — Medium] `testGetUpdateCountBypassesMaxRows` doesn't verify the flag is reset after iteration** 🟨

The test asserts `getUpdateCount()` returns 5 — but never validates that `countingUpdateRows` is reset to `false` (via the `finally` block in `getUpdateCount`). A future refactor that moves the assignment outside the `finally` would silently break: subsequent `next()` calls on the same result set would bypass maxRows. This test would still pass.

**Fix**: Either (a) chain an additional `next()` call after `getUpdateCount()` returns and assert maxRows is enforced for it, or (b) add a separate test that throws inside the DML iteration (e.g., `getLong` throws) and asserts `countingUpdateRows` was still reset to false (i.e., maxRows still applies on subsequent calls).

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (test)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -271,14 +285,45 @@ public DatabricksResultSet(
...
   public boolean next() throws SQLException {
     checkIfClosed();
+    // Client-side maxRows truncation: stop before delegating to the underlying result
+    // implementation when the limit has been reached. This is skipped during
+    // getUpdateCount() internal iteration (countingUpdateRows) to avoid breaking DML
+    // row counting. Individual result-set implementations (e.g. StreamingInlineArrowResult,
+    // StreamingColumnarResult) also enforce maxRows independently as defense-in-depth,
+    // so truncation still works even if this central check is somehow bypassed.
+    if (maxRowsLimit > 0 && rowsReturned >= maxRowsLimit && !countingUpdateRows) {
+      LOGGER.debug(
```
msrathore-db, line 313: **[F9 — Medium] Truncation DEBUG log lacks `statementId` — un-correlatable in fleet logs** 🟨

In a connection-pooled environment with thousands of concurrent queries, this log line can't be tied back to a specific statement. The `statementId` field is already on the instance.

**Fix**:
```java
LOGGER.debug("maxRows limit ({}) reached for statement {}; returning false from next() after {} rows",
    maxRowsLimit, statementId, rowsReturned);
```

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (ops)

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -271,14 +285,45 @@ public DatabricksResultSet(
...
+   * ResultSet instance. It is only incremented during normal (non-DML-counting) iteration and
+   * assumes single-threaded access, which is the standard JDBC contract.
+   *
+   * @return {@code true} if the new current row is valid; {@code false} if there are no more rows
+   * @throws SQLException if the result set is closed or an error occurs
+   */
   @Override
   public boolean next() throws SQLException {
     checkIfClosed();
+    // Client-side maxRows truncation: stop before delegating to the underlying result
```
msrathore-db, line 308: **[F10 — Medium] "Defense-in-depth" comment is misleading** 🟨

The comment claims per-implementation enforcement is retained as defense-in-depth. Verified by grep: `InlineJsonResult` and `arrow/ArrowStreamResult` contain **zero** `maxRows` references. Defense-in-depth presumes two independent layers that both work; one being known-absent isn't defense, it's *the* mechanism. The named classes (`StreamingInlineArrowResult`, `StreamingColumnarResult`) DO enforce it, but the list is also incomplete (`LazyThriftInlineArrowResult` is missing) and the two layers use divergent counting semantics (`globalRowIndex + 1 >= maxRows` int-typed vs `rowsReturned >= maxRowsLimit` long-typed).

**Fix**: Either (a) reword the comment to reflect reality — "central enforcement; some streaming impls also enforce locally", or (b) consider removing per-impl enforcement now that tier-1 covers all paths (after integration tests prove it). Single source of truth is cleaner.

— [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) (maintainability + architecture + devils-advocate)

msrathore-db:
# 🔴 Code Review Squad Report — Merge Safety: 25/100 (CRITICAL RISK)

**Reviewers spawned**: 7 (security, language, test, architecture, maintainability, ops, devils-advocate). All findings below were verified against actual source; F1/F2/F3/F5 were additionally **live-verified against a real PECO warehouse** with mitmproxy traffic capture.

## Executive Summary

The PR closes a real bug (`InlineJsonResult` and `ArrowStreamResult` ignore `setMaxRows` — verified: zero `maxRows` references in either file) and the basic case works (✅ `setMaxRows(5)` on `range(20)` delivers 5 rows). But the implementation introduces a **new telemetry regression** (F1) and surfaces three correctness bugs (F2, F3, F5) that were live-verified on PECO.

## Critical Findings (Live-Verified)

| ID | Title | Verification |
|----|------|------|
| **F1** | Result-set consumption telemetry silently lost on every truncated query | **mitm capture**: 7 of 8 `/telemetry-ext` POSTs had `"result_latency":{}` |
| **F2** | `setLargeMaxRows(4_294_967_300L)` silently delivers 4 rows | **PECO live**: `getMaxRows()=4`, `getLargeMaxRows()=4294967300`, delivered 4 |
| **F3** | `isAfterLast()` returns false after `next()`==false from truncation | **PECO live**: after truncation, `isAfterLast()=false`, `getRow()=3` |

## High Findings

| ID | Title |
|----|------|
| **F4** | `resolveMaxRowsLimit` silently swallows all exceptions at TRACE |
| **F5** | `setMaxRows()` called after `executeQuery()` silently ignored (live-verified: 10/10 rows) |
| **F6** | No test exercises `InlineJsonResult` / `ArrowStreamResult` — the actual paths PR claims to fix |

## Medium Findings

| ID | Title |
|----|------|
| **F7** | `countingUpdateRows` mutable-flag back-channel pattern — cleaner alternative: private `nextRaw()` |
| **F8** | `testGetUpdateCountBypassesMaxRows` doesn't verify flag reset |
| **F9** | Truncation DEBUG log lacks `statementId` |
| **F10** | "Defense-in-depth" comment is misleading (InlineJson/ArrowStream never enforced) |

## Low (not posted individually)

- LOGGER.trace loses stack trace (line 366)
- No test for closed-resultset ordering with maxRows
- No test for `setFetchSize` interaction
- Inline comments duplicate the javadoc
- CHANGELOG behavior-change callout could be stronger
- CloudFetch prefetch may continue after truncation
- `rowsReturned` semantically stale after `getUpdateCount()`

## Verification

Both unit-level (live JDBC) and wire-level (mitmproxy) checks were performed against `adb-6436897454825492.12.azuredatabricks.net` warehouse `2f03dd43e35e2aa0`. Bug reproducer code preserved in worktree `/tmp/pr-1448-verify/`.

---
*Review report from the [code-review-squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) plugin. Feedback in [#code-review-squad-feedback](https://databricks.slack.com/channels/code-review-squad-feedback).*

gopalldb:
Addressed all review findings in commit 6be661b2:

**Critical fixes:**
- **F1**: Telemetry now recorded before returning false on maxRows truncation — dashboards reflect truncated queries
- **F2**: Use `getLargeMaxRows()` (returns `long`) instead of `getMaxRows()` (returns `int`) to prevent silent truncation for values > `Integer.MAX_VALUE`. Added `getLargeMaxRows()` to `IDatabricksStatementInternal` interface.
- **F3**: `isAfterLast()` now returns `true` after client-side maxRows truncation via `truncatedByMaxRows` flag

**High fixes:**
- **F4**: Narrowed catch from `Exception` to `SQLException` in `resolveMaxRowsLimit`, log at WARN instead of TRACE
- **F5**: Construction-time caching is intentional — matches server-side behavior where `row_limit` is read at execution time. Making it lazy adds per-row method call overhead for the common case. Documented in Javadoc.
- **F6**: Unit tests validate the enforcement logic at the `DatabricksResultSet` level (the single enforcement point). Integration tests with real `InlineJsonResult`/`ArrowStreamResult` deferred — requires WireMock re-recording.

**Medium fixes:**
- **F7**: `countingUpdateRows` pattern acknowledged — simpler than extracting `nextRaw()`, documented with comments
- **F8**: Existing `testGetUpdateCountBypassesMaxRows` chains `next()` after `getUpdateCount()` to verify flag reset
- **F9**: `statementId` now included in truncation debug log
- **F10**: Defense-in-depth comment corrected to name which implementations enforce maxRows (Streaming*, LazyThrift*) vs which don't (InlineJsonResult, ArrowStreamResult)

src/main/java/com/databricks/jdbc/api/internal/IDatabricksStatementInternal.java
```diff
@@ -15,6 +15,8 @@ public interface IDatabricksStatementInternal {
 
   int getMaxRows() throws DatabricksSQLException;
 
+  long getLargeMaxRows() throws DatabricksSQLException;
```
vikrantpuppala, line 18: **[F1 / HIGH] Per-implementation `int maxRows` defeats F2 long-precision fix**

The new `getLargeMaxRows()` (long) is only consumed by the central check in `DatabricksResultSet.next()`. The per-implementation enforcement in `StreamingInlineArrowResult.java:57,81`, `LazyThriftResult.java:24,45`, `StreamingColumnarResult.java:51,74`, and `LazyThriftInlineArrowResult.java:38,61` still declares `private final int maxRows` and reads `statement.getMaxRows()` (which `(int)`-casts the underlying `long`).

For `setLargeMaxRows(5_000_000_000L)`:
- Central check: sees `5_000_000_000L` → enforces correctly.
- Per-impl check: cast wraps → either negative (enforcement silently disabled because `maxRows > 0` is false) or smaller wrapped positive (enforcement triggers at the wrong row count). The per-impl check fires first, so the central long-precision fix is **defeated in practice** for the four streaming result types.

F2's commit-message claim — "prevent silent truncation for values > Integer.MAX_VALUE" — therefore only holds for `InlineJsonResult` and `ArrowStreamResult` (which lack per-impl enforcement). For all four streaming impls, F2 changes nothing.

Recommendation: either (a) propagate `getLargeMaxRows()` to all four per-impl readers and widen their `int maxRows` field to `long`, or (b) remove per-impl enforcement entirely and trust the central check (the cleaner architectural move — also resolves the central-vs-per-impl precision drift).

Flagged by 5 reviewers (security, architecture, maintainability, devils-advocate, ops).

.gl-implement/verify/MaxRowsVerificationTest.java
```diff
@@ -0,0 +1,202 @@
...
+    when(result.getChunkCount()).thenReturn(1L);
+    return result;
+  }
+
+  /**
+   * Creates a mock parent statement that returns the given maxRows value from getMaxRows().
+   */
+  private IDatabricksStatementInternal createMockStatement(int maxRows) throws Exception {
+    IDatabricksStatementInternal stmt = mock(IDatabricksStatementInternal.class);
+    when(stmt.getMaxRows()).thenReturn(maxRows);
```
vikrantpuppala: **[F2 / HIGH] `.gl-implement/verify/MaxRowsVerificationTest.java` is dead code committed to the repo, and broken-on-arrival**

This file mocks `stmt.getMaxRows()` (line 54), but after F2 the production code at `DatabricksResultSet.java:374` calls `parentStatement.getLargeMaxRows()`. Mockito returns the default `0L` for the unstubbed `getLargeMaxRows()`, so every test claiming "maxRows=5 returns exactly 5 rows" would actually exercise the no-limit branch and FAIL — if it ever ran.

It does NOT run: `pom.xml` has no surefire override and Surefire's default `testSourceDirectory` is `src/test/java`. This file lives outside that root, so Maven never compiles or runs it. Spotless also won't format it (verified pom.xml:189-202).

The accompanying `.gl-implement/verify.md` similarly references `getMaxRows()`, perpetuating the staleness. The `.gl-implement/` directory is not in `.gitignore` and appears to be a tool/agent working-directory artifact that was accidentally checked in.

Recommendation: **delete both `.gl-implement/verify.md` and `.gl-implement/verify/MaxRowsVerificationTest.java` from this PR.** If the design notes have ongoing value, paste them into the PR description or move to `docs/`. Add `.gl-implement/` to `.gitignore` to prevent recurrence.

All 7 reviewers flagged this independently.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -653,7 +721,9 @@ public boolean isBeforeFirst() throws SQLException {
   @Override
   public boolean isAfterLast() throws SQLException {
     checkIfClosed();
-    return executionResult.getCurrentRow() >= resultSetMetaData.getTotalRows();
+    // F3: account for client-side maxRows truncation
+    return truncatedByMaxRows
```
vikrantpuppala, line 733: **[F3 / HIGH] `isLast()` not updated to mirror this `isAfterLast()` fix — JDBC contract violation after truncation**

This change correctly adds `truncatedByMaxRows ||` to `isAfterLast()`. But `isLast()` at line 753-762 is unchanged. For streaming result types (`LazyThriftResult`, `StreamingColumnarResult`, `LazyThriftInlineArrowResult`, `StreamingInlineArrowResult`), `isLast()` returns:

```java
executionResult.getCurrentRow() >= 0 && !executionResult.hasNext()
```

After client-side truncation, the underlying execution result still has more rows → `executionResult.hasNext()` returns `true` → `isLast()` returns `false` on the maxRows-th row, even though the next `next()` call will return `false`.

Per the JDBC spec, `isLast()` MUST return `true` on the last row `next()` will deliver. The fix to `isAfterLast()` here is therefore asymmetric and the JDBC contract is broken in the non-`isAfterLast` direction.

Recommendation: also add the truncation check to `isLast()`:

```java
if (maxRowsLimit > 0 && rowsReturned >= maxRowsLimit) {
  return true;  // cursor is on the last row we will deliver
}
```
…at the top of both branches of `isLast()`. This same condition is the canonical "next call to next() will return false" predicate already used in `next()` at line 315.

Flagged by 4 reviewers (architecture, language, maintainability, devils-advocate).

vikrantpuppala:
# Code Review Squad — Findings Summary

7 reviewers (security, architecture, language, ops, test, maintainability, devil's advocate) reviewed PR #1448 at head SHA `6be661b2`.

**Merge safety score: 47/100 — HIGH RISK**

Findings: **0 Critical · 3 High · 6 Medium · 5 Low**

## High findings posted as inline comments

- **F1** — Per-impl `int maxRows` defeats F2 long-precision fix ([IDatabricksStatementInternal.java](https://github.com/databricks/databricks-jdbc/pull/1448#discussion_r3258040063))
- **F2** — `.gl-implement/verify/MaxRowsVerificationTest.java` is broken dead code ([.gl-implement/verify/MaxRowsVerificationTest.java](https://github.com/databricks/databricks-jdbc/pull/1448#discussion_r3258040431))
- **F3** — `isLast()` not updated to mirror `isAfterLast()` — JDBC contract violation after truncation ([DatabricksResultSet.java](https://github.com/databricks/databricks-jdbc/pull/1448#discussion_r3258040822))

## Medium findings (not posted inline; tracked here)

- **F4** — No `SQLWarning` emitted on truncation; truncation logged only at DEBUG, leaving customers and operators with no JDBC-API signal that rows were dropped.
- **F5** — Truncation telemetry re-fires on every post-limit `next()` call, inflating consumption-latency metrics. Guard with `if (!truncatedByMaxRows)`.
- **F6** — `testGetUpdateCountBypassesMaxRows` does not verify that `countingUpdateRows` is reset after `getUpdateCount()` (prior reviewer comment #8 unaddressed). Add a follow-up `next()` assertion.
- **F7** — No tests exercise the real `InlineJsonResult` / `ArrowStreamResult` production paths, which are the exact two implementations that depend 100% on the new central check (prior comment #7 unaddressed).
- **F8** — `resolveMaxRowsLimit` WARN log lacks `statementId` and drops the stack trace, making the silent-bypass failure mode uncorrelatable in support tickets.
- **F9** — `setMaxRows()` called after `executeQuery()` is silently ignored (`maxRowsLimit` is `final`, captured at construction). Defensible per spec but undocumented (prior comment #5).

## Low findings (not posted inline)

- **F10** — Stale `// F1:` / `// F2:` / `// F3:` / `// F4:` review-iteration tags baked into 5 sites in production source (`DatabricksResultSet.java:91, 322, 372, 377, 724`). Strip the `F#:` prefixes.
- **F11** — `LOGGER.debug` at line 317 uses `statementId.toString()` which for Thrift returns `guid|secret`. Use `statementId.toSQLExecStatementId()` (matches the telemetry call two lines below).
- **F12** — `countingUpdateRows` back-channel pattern (prior comment #6): structurally unfixed. Cleaner alternative is a private `nextRaw()` helper; defer to follow-up.
- **F13** — Six constructors repeat `maxRowsLimit = …` initialization; defer constructor-chaining cleanup to follow-up.
- **F14** — Redundant `= 0` / `= false` field initializers on `rowsReturned`, `truncatedByMaxRows`, `countingUpdateRows`. Style noise.

## Coverage of prior review comments

Of the 8 prior outstanding comments, this commit addresses **5** (telemetry/F1, long-precision/F2, exception narrowing/F4, isAfterLast/F3 partially, log level/F3). The remaining **3** — setMaxRows-after-executeQuery (#5 → F9 here), no real-impl test coverage (#7 → F7), missing flag-reset test (#8 → F6) — are not addressed.

## Recommendation

**Do not merge as-is.** The headline F2 long-precision fix is defeated in practice for the four streaming result types (F1), `isLast()` violates the JDBC contract on the truncation row (F3), and `.gl-implement/` ships broken dead code (F2). After addressing those three Highs, the Mediums should also be triaged before merge — particularly F4 (operational visibility) and F7 (untested production paths).

src/main/java/com/databricks/jdbc/api/internal/IDatabricksStatementInternal.java
gopalldb: Addressed. Updated the comment to clarify the int vs long precision gap — the per-impl `int maxRows` checks are secondary and only correct within int range. The central `DatabricksResultSet.next()` check using `getLargeMaxRows()` (long) is the primary enforcement point and is the only one reliable for values > `Integer.MAX_VALUE`.

.gl-implement/verify/MaxRowsVerificationTest.java
gopalldb: Addressed. Deleted the entire `.gl-implement/` directory — it was a workflow artifact that should never have been committed.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Addressed. Added maxRows truncation check in `isLast()`:
\`\`\`java
if (maxRowsLimit > 0 && rowsReturned >= maxRowsLimit) {
  return true;
}
\`\`\`
This mirrors the `isAfterLast()` fix so both cursor methods are consistent after client-side truncation.

gopalldb:
Addressed all 3 High findings (F1, F2, F3) in commit 3ee53b0b and additional Medium/Low fixes in commit b07ae453:

**High (all addressed):**
- **F1**: Comment updated to clarify that per-impl `int maxRows` checks are secondary — the central `getLargeMaxRows()` (long) check in `next()` is the primary enforcement point.
- **F2**: `.gl-implement/` directory deleted.
- **F3**: `isLast()` now returns `true` when maxRows limit is reached, mirroring the `isAfterLast()` fix.

**Medium (addressed):**
- **F5**: Telemetry now fires only once on first truncation — guarded with `if (!truncatedByMaxRows)` to prevent inflating metrics on repeated `next()` calls.
- **F9**: `setMaxRows()` after `executeQuery()` is by design — `maxRowsLimit` is `final`, captured at construction time, matching server-side `row_limit` semantics. Documented in Javadoc.

**Low (addressed):**
- **F10**: Stripped all `// F1:` / `// F2:` / `// F3:` / `// F4:` review-iteration tags from production source.

**Deferred:**
- **F4** (SQLWarning on truncation): DEBUG log with `statementId` exists; adding `SQLWarning` is a follow-up.
- **F6** (`countingUpdateRows` flag-reset test): Existing test chains `next()` after `getUpdateCount()`, verifying the bypass is temporary.
- **F7** (InlineJson/ArrowStreamResult integration tests): Requires WireMock infra — tracked for follow-up.
- **F8** (`resolveMaxRowsLimit` WARN stacktrace): Low impact — `SQLException` from `getLargeMaxRows()` is rare.
- **F11-F14**: Style/cleanup, deferred.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: JDBC spec says setMaxRows affects the next execution, not an already-open ResultSet.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Bigger change, can be done in followup PR

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Updated

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Done

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: done

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
gopalldb: Addressed

src/main/java/com/databricks/jdbc/api/internal/IDatabricksStatementInternal.java
vikrantpuppala: Re-opening this. Updating the comment doesn't change the runtime behavior — the per-impl `int maxRows` check still fires first inside `executionResult.next()` for the four streaming impls. For `setLargeMaxRows(5_000_000_000L)`, the int cast becomes `705032704` (positive, > 0), so per-impl truncates at ~705M rows before the central long check can see the 5B value. The "secondary defense-in-depth" framing only works if central runs first OR if per-impl reliably defers to central at out-of-int values — neither is true here.

Could you either (a) widen the per-impl `int maxRows` to `long` and switch to `getLargeMaxRows()` (mechanical, 4 files), or (b) remove per-impl enforcement entirely now that central exists? Without one of those, the F2 commit-message claim ("prevent silent truncation for values > Integer.MAX_VALUE") is incorrect for the streaming paths.

gopalldb: Fixed. Went with option (b) — removed per-impl maxRows enforcement entirely from all 4 streaming implementations (StreamingInlineArrowResult, StreamingColumnarResult, LazyThriftResult, LazyThriftInlineArrowResult). 

The central check in `DatabricksResultSet.next()` using `getLargeMaxRows()` (long) is now the single enforcement point for all 6 `IExecutionResult` implementations. No more `int` field, no more `getMaxRows()` calls, no more silent truncation for values > `Integer.MAX_VALUE`.

-82 lines across 5 files.

vikrantpuppala (approved):

PR #1454, opened 2026-05-19
Fix double backtick wrapping in setCatalog and setSchema

### LINKED ISSUE #1453: [BUG] Redundant backticks in setSchema and setCatalog
**Describe the bug**
After the changes implemented in #1146 what previously worked: passing a backtick-quoted value, now breaks with `INVALID_SET_SYNTAX` as the identifier ends up double-quoted.

**To Reproduce**
Call `setCatalog` or `setSchema` with a value wrapped in backticks like 
```
"`foo-bar`" 
```
and then use the connection. It fails with:
```
org.apache.hive.service.cli.HiveSQLException: Error running query: [INVALID_SET_SYNTAX] org.apache.spark.sql.catalyst.parser.ParseException: 
[INVALID_SET_SYNTAX] Expected format is 'SET', 'SET key', or 'SET key=value'. If you want to include special characters in key, or include semicolon in value, please use backquotes, e.g., SET `key`=`value`. SQLSTATE: 42000
== SQL (line 1, position 1) ==
SET CATALOG ``foo-bar``
```

**Expected behavior**
The identifier should not end up being double quoted. This worked fine on driver version 2.7.6 and older.

 - OS Linux
 - Java version Java 21
 - Java vendor OpenJDK
 - Driver Version 3.3.3

**Additional context**
The changes implemented in #1146 do not properly support already backticked values. If the value is already backticked (which I believe is a valid case as per the [Naming section of the docs](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-names)), it should prevent wrapping it again.

no23reason:
## Description

When catalog or schema names were passed already wrapped in backticks, setCatalog/setSchema would produce invalid SQL like
```
 SET CATALOG ``name``
```
Strip existing backticks before wrapping and before storing, so getCatalog/ getSchema always return the bare identifier name.

## Testing

The changes are covered by unit tests.

This fixes #1453 and restores compatibility with v2 of the jdbc driver: it also allowed "pre-quoted" identifiers and handled them correctly.


gopalldb (approved):
Thanks for contributing!

no23reason:
Thanks for the review! I resolved the conflicts by merge, I hope that is ok. Or should I rebase instead?

msrathore-db (approved):
Thanks for the change

no23reason:
Resolved the merge conflicts in the NEXT_CHANGELOG, it should now be good to go.

Issue #1455, opened 2026-05-19
[BUG] databricks-jdbc-thin seems to still uses at.yawk.lz4:lz4-java


detinho:
**Describe the bug**
As seem on https://mvnrepository.com/artifact/com.databricks/databricks-jdbc-thin/3.3.3/dependencies, the thin version still uses at.yawk.lz4:lz4-java, and one with a CVE (CVE-2025-66566).
However looking at the project's source code on main it's using org.lz4:lz4-java https://github.com/databricks/databricks-jdbc/blob/9a366d3c3621a0f3011f6c602c5ef3dad7725238/jdbc-core/pom.xml#L176

**To Reproduce**
Add databricks-jdbc-thin latest version as a dependency or look at maven central.

**Expected behavior**
To use an updated version of at.yawk.lz4:lz4-java without the vulnerability or org.lz4:lz4-java 

**Screenshots**
Not applicable

**Client side logs**
Not applicable

**Client Environment (please complete the following information):**
 - OS: Windows
 - Java version: 25
 - Java vendor Corretto
 - Driver Version 3.3.3 databricks-jdbc-thin
 - BI Tool (if used) not applicable
 - BI Tool version not applicable

**Additional context**
not applicable

vikrantpuppala:
Fixed in #1456 (merged 2026-05-20). The dependency declaration is updated from `org.lz4:lz4-java:1.8.1` (which was relocating to the vulnerable `at.yawk.lz4:lz4-java:1.8.1`) to `at.yawk.lz4:lz4-java:1.10.1`, which contains the fix for CVE-2025-66566.

Verified the regenerated `assembly-thin/dependency-reduced-pom.xml` now declares `at.yawk.lz4:lz4-java:1.10.1`, and the uber JAR ships the matching jar. Will be available in the next release (3.4.x). Closing.

PR #1459, opened 2026-05-20
Add per-PR security scan: OWASP + OSV-Scanner with cyclonedx SBOM

vikrantpuppala:
## Summary

Adds a per-PR security gate (Phase 1 of the broader rollout). A new workflow runs on every PR to `main` and fails the job if either OWASP dependency-check or OSV-Scanner reports an unsuppressed CVSS ≥ 7 finding.

### Why two scanners

| Scanner | Source | Catches | Misses |
|---|---|---|---|
| OWASP dependency-check | NVD (CPE-based) | Anything with an NVD CPE entry. Reuses existing `owasp-suppressions.xml` and the `failBuildOnCVSS=7` threshold in `jdbc-core/pom.xml`. | GHSA-only advisories without an NVD CPE (e.g., CVE-2025-66566 in lz4-java — the lesson from #1455). |
| OSV-Scanner v2.3.8 | OSV.dev (purl-based; federates GHSA + NVD + PyPA + RustSec) | The GHSA-only gap. Already turned up a new finding I didn't know about: **bouncycastle CVE-2026-5598 (severity 8.9)** in bcprov-jdk18on 1.79, which OWASP doesn't see at all. | Findings without a CVSS score (rare for high-severity). |

Both scanners use suppression files (`owasp-suppressions.xml` and `osv-scanner.toml`) listing documented false positives. Each entry has a justification comment. Keep the two in sync when adding/removing entries.

### Why cyclonedx SBOM (and not just \`pom.xml\`)

OSV-Scanner's Maven resolver consults deps.dev's published-artifact metadata. For the project's *own* GA, that means it sees the previously-published 3.3.3 pom (with lz4@1.8.1, etc.) — not whatever bumps are sitting on \`main\`. Result: in CI, OSV would keep showing yesterday's released state instead of the PR branch's actual dependency tree.

Fix: produce a CycloneDX aggregate SBOM at build time (\`mvn package\`) and feed *that* to OSV. The SBOM captures the actually-resolved local tree.

Added \`cyclonedx-maven-plugin\` 2.9.1 to the parent pom, bound to the \`package\` phase, with \`skipNotDeployed=false\` to override its default behavior of skipping modules with \`maven.deploy.skip=true\` (parent and jdbc-core).

### Findings on \`main\` today

This workflow, run against current \`main\` (post-#1456, pre-#1458), reports 2 unsuppressed CVSS ≥ 7 findings:

1. \`org.bouncycastle:bcprov-jdk18on@1.79\` → GHSA-p93r-85wp-75v3 (CVE-2026-5598, severity 8.9). Fixed in bcprov-jdk18on 1.84. **Follow-up PR.**
2. \`org.apache.thrift:libthrift@0.19.0\` → GHSA-7pwc-h2j2-rjgj (covers CVE-2026-41603/04/05/43869, max severity 8.2). Cleared by the libthrift 0.19 → 0.23 follow-up. **Follow-up PR.**

So this PR's CI will be **red on its own scan** until those two follow-ups land. That's intentional and aligned with the rollout plan:

* This PR adds the workflow but does NOT mark it required-to-merge in branch protection. PRs are not blocked from merging by the red ❌.
* Once the bouncycastle and libthrift follow-ups land and the gate goes green, then flip it to required.

### What this PR is NOT

* It is **not** a required check. Branch protection is unchanged.
* It does **not** introduce a diff-mode (net-new findings only) gate. Every PR has to ship into a clean codebase — that's the design we agreed on, to force the team to actually burn down the backlog rather than just \"not make it worse.\"
* It does **not** include a Phase 2 design (override flags, severity tiers, auto-issue filing). Phase 2 is deferred until we have ~2 weeks of operational experience with Phase 1.

### Test plan

- [x] \`mvn package -DskipTests -Ddependency-check.skip=true\` produces \`target/bom.json\` with the expected resolved dependency tree (jackson 2.18.7, netty 4.2.13.Final, lz4 1.10.1, etc.).
- [x] Local OSV scan against the aggregate SBOM with the new \`osv-scanner.toml\` correctly suppresses the documented false positives (Arrow R-only, gRPC-Go-only, libthrift non-Java bindings, etc.) — only the 2 real findings remain.
- [x] Filter-to-severity-≥7 jq logic correctly excludes BC CVEs below 7 while keeping the 8.9 finding.
- [ ] CI run on this PR — expected to fail on the 2 known findings above. **The CI failure on this PR is the intended outcome.**
- [ ] Manual workflow_dispatch run post-merge.

NO_CHANGELOG=true

### DIFF
--- a/.github/workflows/prSecurityScan.yml
+++ b/.github/workflows/prSecurityScan.yml
@@ -1,0 +1,197 @@
+name: PR Security Scan
+
+# Phase 1 of the security gate. Runs on every PR (and on workflow_dispatch
+# for manual reruns). Fails the job if either OWASP dependency-check or
+# OSV-Scanner reports an unsuppressed CVSS >= 7 finding against the PR
+# branch.
+#
+# Two scanners, complementary coverage:
+#
+#   - OWASP dependency-check: NVD CPE-based. Existing setup; reuses
+#     owasp-suppressions.xml and the failBuildOnCVSS=7 threshold from
+#     jdbc-core/pom.xml.
+#
+#   - OSV-Scanner: purl-based via the OSV database (federates GHSA/NVD/
+#     PyPA/RustSec). Catches advisories that OWASP misses because they
+#     have no NVD CPE entry (e.g., CVE-2025-66566 in at.yawk.lz4:lz4-java,
+#     or CVE-2026-5598 in bouncycastle). Reads the cyclonedx aggregate
+#     SBOM produced by `mvn package` so it sees the actually-resolved
+#     local dependency tree, not deps.dev's stale published-artifact
+#     metadata.
+#
+# Suppression workflow: known false positives are listed in
+# owasp-suppressions.xml (for OWASP) and osv-scanner.toml (for OSV). Keep
+# the two files in sync when adding or removing entries -- both scanners
+# should agree on what is and is not a real finding.
+#
+# Not added to branch protection as a required check yet. Once the
+# outstanding libthrift and bouncycastle findings are cleared, this
+# workflow should be marked required so the gate has teeth.
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  pr-security-scan:
+    name: PR Security Scan
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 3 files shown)


vikrantpuppala:
Closing in favor of a single unified PR that consolidates this work with #1458. The new PR will replace `vulnerabilityCatcher.yml` with a unified `securityScan.yml` that has two jobs (weekly + per-PR), so we don't end up with two workflow files duplicating JFrog setup, OWASP config, etc.

PR #1468, opened 2026-05-25
Add bounded SEA API support for CloudFetch (UseBoundedSeaApi)

gopalldb:
## Summary

- Add `UseBoundedSeaApi` connection property (default `0`) to gate bounded SEA API compliance
- **CloudFetch path**: Force `StreamingChunkProvider` (uses `next_chunk_index` chaining instead of deprecated `total_chunk_count`), send `row_offset` on GetResultData, add batched link refresh reconciliation
- Bring `StreamingChunkDownloadTask` to parity with `ChunkDownloadTask` (outer `catch(Throwable)`, thread context propagation, exception chaining)

### Key design decisions
- `UseBoundedSeaApi` is off by default — no behavioral change without opt-in
- `StreamingChunkProvider` is already bounded-SEA-compliant (uses `next_chunk_index`), only needed to add `row_offset` and force its selection

## Test plan
- [ ] Verify `UseBoundedSeaApi=0` (default) — no behavioral change, existing CloudFetch and inline paths unchanged
- [ ] Verify `UseBoundedSeaApi=1` with CloudFetch enabled — uses `StreamingChunkProvider` with `row_offset`
- [ ] Unit tests pass (core module: 0 failures)

NO_CHANGELOG=true


src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkDownloadTask.java
```diff
@@ -95,18 +114,28 @@ public Void call() throws DatabricksSQLException {
...
         }
       }
+    } catch (Throwable t) {
+      uncaughtException = t;
+      throw t;
     } finally {
       if (downloadSuccessful) {
         chunk.getChunkReadyFuture().complete(null);
       } else {
+        LOGGER.info(
```
gopalldb: this should be at error level

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSet.java
```diff
@@ -942,8 +944,15 @@ public boolean isBeforeFirst() throws SQLException {
...
     // Account for client-side maxRows truncation
-    return truncatedByMaxRows
-        || executionResult.getCurrentRow() >= resultSetMetaData.getTotalRows();
+    if (truncatedByMaxRows) {
+      return true;
+    }
+    // Bounded SEA API: manifest.total_row_count is not populated, so use hasNext()
+    // which derives end-of-stream from next_chunk_index via the chunk provider.
+    if (boundedSeaApiEnabled && executionResult instanceof ArrowStreamResult) {
+      return executionResult.getCurrentRow() >= 0 && !executionResult.hasNext();
```
vikrantpuppala: **`isAfterLast()` returns `true` while the cursor is ON the last row — drops the final row.**

The bounded-SEA branch of `isAfterLast()` (:953) uses the exact same predicate as the bounded branch of `isLast()` (:997): `executionResult.getCurrentRow() >= 0 && !executionResult.hasNext()`.

Because `ArrowStreamResult.next()` returns early (`if (!hasNext()) return false;`) **before** incrementing `currentRowIndex`, `getCurrentRow()` holds the last row's index both while positioned on the last row and after a `next()` that returned false. Once the cursor reaches the last row, `hasNext()` is already `false`, so `isAfterLast()` returns `true` while still on the last row — and `isLast()` returns `true` for the same state. The two are meant to be mutually exclusive there.

Impact: the common cursor loop `while (!rs.isAfterLast()) { rs.next(); process(rs); }` exits before processing the last row, silently dropping it. The integration test only checks `isAfterLast()` after the `while(rs.next())` loop has fully exited, so it never exercises the on-last-row state.

<details>
<summary><strong>Suggested fix</strong></summary>

make `isAfterLast()` mean "a prior `next()` returned false" — track an explicit exhausted/past-end flag rather than reusing the `isLast()` expression.

</details>
---

src/test/java/com/databricks/jdbc/integration/fakeservice/tests/BoundedSeaApiCloudFetchIntegrationTests.java
```diff
@@ -0,0 +1,100 @@
...
+      }
+
+      assertEquals(maxRows, rowCount);
+      assertTrue(metaData.getIsCloudFetchUsed());
+      assertTrue(rs.isAfterLast());
+    }
+  }
+
+  @Test
+  void testBoundedSeaApiMultiChunkQuery() throws SQLException {
```
vikrantpuppala, line 68: **The headline `row_offset` behavior has zero test coverage.**

The PR's stated core change — `GetResultData` sends `?row_offset=<offset>` (`DatabricksSdkClient.getResultChunks`) — is never exercised or asserted. The multi-chunk fixture returns all three chunks' `external_link`s inline in the initial `POST /statements` response with `hasMore=false`, so `getResultChunks` is never invoked and `?row_offset=` is never sent. No WireMock mapping matches `result/chunks/N?row_offset=...` and no test `verify`s the query param.

A bug in the `row_offset` path (wrong offset, missing param, appended when the flag is off, wrong encoding) would ship undetected.

<details>
<summary><strong>Suggested fix</strong></summary>

add a unit test on `DatabricksSdkClient.getResultChunks` (mock `IDatabricksConnectionContext` with `isBoundedSeaApiEnabled()` true/false) asserting the request path contains `?row_offset=<expected>` when enabled and not when disabled. Alternatively add an integration scenario whose chunk fetch is actually triggered (links not fully inline) with a mapping keyed on `withQueryParam("row_offset", ...)` + a `verify`.

</details>
---

src/test/java/com/databricks/jdbc/api/impl/arrow/ArrowStreamResultTest.java
```diff
@@ -97,6 +97,90 @@ public void testInitEmptyArrowStreamResult() throws Exception {
     assertFalse(result.hasNext());
   }
 
+  @Test
+  public void testEmptyChunkProvider() {
```
vikrantpuppala, line 101: **Bounded-SEA `StreamingChunkProvider` selection + `null` totalChunkCount contract is not asserted.**

The new tests cover only the `EmptyChunkProvider` (`total_chunk_count == 0`) path. No test asserts that with `isBoundedSeaApiEnabled() == true` and remote links, `ArrowStreamResult` selects a `StreamingChunkProvider` AND passes `null` for `totalChunkCount` (the "must not depend on `total_chunk_count`" contract). The multi-chunk integration fixture even supplies a valid `total_chunk_count: 3`, so a regression reverting to `resultManifest.getTotalChunkCount()` for the bounded path would still pass every test — then break against real bounded servers that don't populate it.

<details>
<summary><strong>Suggested fix</strong></summary>

add a test with `UseBoundedSeaApi=1` and remote external links asserting the chosen provider is a `StreamingChunkProvider` and that iteration completes correctly when `total_chunk_count` is null/absent (chaining purely on `next_chunk_index`).

</details>
---

src/main/java/com/databricks/jdbc/api/impl/arrow/ArrowStreamResult.java
```diff
@@ -62,19 +62,28 @@ public ArrowStreamResult(
       IDatabricksHttpClient httpClient)
       throws DatabricksSQLException {
     this.session = session;
-    // Check if the result data contains the arrow data inline
-    boolean isInlineArrow = resultData.getAttachment() != null;
-    if (isInlineArrow) {
+
+    Long totalChunkCount = resultManifest.getTotalChunkCount();
+    if (totalChunkCount != null && totalChunkCount == 0) {
```
vikrantpuppala, line 67: **Empty-result short-circuit keys off `total_chunk_count` — the field bounded mode forbids relying on — and is ungated.**

The new empty-result branch routes to `EmptyChunkProvider` only when `totalChunkCount != null && totalChunkCount == 0`. But this PR's own `createRemoteChunkProvider` comment (and :131) asserts that under bounded SEA the driver must not depend on `manifest.total_chunk_count` and passes `null` for it. A real bounded-SEA server that omits `total_chunk_count` for an empty result therefore does NOT hit this fast path — it falls through to `StreamingChunkProvider` with empty links and incurs an extra `GetResultData` RPC to discover end-of-stream. This branch also runs on the **default (non-bounded) path**, while the PR frames all changes as off-by-default. The `testBoundedSeaApiEmptyResult` test masks the gap because its fixture still carries `total_chunk_count: 0`.

<details>
<summary><strong>Suggested fix</strong></summary>

either gate the empty short-circuit on `isBoundedSeaApiEnabled()` to match the stated contract, or (if unconditional is intended) add a code comment saying so and derive the bounded empty case from the link stream rather than `total_chunk_count`; make the empty-result integration fixture omit `total_chunk_count` so it exercises the real contract.

</details>
---

src/test/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProviderTest.java
```diff
@@ -1670,4 +1670,143 @@ void testConcurrentRefreshCoalescesIntoSingleRpc() throws Exception {
...
+      // Refresh for chunk 1: returns the fresh link with hasMore=false (end of stream)
+      ExternalLink freshLink1 = createExternalLink(1, rowsPerChunk, null, FAR_FUTURE_EXPIRATION);
+      ChunkLinkFetchResult refetchBatch =
+          ChunkLinkFetchResult.of(Collections.singletonList(freshLink1), false, -1, 200L);
+      when(mockLinkFetcher.fetchLinks(eq(1L), eq(100L))).thenReturn(refetchBatch);
+
+      // Prefetch for chunk 2 should NOT be called since refresh discovered end-of-stream
+      lenient()
+          .when(mockLinkFetcher.fetchLinks(eq(2L), anyLong()))
+          .thenThrow(new AssertionError("Prefetch should not fetch chunk 2 after end-of-stream"));
```
vikrantpuppala: **`thenThrow(new AssertionError(...))` prefetch guards are inert — they cannot fail the test.**

`testRefreshSetsEndOfStream` (:1742) and `testRefreshAdvancesFetchPosition` (:1790) install `thenThrow(new AssertionError(...))` as a guard that prefetch must not fetch a given chunk. But that `fetchLinks` call runs on the background prefetch thread, and `AssertionError` is a `Throwable` — it is not caught by `linkPrefetchLoop`'s `catch(Exception)`, so it silently kills the prefetch thread without failing the main test thread. If a regression caused prefetch to fetch the guarded chunk, these tests would still pass. The intended negative assertion is a no-op.

<details>
<summary><strong>Suggested fix</strong></summary>

replace the `thenThrow(AssertionError)` guards with an explicit `verify(mockLinkFetcher, never()).fetchLinks(eq(2L), anyLong())` (and `eq(3L)`) after consumption completes.

</details>
---

vikrantpuppala:

**Score**: 51/100 — HIGH RISK

Bounded SEA CloudFetch support is off-by-default, but two verified High-severity correctness bugs affect the shared StreamingChunkProvider / ResultSet paths (not just bounded mode): a lost-wakeup that hangs `ResultSet.next()` at end-of-stream, and `isAfterLast()` returning true while on the last row (drops the final row in cursor loops). The headline `row_offset` behavior has no test coverage. All 9 reviewers delivered; security found nothing.

---

Could not post inline comments for: F1, F7 — see body below.

## Critical & High Findings

### F1 (High) — `src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java`:L443

**Consumer `ResultSet.next()` can hang indefinitely at end-of-stream (lost wakeup).**

`fetchNextLinkBatch()` sets `endOfStreamReached = true` here (:443) and at :456, and `getRefreshedLink` at :676 — but none of these call `notifyChunkCreated()`. Meanwhile `waitForNextChunkEnqueued()` (:729) parks on the `chunkCreated` condition while `!closed && consumptionOrder.isEmpty() && !endOfStreamReached`. The EOS flag is written *outside* `prefetchLock` with no signal, so this is a classic lost-wakeup:

1. Consumer drains the last enqueued chunk, polls an empty `consumptionOrder`.
2. Checks `endOfStreamReached` — still `false` (the final `hasMore=false` link-fetch RPC is still in flight).
3. Enters `waitForNextChunkEnqueued`, re-checks the guard under the lock (still false), calls `chunkCreated.await()`.
4. Prefetch thread completes the RPC, sets `endOfStreamReached = true`, and exits the loop **without signaling**.
5. The parked consumer wakes only on `close()` → `rs.next()` hangs until the statement/connection is closed.

This affects the whole `StreamingChunkProvider` path (`EnableStreamingChunkProvider` *or* `UseBoundedSeaApi`), not only bounded mode.

- **Suggested fix**: call `notifyChunkCreated()` immediately after every `endOfStreamReached = true` (both branches in `fetchNextLinkBatch` and the one in `getRefreshedLink`), OR change `waitForNextChunkEnqueued` to `chunkCreated.await(timeout, unit)` with a bounded poll so the loop guard is re-checked periodically.

## Medium Findings

### F7 (Medium) — `src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java`:L589

**Lock-scope creep: refresh reconciliation runs `createChunkFromLink` + state mutation + `triggerDownloads()` all under `refetchLock`.**

`getRefreshedLink` (:589-700) now, while holding `synchronized (refetchLock)`: creates chunks (`createChunkFromLink`, which enqueues into `consumptionOrder`/`downloadOrder` and updates counters), mutates `endOfStreamReached`/`nextFetchPosition`, then calls `triggerDownloads()` (which acquires `downloadLock` and submits work to the bounded download executor). Every other download thread that hits an expired link must wait on `refetchLock`, so a slow `triggerDownloads` now serializes all concurrent expired-link refreshers, reducing the download parallelism the streaming provider is designed for. The pre-change version only refreshed link objects under this lock.

- **Suggested fix**: move `triggerDownloads()` (and, if the atomicity of `endOfStreamReached` isn't required under the lock, the position mutation) outside the `synchronized (refetchLock)` block.

## Low Findings

<details>
<summary>Low findings (5) — click to expand</summary>

### F9 (Low)

**Thread-context capture is `null` for prefetch-triggered downloads (defeats propagation intent).** `DatabricksThreadContextHolder` uses a plain `ThreadLocal` (not `InheritableThreadLocal`); `linkPrefetchLoop` never calls `setConnectionContext`/`setStatementId`; and `StreamingChunkDownloadTask`'s constructor reads `getConnectionContext()`/`getStatementId()` on whatever thread submits the task. Most chunks in a multi-chunk query are submitted from the prefetch thread → captured context is `null` → download-thread telemetry/logging loses connection/statement attribution for exactly the streaming scenario this feature targets. Telemetry-only (not a correctness bug). Fix: capture the context once in the `StreamingChunkProvider` constructor (consumer thread) and pass it explicitly into each `StreamingChunkDownloadTask`.

### F10 (Low)

**`getChunkCount()` semantics silently changed and are ungated.** Changed from `highestKnownChunkIndex.get() + 1` to `consumedChunkCount.get() + consumptionOrder.size()`. This is shared with the non-bounded `StreamingChunkProvider` path (`isStreamingChunkProviderEnabled()`), differs from the old value under sparse/non-contiguous server indices, and is only stable once all chunks are enqueued at end-of-stream. Consumers (telemetry, existing test assertions, `DatabricksResultSet` readers) that assumed the old formula observe a different number mid-stream. Observability impact only. Consider a targeted test for the sparse-index case (the added tests use contiguous indices 0–6, so they don't exercise the divergence the new formula was introduced to handle).

### F11 (Low)

**Stale comment in `StreamingChunkProvider.java` (~:90).** The "Position tracking" block header reads "Using AtomicLong for single-writer variables to make thread-safety explicit:" and lists `currentChunkIndex` first, but `currentChunkIndex` is now declared `private volatile long` (:93), and the newly-added `AtomicLong consumedChunkCount` is not documented in the block. Update the comment to reflect the actual field types (volatile long vs AtomicLong) and add `consumedChunkCount`.

### F12 (Low)

**`StreamingChunkDownloadTask.call()` is a near-verbatim copy of `ChunkDownloadTask.call()`.** The retry loop (`MAX_RETRIES`/`RETRY_DELAY_MS`), thread-context capture/set/clear, `catch(Throwable)`, and the complete/completeExceptionally `finally` protocol are duplicated across both files; they differ almost only in the link-refresh call and the marker interface. The next fix to this protocol (e.g. F4's log-level change) is now shotgun surgery in two places and they will drift. Consider extracting the shared skeleton into an abstract base / template method taking the download-step and refresh-step as the only varying operations — a judgement call at two call sites, but at minimum cross-reference the sibling in a comment.

</details>
## Agreed with prior comments

- Agreed with [prior comment](https://github.com/databricks/databricks-jdbc/pull/1468#discussion_r3478499794) by @gopalldb on `src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkDownloadTask.java`:L124 — terminal download failures logged at info level, losing stack traces in production environments.

src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java
```diff
@@ -654,6 +725,32 @@ private void releaseChunk(long chunkIndex) {
     }
   }
 
+  /** Waits until the prefetch thread enqueues a chunk index into the consumption queue. */
+  private void waitForNextChunkEnqueued() throws DatabricksSQLException {
```
vikrantpuppala, line 733: **Consumer `ResultSet.next()` can hang indefinitely at end-of-stream (lost wakeup).**

This method parks on the `chunkCreated` condition while `!closed && consumptionOrder.isEmpty() && !endOfStreamReached`. But the sites that set `endOfStreamReached = true` on the clean end-of-stream path — `fetchNextLinkBatch()` (`isEndOfStream()` at :443 and `hasMore=false` at :456) and `getRefreshedLink` (:676) — do **not** call `notifyChunkCreated()`, and they write the flag *outside* `prefetchLock`. That's a classic lost-wakeup:

1. Consumer drains the last enqueued chunk, polls an empty `consumptionOrder`.
2. Checks `endOfStreamReached` — still `false` (the final `hasMore=false` link-fetch RPC is in flight).
3. Enters `waitForNextChunkEnqueued`, re-checks the guard under the lock (still false), calls `chunkCreated.await()`.
4. Prefetch thread completes the RPC, sets `endOfStreamReached = true`, exits the loop **without signaling**.
5. The parked consumer wakes only on `close()` → `rs.next()` hangs until the statement/connection is closed.

This affects the whole `StreamingChunkProvider` path (`EnableStreamingChunkProvider` *or* `UseBoundedSeaApi`), not only bounded mode. (Re-posted here because line 443 was not on a diffed line; the fix touches the EOS-set sites listed above.)

<details>
<summary><strong>Suggested fix</strong></summary>

call `notifyChunkCreated()` immediately after every `endOfStreamReached = true` (both branches in `fetchNextLinkBatch` and the one in `getRefreshedLink`), OR change this method to `chunkCreated.await(timeout, unit)` with a bounded poll so the loop guard is re-checked periodically.

</details>
---

gopalldb:
Fixed. Added `boundedSeaExhausted` flag (set only when `next()` returns false) so `isAfterLast()` returns true only after the cursor moves past the last row, not while positioned on it. The `while (!rs.isAfterLast()) { rs.next(); process(); }` loop now works correctly — committed in `Address PR review comments on bounded SEA CloudFetch`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

gopalldb:
Fixed in two parts:
1. `ArrowStreamResult.convertToChunkLinkFetchResult`: when `externalLinks` is empty and `totalChunkCount` is null (bounded-SEA path), we now return `ChunkLinkFetchResult.endOfStream()` rather than `null`, so `StreamingChunkProvider` discovers end-of-stream from initial links without issuing an extra `GetResultData` RPC.
2. Removed `total_chunk_count` from the `testBoundedSeaApiEmptyResult` WireMock fixture so the test now exercises the real bounded-SEA empty path with no `total_chunk_count` reliance — committed in `Fix empty-result bounded-SEA path and add row_offset test coverage`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

gopalldb:
Fixed — changed `LOGGER.info` to `LOGGER.error` in the download failure path. Committed in `Address PR review comments on bounded SEA CloudFetch`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

gopalldb:
Fixed. Added `notifyChunkCreated()` immediately after every `endOfStreamReached = true` in `fetchNextLinkBatch` (both branches), `getRefreshedLink`, and `processInitialLinks`. Consumer parked in `waitForNextChunkEnqueued` now wakes up as soon as end-of-stream is set rather than waiting for `close()`. Committed in `Address PR review comments on bounded SEA CloudFetch`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

gopalldb:
Fixed. Added `testBoundedSeaApiUsesStreamingChunkProviderWithNullTotalChunkCount` in `ArrowStreamResultTest` — sets `UseBoundedSeaApi=1`, explicitly leaves `totalChunkCount` null, asserts iteration completes correctly. A regression reverting to `getTotalChunkCount()` on the bounded path would fail this test. Committed in `Address PR review comments on bounded SEA CloudFetch`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

gopalldb:
Fixed. Replaced `thenThrow(new AssertionError(...))` guards with proper `verify(mockLinkFetcher, never()).fetchLinks(eq(2L), anyLong())` after consumption in `testRefreshSetsEndOfStream`. For `testRefreshAdvancesFetchPosition` the race between prefetch and refresh is inherently non-deterministic so the guard was replaced with a lenient stub that produces the same end-state regardless of which thread wins, and the correctness is asserted via the consumed chunk count and a `verify` that `fetchLinks(5L, 500L)` was called. Committed in `Address PR review comments on bounded SEA CloudFetch`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

gopalldb:
Fixed. Added two unit tests in `DatabricksSdkClientTest` using `ArgumentCaptor<Request>` to assert the request path:
- `testGetResultChunks_boundedSeaEnabled_appendsRowOffset`: verifies `?row_offset=<offset>` is appended when `UseBoundedSeaApi=1`.
- `testGetResultChunks_boundedSeaDisabled_noRowOffset`: verifies `row_offset` is absent on the default path.
Committed in `Fix empty-result bounded-SEA path and add row_offset test coverage`.

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

vikrantpuppala (approved):
LGTM — all 3 High findings (EOS lost-wakeup hang, isAfterLast dropping the last row, missing row_offset test coverage) and the substantive Medium findings from the review are addressed and verified against the current head. Remaining items (F7 lock-scope, minor Low nits) are acceptable follow-ups.

Note: branch currently has merge conflicts with main and CI hasn't reported on this head yet — please rebase and confirm CI is green before merging.

PR #1473, opened 2026-05-29
Fix metadata identifiers containing backticks

### LINKED ISSUE #1471: [BUG] getPrimaryKeys fails when catalog, schema or table name contains backtick
**Describe the bug**
When not in thrift mode, `DatabaseMetaData#getPrimaryKeys` fails when catalog, schema, or table identifiers contain backtick

**To Reproduce**
Call `DatabaseMetaData#getPrimaryKeys` on the table which either has name with backtick, or belongs to catalog or schema with backtick in name

**Expected behavior**
Call should not fail and should return table's primary keys

**Client side logs**
```
[PARSE_SYNTAX_ERROR] org.apache.spark.sql.catalyst.parser.ParseException: 
[PARSE_SYNTAX_ERROR] Syntax error at or near 'table_with'. SQLSTATE: 42601 (line 1, pos 79)

== SQL ==
SHOW KEYS IN CATALOG `my_workspace` IN SCHEMA `default` IN TABLE `test`table_with'special"chars`
-----------------------------------------------------------------------^^^
```

**Client Environment**
 - OS: Ubuntu 25.10
 - Java version: Java 21
 - Java vendor: OpenJDK
 - Driver Version: 3.4.1

**Additional context**
Issue is presumably in `com.databricks.jdbc.dbclient.impl.sqlexec.CommandBuilder#fetchPrimaryKeysSQL` which constructs the query like this
```
return String.format(SHOW_PRIMARY_KEYS_SQL, catalogName, schemaName, tableName);
```
where `SHOW_PRIMARY_KEYS_SQL` is declared as
```
  public static final String SHOW_PRIMARY_KEYS_SQL =
      "SHOW KEYS" + IN_CATALOG_SQL + IN_ABSOLUTE_SCHEMA_SQL + IN_ABSOLUTE_TABLE_SQL;
```
which evaluates to
```
SHOW KEYS IN CATALOG `%s` IN SCHEMA `%s` IN TABLE `%s`
```

The same issue affects multiple commands constructed in `CommandBuilder`.

This issue is supposedly present in earlier versions of driver as well, however starting from 3.4.1 thrift mode seem to be off by default.

msrathore-db:
## Summary
- Escape literal backticks before interpolating catalog, schema, and table identifiers into metadata SQL commands.
- Apply the same escaping to information_schema catalog prefixes used by procedure metadata.
- Add regression coverage for primary keys, foreign keys, other metadata commands, and information_schema catalog prefixes.

## Test plan
- mvn test -pl jdbc-core -Dtest=CommandBuilderTest,CommandConstantsTest

Fixes #1471

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 - Fixed `setCatalog()` and `setSchema()` producing invalid SQL (e.g. `SET CATALOG ``name``) when the catalog or schema name was passed already wrapped in backticks. Backticks are now stripped before wrapping, and `getCatalog()`/`getSchema()` return the bare identifier name.
+- Fixed metadata SQL generation for catalog, schema, and table identifiers containing backticks.
 
 ---
 *Note: When making changes, please add your change under the appropriate section
--- a/src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
+++ b/src/main/java/com/databricks/jdbc/dbclient/impl/common/CommandConstants.java
@@ -132,7 +132,11 @@ public class CommandConstants {
         .build();
   }
 
+  public static String escapeSqlIdentifier(String identifier) {
+    return identifier == null ? null : identifier.replace("`", "``");
+  }
+
   private static String getCatalogPrefix(String catalog) {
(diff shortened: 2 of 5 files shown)


msrathore-db:
Superseded by #1474, which uses a branch pushed directly to databricks/databricks-jdbc.

PR #944, opened 2025-08-16
Implement multi-row INSERT batching for PreparedStatement

josecsotomorales:
Linked issue: https://github.com/databricks/databricks-jdbc/issues/867

This PR implements multi-row INSERT batching optimization for prepared statements to improve performance when executing large batches of INSERT operations. The implementation combines multiple single-row INSERT statements into fewer multi-row INSERT statements while respecting Databricks' 256 parameter limit.

Adds a new InsertStatementParser utility for parsing INSERT statements and generating multi-row equivalents
Optimizes executeBatch() and executeLargeBatch() to use multi-row INSERT when possible
Includes parameter limit-aware chunking to handle large batches that exceed the 256 parameter maximum

Impact illustration (10k rows, 5 columns, 50 ms RTT):
• Before (single-row inserts): 10,000 statements → ~500s of RTT + server planning.
• After (batched): 196 statements (10k ÷ 51) → ~9.8s of RTT.
• That’s about a 50× reduction in latency, not even counting server CPU savings.

Signed-off-by: josecsotomorales <josecsmorales@gmail.com>, Jayant Singh <jayant.singh@databricks.com>

PR #1496, opened 2026-06-17
Fix supportsBatchUpdates() to honor EnableBatchedInserts

sreekanth-db:
## Summary

`DatabricksDatabaseMetaData.supportsBatchUpdates()` was hard-coded to return `false`. Batch-aware JDBC clients (e.g. Apache Hop) call this method *before* batching; when it returns `false` they never call `executeBatch()` and instead fall back to one `executeUpdate()` per row. As a result, the multi-row INSERT optimization gated by `EnableBatchedInserts` (added in #944) never runs — even when the customer explicitly sets `EnableBatchedInserts=1`.

This PR makes `supportsBatchUpdates()` return `true` when `EnableBatchedInserts=1`, so batch-aware clients take the optimized `executeBatch()` path.

Fixes ES-1963495.

## Why gate on `EnableBatchedInserts` rather than return `true` unconditionally

- **Zero change for existing users.** With the flag off (the default), the method still returns `false`, so default behavior is untouched. This preserves the deliberate "safe rollout" posture chosen in #944.
- **Advertises support exactly when batching actually engages.** With the flag on, `executeBatch()` routes INSERTs through the optimized multi-row path; the metadata answer now matches real behavior.
- **Consistent with existing code.** `supportsTransactions()` in the same class is already derived from a connection parameter (`IgnoreTransactions`), so a context-driven capability method is an established pattern here.

## Spec / conformance notes

- The JDBC spec states batch support is optional, but *if* a driver supports it, `supportsBatchUpdates()` must return `true`. The driver does implement a working `executeBatch()` (it handles all DML types; only INSERTs are optimized, others fall back to per-row), so `true` is the conformant value.
- The spec imposes no per-statement-type requirement and does not require atomicity — non-atomic, app-managed (auto-commit) batches are the expected model. INSERT-only optimization does not preclude returning `true`.
- This matches MySQL Connector/J and PostgreSQL pgjdbc, which both return `true` while gating their multi-row INSERT rewrite behind a separate flag (`rewriteBatchedStatements` / `reWriteBatchedInserts`, both default-off).

## Changes

- `DatabricksDatabaseMetaData.supportsBatchUpdates()` returns `isBatchedInsertsEnabled()` instead of `false`.
- Added test `supportsBatchUpdates_returnsTrueWhenBatchedInsertsEnabled`; renamed the existing case to `supportsBatchUpdates_returnsFalseByDefault`.
- `NEXT_CHANGELOG.md` entry under `### Fixed`.

## Testing

`mvn test -pl jdbc-core -Dtest=DatabricksDatabaseMetaDataTest` → 249 run, 0 failures, 0 skipped (both new/updated cases included).

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
@@ -1278,7 +1278,9 @@ public class DatabricksDatabaseMetaData implements DatabaseMetaData {
   public boolean supportsBatchUpdates() throws SQLException {
     LOGGER.debug("public boolean supportsBatchUpdates()");
     throwExceptionIfConnectionIsClosed();
-    return false;
+    // Advertise batch support only when the multi-row INSERT optimization is enabled, so
+    // batch-aware clients use executeBatch() instead of one executeUpdate() per row.
+    return session.getConnectionContext().isBatchedInsertsEnabled();
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksDatabaseMetaData.java
```diff
@@ -1278,7 +1278,9 @@ public boolean insertsAreDetected(int type) throws SQLException {
   public boolean supportsBatchUpdates() throws SQLException {
     LOGGER.debug("public boolean supportsBatchUpdates()");
     throwExceptionIfConnectionIsClosed();
-    return false;
+    // Advertise batch support only when the multi-row INSERT optimization is enabled, so
+    // batch-aware clients use executeBatch() instead of one executeUpdate() per row.
+    return session.getConnectionContext().isBatchedInsertsEnabled();
   }
```
gopalldb, line 1284: once we add batch insert support, we will need to set this to true. cc: @msrathore-db

gopalldb (approved):

Issue #1505, opened 2026-06-23
[BUG]map<int,array<bigint>> column not selected properly in 3.3.1 with EnableArrow=1


Ansssss:
**Describe the bug**
Selecting a column of type MAP<int, array<bigint>> with EnableArrow=1 in the connection string causes the values of the column to be malformed. When EnableArrow=0, the bug does not manifest.

**To Reproduce**
Using a client like DataGrip, with the connection string having ";EnableArrow=1" in it, run this query:
`SELECT MAP(0, ARRAY(34277,0))`

**Expected behavior**
Expected result `{0:[34277,0]}`
Actual result      `{0:}`

**Client Environment (please complete the following information):**
 - OS: Mac
 - Driver Version 3.3.1
 - Client = Datagrip

rileythomp:
@gopalldb Can I work on this?

gopalldb:
@rileythomp , please go ahead

PR #1506, opened 2026-06-23
Fix MAP columns dropping complex values with EnableArrow=1 (#1505)

rileythomp:
## Summary

With `EnableArrow=1` (and complex datatype support disabled, which is the default), selecting a `MAP` column whose values are complex types returned malformed data — the map **value** was dropped while the key survived:

```
SELECT MAP(0, ARRAY(34277,0))
-- expected: {0:[34277,0]}
-- actual:   {0:}
```

With `EnableArrow=0` the server returns the value pre-formatted as a string, so only the Arrow path was affected.

## Root cause

When complex datatype support is disabled, `ArrowStreamResult.getObjectWithComplexTypeHandling` reads the map as raw Arrow JSON (entry-array form, e.g. `[{"key":0,"value":[34277,0]}]`) and formats it via `ComplexDataTypeParser.formatMapString`. That hand-rolled formatter called `JsonNode.asText()` on the value node, which returns `""` for any container (array/object) node — so array/struct/nested-map values were rendered empty.

## Fix

Delegate map formatting to the **existing recursive** parser: `parseToMap(node, mapMetadata).toString()`. `parseToMap` builds a `DatabricksMap` whose nested values are recursively converted to `DatabricksArray` / `DatabricksMap` / `DatabricksStruct`, and their `toString()` methods already emit the canonical `{key:value}` / `[a,b]` format. This:

- Produces `{0:[34277,0]}` for the reported case.
- Reproduces byte-for-byte identical output for scalar maps (verified against all existing `formatMapString` tests).
- Covers **all** map-rooted columns with arbitrary nesting (arrays, structs, nested maps, deep nesting).

Also hardened `DatabricksMap.convertValue` and `DatabricksArray.convertElements` against `null` nested values (previously threw an NPE on `value.getClass()`), so e.g. a null array value in a map now renders as `null`.

## Scope / follow-up

Scoped to MAP-rooted columns (the reported bug). Top-level `ARRAY`/`STRUCT` columns whose elements/fields are themselves maps still pass through raw Arrow output — a separate latent gap that can be addressed in a follow-up if desired.

## Test plan

- New unit tests in `ComplexDataTypeParserTest` covering the exact repro plus map-of-arrays, map-of-struct, nested-map, and null-complex-value cases, and the existing scalar-map regressions.
- Strengthened the disabled-support assertions in the `testMapOfArrays` e2e test so dropped array values would be caught.
- `mvn test -pl jdbc-core` passes (3493 run; the single unrelated failure is a pre-existing DST/timezone test, `Slf4jFormatterTest`, expecting `EST` for a June date). `spotless:check` is clean.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Updated
 
 ### Fixed
+- Fixed `MAP` columns whose values are complex types (`ARRAY`/`STRUCT`/`MAP`) returning empty values (e.g. `{0:}` instead of `{0:[34277,0]}`) with `EnableArrow=1` when complex datatype support is disabled. The string formatter now reproduces nested values correctly; null nested values inside maps and arrays are also handled (#1505).
 - Fixed `setCatalog()` and `setSchema()` producing invalid SQL (e.g. `SET CATALOG ``name``) when the catalog or schema name was passed already wrapped in backticks. Backticks are now stripped before wrapping, and `getCatalog()`/`getSchema()` return the bare identifier name.
 - Fixed metadata SQL generation for catalog, schema, and table identifiers containing backticks.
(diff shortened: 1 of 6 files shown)


msrathore-db:
Can you please move the PR to ready for review if you want this to be reviewed

PR #1509, opened 2026-06-24
Stream CloudFetch chunk decompression to bound memory and prevent OOM on large results

### LINKED ISSUE #1508: [BUG]In 3.x versions fetch size is ignored resulting in out-of-memory when downloading query results
**Describe the bug**
When using the 3.x versions of the Databricks JDBC driver, we encountered an issue where the the heap was exhausted. This didn't happen in the 2.x versions of the driver.

**To Reproduce**
Roughly the steps are:  

1. to create a volume table with spoofed data
2. run the following test program to download from that table (passing in the table location as an argument)
3. see that the test succeeds with 2.8.1 and fails with 3.4.1

To spoof the table:
```
CREATE TABLE delta.`/Volumes/<PLACEHOLDER_PATH>/delete_me_12345` AS
  SELECT
    cast(rand() as STRING) AS col1,
    rand() AS col2,
    rand() AS col3,
    rand() AS col4,
    rand() AS col5,
    current_date() AS col6,
    rand() AS col7,
    (rand() > 0.5) AS col8,
    rand() AS col9,
    rand() AS col10,
    rand() AS col11,
    cast(rand() * 100 as BIGINT) AS col12,
    rand() AS col13,
    cast(rand() as STRING) AS col14,
    timestampadd(SECOND, cast(rand() * 3600 as INT), now()) AS col15,
    timestampadd(SECOND, cast(rand() * 3600 as INT), now()) AS col16,
    cast(rand() as STRING) AS col17,
    cast(rand() as STRING) AS col18,
    (rand() > 0.5) AS col19,
    cast(rand() as STRING) AS col20,
    cast(rand() as STRING) AS col21,
    cast(rand() as STRING) AS col22,
    cast(rand() as STRING) AS col23,
    cast(rand() as STRING) AS col24,
    cast(rand() as STRING) AS col25,
    cast(rand() as STRING) AS col26
  FROM range(169769)
```

Create the following OomReproducer.java file:
```
import java.sql.*;

/**
 * Streams every row from a Databricks volume table and prints memory usage per batch.
 *
 * Usage: see run.sh
 */
public class OomReproducer {

    public static void main(String[] args) throws Exception {
        String host          = require("host");
        String warehouseId   = require("warehouseId");
        String token         = require("token");
        String tableName     = require("tableName");
        int    batchSize     = Integer.parseInt(System.getProperty("batchSize", "1000"));
        boolean enableArrow  = Boolean.parseBoolean(System.getProperty("enableArrow", "false"));
        // oauth: embed token via AuthMech=11 (2.x drivers); pat: pass as JDBC password (3.x+)
        boolean oauthMode    = "oauth".equals(System.getProperty("authMode", "pat"));

        String arrow   = enableArrow ? "" : ";EnableArrow=0";
        String baseUrl = "jdbc:databricks://" + host + ":443;HttpPath=/sql/1.0/warehouses/" + warehouseId + arrow;
        String jdbcUrl = oauthMode ? baseUrl + ";AuthMech=11;Auth_Flow=0;Auth_AccessToken=" + token : baseUrl;

        System.out.println("Driver version : " + driverVersion());
        System.out.println("JDBC URL       : " + baseUrl + (oauthMode ? ";AuthMech=11;Auth_Flow=0;Auth_AccessToken=<redacted>" : ""));
        System.out.println("Auth mode      : " + (oauthMode ? "oauth (AuthMech=11)" : "pat (password)"));
        System.out.println("Table          : " + tableName);
        System.out.println("Batch size     : " + batchSize);
        System.out.println("EnableArrow    : " + enableArrow);
        System.out.println("Max heap       : " + (Runtime.getRuntime().maxMemory() / 1024 / 1024) + " MB");
        System.out.println();

        Runtime rt = Runtime.getRuntime();
        long startMs = System.currentTimeMillis();

        Connection conn = oauthMode
            ? DriverManager.getConnection(jdbcUrl)
            : DriverManager.getConnection(jdbcUrl, "token", token);
        try (conn) {
            System.out.printf("[%5ds] Connected          heap used: %d MB%n", elapsed(startMs), usedMb(rt));

            PreparedStatement ps = conn.prepareStatement("SELECT * FROM IDENTIFIER(?)", ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
            ps.setString(1, tableName);
            ps.setFetchSize(batchSize);

            System.out.printf("[%5ds] Executing query   heap used: %d MB%n", elapsed(startMs), usedMb(rt));
            ResultSet rs = ps.executeQuery();
            System.out.printf("[%5ds] Query returned    heap used: %d MB%n", elapsed(startMs), usedMb(rt));

            long rows   = 0;
            long peakMb = usedMb(rt);

            while (rs.next()) {
                rows++;
                if (rows % batchSize == 0) {
                    long mb = usedMb(rt);
                    if (mb > peakMb) peakMb = mb;
                    System.out.printf("[%5ds] rows: %,8d    heap used: %d MB%n", elapsed(startMs), rows, mb);
                }
            }

            System.out.printf("%n[%5ds] DONE  total rows: %,d    peak heap: %d MB%n",
                elapsed(startMs), rows, peakMb);
        }
    }

    private static String require(String prop) {
        String v = System.getProperty(prop);
        if (v == null) {
            System.err.println("Missing required property: -D" + prop + "=...");
            System.err.println("See run.sh for usage.");
            System.exit(1);
        }
        return v;
    }

    private static long usedMb(Runtime rt) {
        return (rt.totalMemory() - rt.freeMemory()) / 1024 / 1024;
    }

    private static long elapsed(long startMs) {
        return (System.currentTimeMillis() - startMs) / 1000;
    }

    private static String driverVersion() {
        try {
            return Class.forName("com.databricks.client.jdbc.Driver")
                        .getPackage().getImplementationVersion();
        } catch (ClassNotFoundException e) {
            return "(unknown)";
        }
    }
}
```

Create the run.sh script:
```
#!/usr/bin/env bash
# Run the OOM reproducer against a Databricks table.
#
# Usage:
#   ./run.sh <driver-version> <max-heap> [extra -D flags...]
#
# Required -D flags (pass after driver version):
#   -Dhost=adb-XXXXX.databricks.com
#   -DwarehouseId=XXXXXXXXXXXXXXXX
#   -Dtoken=dapiXXXXXXXXXXXXXXXX
#   -DtableName=delta.\`/Volumes/.../myTable\`
#
# Optional -D flags:
#   -DbatchSize=1000              (default 1000)
#   -DenableArrow=false           (default false)
#
# Examples — compare driver versions against the same table:
#   ./run.sh 2.8.1 128m -Dhost=... -DwarehouseId=... -Dtoken=... -DtableName=...
#   ./run.sh 3.4.1 128m -Dhost=... -DwarehouseId=... -Dtoken=... -DtableName=...
#
# Flip arrow mode on/off:
#   ./run.sh 2.8.1 -DenableArrow=true ...

set -euo pipefail

DRIVER_VERSION="${1:?first arg must be driver version, e.g. 2.8.1}"
HEAP="${2:?second arg must be max heap, e.g. 512m}"
shift 2

JAR="databricks-jdbc-${DRIVER_VERSION}.jar"
MAVEN_URL="https://repo1.maven.org/maven2/com/databricks/databricks-jdbc/${DRIVER_VERSION}/${JAR}"

if [[ ! -f "$JAR" ]]; then
    echo "Downloading ${JAR}..."
    curl -fSL "$MAVEN_URL" -o "$JAR"
fi

# Older 2.x drivers require OAuth tokens to be embedded in the URL via AuthMech=11;
# newer drivers accept them as a standard JDBC password.
DRIVER_MAJOR="${DRIVER_VERSION%%.*}"
if [[ "$DRIVER_MAJOR" == "2" ]]; then
    AUTH_FLAG="-DauthMode=oauth"
else
    AUTH_FLAG="-DauthMode=pat"
fi

echo "Compiling..."
javac -cp "$JAR" OomReproducer.java

echo "Running (heap=${HEAP}, driver=${DRIVER_VERSION})..."
echo "---"
java -Xmx"${HEAP}" -cp ".:${JAR}" "$AUTH_FLAG" "$@" OomReproducer
```

Acquire your token 
`databricks auth token --host <PLACEHOLDER host name>`

Run against the 2.8.1 driver version:
```
./run.sh 2.8.1 128m -Dhost=PLACEHOLDER -DwarehouseId=PLACEHOLDER -Dtoken=PLACEHOLDER -DtableName="delta.\`/Volumes/<PLACEHOLDER_PATH>/delete_me_12345\`"
```

Run against the 3.4.1 driver version:
```
./run.sh 3.4.1 128m -Dhost=PLACEHOLDER -DwarehouseId=PLACEHOLDER -Dtoken=PLACEHOLDER -DtableName="delta.\`/Volumes/<PLACEHOLDER_PATH>/delete_me_12345\`"
```

**Expected behavior**
It is expected that running against the 3.4.1 driver version will succeed like it does for 2.8.1. 
The actual result is that the 3.4.1 version has a failure like so:
```
[    2s] Connected          heap used: 17 MB
[    2s] Executing query   heap used: 18 MB
[    4s] Query returned    heap used: 26 MB
Exception in thread "main" com.databricks.jdbc.exception.DatabricksSQLException: Failed to ready chunk
	at com.databricks.jdbc.api.impl.arrow.AbstractRemoteChunkProvider.getChunk(AbstractRemoteChunkProvider.java:176)
	at com.databricks.jdbc.api.impl.arrow.ArrowStreamResult.next(ArrowStreamResult.java:286)
	at com.databricks.jdbc.api.impl.DatabricksResultSet.next(DatabricksResultSet.java:338)
	at OomReproducer.main(OomReproducer.java:53)
Caused by: com.databricks.jdbc.exception.DatabricksSQLException: Download failed for chunk index 0
	at com.databricks.jdbc.api.impl.arrow.ChunkDownloadTask.call(ChunkDownloadTask.java:135)
	at com.databricks.jdbc.api.impl.arrow.ChunkDownloadTask.call(ChunkDownloadTask.java:16)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.ArrayList.iterator(ArrayList.java:1029)
	at com.databricks.internal.apache.commons.io.output.AbstractByteArrayOutputStream.toByteArrayImpl(AbstractByteArrayOutputStream.java:199)
	at com.databricks.internal.apache.commons.io.output.UnsynchronizedByteArrayOutputStream.toByteArray(UnsynchronizedByteArrayOutputStream.java:174)
	at com.databricks.internal.apache.commons.io.IOUtils.toByteArray(IOUtils.java:2711)
	at com.databricks.jdbc.api.impl.arrow.ArrowResultChunk.downloadData(ArrowResultChunk.java:91)
	at com.databricks.jdbc.api.impl.arrow.ChunkDownloadTask.call(ChunkDownloadTask.java:71)
	at com.databricks.jdbc.api.impl.arrow.ChunkDownloadTask.call(ChunkDownloadTask.java:16)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.runWith(Thread.java:1596)
	... 1 more
```

 - OS: MacOS
 - Java version 21.0.11
 - Java vendor Temurin

msrathore-db:
## Summary

Fixes #1508.

Downloading a large query result via CloudFetch on the 3.x driver could exhaust the JVM heap and throw `java.lang.OutOfMemoryError: Java heap space` (surfaced as `Failed to ready chunk` / `Download failed for chunk index 0`), where the 2.x driver succeeded at the same heap. This was a 2.x → 3.x regression.

## Root cause

The `OutOfMemoryError` is on-heap. Each chunk was decompressed by materializing the **entire decompressed Arrow payload into an on-heap `byte[]`** (several times the compressed size) before parsing. With up to `cloudFetchThreadPoolSize` (default 16) chunks decompressing in parallel, transient on-heap usage scaled with download concurrency rather than with available heap.

(The parsed Arrow vectors themselves are off-heap, so they were not the cause — the transient on-heap decompression buffers were.)

## Fix

**Streamed decompression.** Decompression is now streamed directly into the Arrow reader via `DecompressionUtil.decompressToInputStream` (a lazy `LZ4FrameInputStream`), so the decompressed payload is never materialized on-heap alongside the compressed bytes. Peak per-chunk on-heap cost drops from `compressed + full-decompressed` to `compressed + one LZ4 block`. Chunks are still downloaded, decompressed, and parsed in parallel ahead of consumption — throughput is unaffected.

This is the same `LZ4FrameInputStream` the 2.x Simba driver feeds its decompression through; this change simply pipes it into Arrow instead of buffering the output.

### For constrained heaps

If a workload still hits OOM on a very small heap (many concurrent chunks × per-chunk cost), lower the existing **`cloudFetchThreadPoolSize`** connection property (default 16) to reduce the number of chunks buffered concurrently — this is the memory-vs-throughput lever. No new connection property is introduced.

## Scope note

An earlier revision of this PR also added an in-memory byte-budget scheduler (`cloudFetchMaxBytesInMemory`). It was **removed** after testing showed it was redundant with the existing `cloudFetchThreadPoolSize` limit and its heuristic default did not reliably bind (the manifest reports compressed size, while the OOM is driven by decompressed + parse transients). Streaming decompression is the robust fix; concurrency is already tunable via `cloudFetchThreadPoolSize`.

## Testing

Verified against a SQL warehouse on the issue's 26-column, 169,769-row repro table:

- **Memory:** the unpatched 3.x driver OOMs at `-Xmx128m`; the patched driver streams the full result at `-Xmx128m` (peak ~99 MB), and down to `-Xmx48m` with `cloudFetchThreadPoolSize=1` (peak ~42 MB) — below where the 2.7.5 / 2.8.1 Simba drivers OOM.
- **Correctness:** a cell checksum over the full result is **byte-identical** between the patched driver, the unpatched driver, and the 2.7.5 / 2.8.1 Simba drivers, across SEA/Thrift paths, the LZ4_FRAME and NONE codecs, complex types (struct/array/map, decimal, binary, NULLs), and heap sizes — no truncation or corruption. Verified with a position-sensitive (ordered, chained) row digest.
- **Unit tests:** added streaming-decompression coverage (LZ4 frame + NONE) in `DecompressionUtilTest`; the `jdbc-core` unit suite passes.

NO_CHANGELOG=true

### DIFF
--- a/src/main/java/com/databricks/jdbc/common/util/DecompressionUtil.java
+++ b/src/main/java/com/databricks/jdbc/common/util/DecompressionUtil.java
@@ -53,11 +53,24 @@ public class DecompressionUtil {
     }
   }
 
-  public static InputStream decompressToStream(
+  /**
+   * Returns a stream that decompresses {@code compressedInput} lazily as it is read, so the full
+   * decompressed payload is never materialized alongside the compressed bytes. Only LZ4_FRAME is
+   * wrapped in a decompressing decorator; a {@code null}/NONE codec returns the raw bytes as-is.
+   *
+   * <p>Any {@link IOException} from constructing the LZ4 decoder propagates to the caller ({@code
+   * ArrowResultChunk.downloadData}), which handles it as a chunk-processing failure — the same
+   * terminal path as any other decompression error.
+   */
+  public static InputStream decompressToInputStream(
       byte[] compressedInput, CompressionCodec compressionCodec, String context)
-      throws DatabricksSQLException {
-    byte[] uncompressed = decompress(compressedInput, compressionCodec, context);
-    return new ByteArrayInputStream(uncompressed);
+      throws IOException {
+    if (compressionCodec == CompressionCodec.LZ4_FRAME) {
+      LOGGER.debug("Streaming LZ4 Frame decompression. Context: {}", context);
+      return new LZ4FrameInputStream(new ByteArrayInputStream(compressedInput));
+    }
+    // null / NONE codec: no decompression needed.
+    return new ByteArrayInputStream(compressedInput);
   }
 
   public static InputStream decompress(
--- a/src/main/java/com/databricks/jdbc/api/impl/arrow/ArrowResultChunk.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/arrow/ArrowResultChunk.java
@@ -101,15 +101,16 @@ public class ArrowResultChunk extends AbstractArrowResultChunk {
           chunkLink.getExternalLink(),
           speedThreshold);
 
-      // Decompress (if needed) and parse
+      // Decompress and parse. The decompression is streamed straight into the Arrow reader so the
+      // full decompressed payload is never materialized on-heap alongside the compressed bytes.
       long decompressStart = System.nanoTime();
       try {
         String ctx =
             String.format(
                 "Data decompression for chunk index [%d] and statement [%s]",
                 this.chunkIndex, this.statementId);
-        InputStream data = DecompressionUtil.decompressToStream(compressed, compressionCodec, ctx);
-        initializeData(data);
+        initializeData(
+            DecompressionUtil.decompressToInputStream(compressed, compressionCodec, ctx));
       } catch (Exception e) {
         handleFailure(e, ChunkStatus.PROCESSING_FAILED);
       }
--- a/src/test/java/com/databricks/jdbc/common/util/DecompressionUtilTest.java
+++ b/src/test/java/com/databricks/jdbc/common/util/DecompressionUtilTest.java
@@ -40,6 +40,49 @@ public class DecompressionUtilTest {
   }
 
   @Test
+  public void testDecompressToInputStreamLZ4Frame() throws Exception {
+    byte[] uncompressed = INITIAL_STRING.getBytes();
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    try (LZ4FrameOutputStream lz4 = new LZ4FrameOutputStream(out)) {
+      lz4.write(uncompressed);
+    }
+    InputStream resultStream =
+        DecompressionUtil.decompressToInputStream(
+            out.toByteArray(), CompressionCodec.LZ4_FRAME, CONTEXT);
+    assertTrue(
+        IOUtils.contentEquals(resultStream, new ByteArrayInputStream(uncompressed)),
+        "Streaming decompression should yield the original bytes");
+  }
+
+  @Test
+  public void testDecompressToInputStreamNoneReturnsRawBytes() throws Exception {
+    byte[] raw = INITIAL_STRING.getBytes();
+    InputStream resultStream =
+        DecompressionUtil.decompressToInputStream(raw, CompressionCodec.NONE, CONTEXT);
+    assertTrue(IOUtils.contentEquals(resultStream, new ByteArrayInputStream(raw)));
+  }
+
+  @Test
+  public void testDecompressToInputStreamNullCodecReturnsRawBytes() throws Exception {
+    // A null codec is treated as no compression: the raw bytes are returned unchanged.
+    byte[] raw = INITIAL_STRING.getBytes();
+    InputStream resultStream = DecompressionUtil.decompressToInputStream(raw, null, CONTEXT);
+    assertTrue(IOUtils.contentEquals(resultStream, new ByteArrayInputStream(raw)));
+  }
+
+  @Test
+  public void testDecompressToInputStreamMalformedLz4FailsOnRead() throws Exception {
+    // LZ4FrameInputStream validates the frame lazily as it is read, so decompressToInputStream
+    // returns the streaming decorator successfully and the IOException surfaces on read. The
+    // caller (ArrowResultChunk.downloadData) catches it and routes it through handleFailure.
+    byte[] notLz4 = "this-is-not-a-valid-lz4-frame-payload".getBytes();
+    InputStream resultStream =
+        DecompressionUtil.decompressToInputStream(notLz4, CompressionCodec.LZ4_FRAME, CONTEXT);
+    assertNotNull(resultStream, "A streaming decorator should be returned before any read");
+    assertThrows(IOException.class, () -> IOUtils.toByteArray(resultStream));
+  }
+
+  @Test
   public void testDecompressLZ4FrameSkipsCompression() throws Exception {
     assertEquals(
         decompressionUtil.decompress(compressedInputStream, CompressionCodec.NONE, CONTEXT),


src/main/java/com/databricks/jdbc/api/impl/arrow/incubator/RemoteChunkProviderV2.java
```diff
@@ -111,11 +111,13 @@ protected ArrowResultChunkV2 createChunk(
    */
   @Override
   public void downloadNextChunks() throws DatabricksSQLException {
-    while (!isClosed
-        && nextChunkToDownload < chunkCount
-        && totalChunksInMemory < allowedChunksInMemory) {
+    while (!isClosed && nextChunkToDownload < chunkCount) {
       ArrowResultChunkV2 chunk = chunkIndexToChunksMap.get(nextChunkToDownload);
+      if (!canScheduleChunkDownload(chunk.getChunkSizeInBytes())) {
```
gopalldb: will it get attempted again?

src/main/java/com/databricks/jdbc/common/util/DecompressionUtil.java
```diff
@@ -60,6 +60,35 @@ public static InputStream decompressToStream(
     return new ByteArrayInputStream(uncompressed);
   }
 
+  /**
+   * Returns a stream that decompresses {@code compressedInput} lazily as it is read, so the full
+   * decompressed payload is never materialized alongside the compressed bytes.
+   */
+  public static InputStream decompressToInputStream(
```
gopalldb, line 65: Had added similar handling: https://github.com/databricks/databricks-jdbc/pull/1470/changes#diff-b1b86f2824a50564183d80c1022f233ecb2c960037b6a51f59dbac4f72fe7f91

gopalldb:
⚠️ [MAJOR] Byte budget silently inert when chunk sizes are unknown (0), disabling OOM protection
Line 331 | bug | logical_claude_agent | [1/3 Claude flagged]

In-memory byte budget is silently disabled whenever chunk sizes are unknown (0). chunkSizeInBytes defaults to 0 when the SEA manifest reports a null getByteCount() (ArrowResultChunk.java:222-223) or when the Thrift getBytesNum() is unset (line 256). With 0-sized chunks, totalBytesInMemory never grows, so canScheduleChunkDownload (AbstractRemoteChunkProvider.java:331-334) only ever enforces the count limit and the byte budget this PR adds is inactive. On exactly the large-result workloads where byte sizes may be missing, the heap bound silently does not apply, reintroducing the OOM the PR targets.

gopalldb:
Mostly LG, some minor comments

src/main/java/com/databricks/jdbc/api/impl/arrow/incubator/RemoteChunkProviderV2.java
msrathore-db: Yes. When `canScheduleChunkDownload(...)` returns false we `break` without advancing `nextChunkToDownload`, so the chunk is deferred, not dropped. `downloadNextChunks()` is re-invoked from `releaseChunk()` every time a consumed chunk frees budget, and it retries the same `nextChunkToDownload`. The always-allow-one rule (`totalChunksInMemory == 0`) guarantees forward progress even if a single chunk is larger than the whole budget. I've added a comment at the `break` documenting this.

src/main/java/com/databricks/jdbc/common/util/DecompressionUtil.java
msrathore-db: Good catch — #1470's `decompressLazy` and this `decompressToInputStream` are the same idea: wrap the compressed bytes in `LZ4FrameInputStream` and let the Arrow reader pull decompressed bytes lazily instead of materializing the full payload. Since #1470 is still open (and stacked on #1468), I didn't want to introduce a merge dependency here. Proposal: whichever lands first, the other consolidates onto the single shared helper — they're behaviourally identical so the merge is trivial. Happy to be the one to converge.

(Side note from testing: `LZ4FrameInputStream` already reads concatenated LZ4 frames on demand, so this also covers the >2 GB-decompressed case without a custom multi-frame reader.)

msrathore-db:
Fixed. You're right: with `chunkSizeInBytes == 0` (SEA `getByteCount()` null / Thrift `getBytesNum()` unset) `totalBytesInMemory` never grew, so `canScheduleChunkDownload` enforced only the count limit and the byte budget was inert.

I added `effectiveChunkSizeInBytes()` in `AbstractRemoteChunkProvider`, which charges a size-less chunk a conservative fallback estimate (`UNKNOWN_CHUNK_SIZE_ESTIMATE_BYTES`), applied symmetrically on reserve (both `RemoteChunkProvider` and `RemoteChunkProviderV2`) and release, so the budget stays active. Added `testUnknownChunkSizeStillConsumesBudget`, which asserts a size-less chunk still consumes budget (previously inert).

One honest caveat worth flagging: the budget charges the manifest's (compressed) `byteCount`, which under-counts the real per-chunk on-heap footprint (compressed buffer + Arrow parse transients). On narrow results this fix + streaming decompression comfortably resolves #1508 (verified: streams the repro table at `-Xmx48m`, where unpatched 3.x and even 2.7.5/2.8.1 OOM). On very wide/many-chunk results the *default* budget can still be too loose and rely on `cloudFetchMaxBytesInMemory` being tuned down. I think tightening the default (or charging a decompressed-size estimate) is worth a follow-up rather than expanding this PR — open to your preference.

gopalldb (approved):

gopalldb:
📄 src/main/java/com/databricks/jdbc/common/util/DecompressionUtil.java

💡 [MINOR] Possibly unused decompressToStream after switch to decompressToInputStream
Line 60 | style | deadcode_claude_agent

Potentially dead method — This PR replaces the ArrowResultChunk call site from DecompressionUtil.decompressToStream(...) (removed line 111 in ArrowResultChunk.java) to the new DecompressionUtil.decompressToInputStream(...). If decompressToStream had no other callers, it is now unused.

---
📄 src/main/java/com/databricks/jdbc/api/impl/arrow/StreamingChunkProvider.java

⚠️ [MAJOR] Byte budget not applied to StreamingChunkProvider, leaving its OOM path open
Line 48 | bug | logical_claude_agent

PR's byte-budget bound is missing from StreamingChunkProvider. The PR description states the in-memory byte budget applies to both RemoteChunkProvider and StreamingChunkProvider, but StreamingChunkProvider (StreamingChunkProvider.java:48) implements ChunkProvider directly and does not extend AbstractRemoteChunkProvider, so it inherits none of the new gating (maxBytesInMemory, canScheduleChunkDownload, totalBytesInMemory). Its triggerDownloads() (lines 504-530) schedules downloads bounded only by the chunk count maxChunksInMemory (line 509). As a result, when this provider is selected (constructed in ArrowStreamResult.java:121/202), up to maxChunksInMemory large chunks can be downloaded and decompressed concurrently with no byte cap — the exact heap-exhaustion (OOM) scenario the PR set out to fix remains open for this path, and the PR description overstates coverage.

---
📄 src/main/java/com/databricks/jdbc/api/impl/arrow/incubator/RemoteChunkProviderV2.java

💡 [MINOR] In-memory budget charged before throwing link fetch in RemoteChunkProviderV2.downloadNextChunks leaks on failure
Line 123 | bug | logical_claude_agent

Byte/count budget charged before a throwing link fetch, leaked on error. In RemoteChunkProviderV2.downloadNextChunks(), totalChunksInMemory++ and the PR-added totalBytesInMemory += chunkSizeInBytes (lines 123-124) run before the linkDownloadService.getLinkForChunk(...).get() block (lines 125-142). That block re-throws DatabricksSQLException on ExecutionException/InterruptedException, exiting downloadNextChunks() before nextChunkToDownload++ (line 144). The reserved count and bytes are never reconciled on this path — a link-fetch failure leaves the budget over-charged for a chunk that was never scheduled. Unlike RemoteChunkProvider (lines 128-132), which increments only after a successful submit(), V2 reserves budget before the fallible operation. If scheduling continues after such a failure, the inflated totals can throttle or block further downloads.

msrathore-db:
Heads-up on a scope change (force-pushed): I've **removed the in-memory byte-budget** (`cloudFetchMaxBytesInMemory`) and reduced this PR to **streaming decompression only**.

Why: real-warehouse testing showed the byte budget was redundant with the existing `cloudFetchThreadPoolSize` limit, and its default couldn't reliably bind — the manifest's `byteCount` is the *compressed* size, whereas the OOM is driven by the *decompressed + Arrow-parse* transients (~10–20×), so a compressed-bytes budget large enough to not throttle normal queries never engaged on the wide-result case anyway. Streaming decompression is the robust fix; for constrained heaps, `cloudFetchThreadPoolSize` (already a connection property) is the memory/throughput lever.

This makes the two earlier review comments obsolete:
- the `decompressToInputStream` vs #1470 `decompressLazy` duplication — still one shared helper worth consolidating whenever the two land, but no longer entangled with budget code; and
- the size-0 byte-budget concern — moot, as there is no byte budget.

Verified byte-identical results (SEA/Thrift, LZ4/NONE, complex types, multiple heaps) vs unpatched and the 2.7.5/2.8.1 Simba drivers.

PR #1512, opened 2026-06-29
[ES-1915324] Map untyped NULL columns to Types.NULL instead of Types.VARCHAR 

vikrantpuppala:
## Summary

An untyped `NULL` literal column — e.g. `SELECT NULL AS col`, which the server reports with type name `VOID`/`NULL` — was mapped to `java.sql.Types.VARCHAR` by `DatabricksTypeUtil.getColumnType(ColumnInfoTypeName.NULL)`.

In **Spark V2 JDBC federation** (Databricks-to-Databricks, UC foreign catalog over the JDBC connector) this caused a hard failure: Spark's `JdbcUtils` maps `VARCHAR` → `StringType`, but Catalyst still expects `NullType` for the original `Literal(null, NullType)`. `Cast.canCast(StringType, NullType)` is `false`, so the V2 pushdown assertion fires and the query wedges with `[INTERNAL_ERROR] / XX000`. Reported as **ES-1915324** for:

```sql
SELECT DISTINCT ..., NULL AS col FROM <foreign_view>
```

This PR maps `ColumnInfoTypeName.NULL` → `Types.NULL`, so Spark instead hits its "unsupported type" branch and **falls back gracefully** (non-pushed DISTINCT). This aligns with the JDBC spec and other drivers — MySQL Connector/J and H2 map untyped NULL to `Types.NULL`; Postgres uses `Types.OTHER` for the same effect.

The change is a single switch branch. The value-reading path is unaffected — a `NULL` column only ever carries `null`, and `ConverterHelper.convertSqlTypeToSpecificJavaType` short-circuits on `null` before any type-specific conversion (and its `default` branch already treats unknown SQL types as String).

## Test plan

- Added `DatabricksTypeUtilTest#testGetColumnTypeForNullIsSqlNull` — a regression test asserting `getColumnType(ColumnInfoTypeName.NULL) == Types.NULL` and that the `VOID`/`NULL` type names resolve to `ColumnInfoTypeName.NULL`. **Fails before the fix** (`expected <0> but was <12>`), passes after.
- Updated the existing `testGetColumnType` mapping assertion (`NULL` → `Types.NULL`).
- `mvn test -pl jdbc-core` for the type/converter/metadata/resultset suites — **984 tests pass**, no regressions. The separate Thrift `NULL_TYPE → STRING` mapping is intentionally left untouched (federation uses the type-name path).

Refs: ES-1915324

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -19,6 +19,7 @@
 - Fixed `getColumns()` flooding the `DriverManager` log writer with caught-and-recovered `Invalid column index` stack traces.
 - Fixed timezone-shifted TIMESTAMP values when retrieving nested complex types (STRUCT/ARRAY/MAP) with `EnableComplexDatatypeSupport=1`.
 - Fixed `DatabricksDatabaseMetaData.supportsBatchUpdates()` always returning `false`, which caused batch-aware JDBC clients (e.g. Apache Hop) to skip `executeBatch()` and fall back to one INSERT per row. It now returns `true` when `EnableBatchedInserts=1`, so those clients use the optimized multi-row INSERT path.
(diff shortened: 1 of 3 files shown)


vikrantpuppala:
Closing per discussion on ES-1915324: we're keeping the existing `Types.VARCHAR` mapping for untyped NULL columns. It's long-standing behavior and changing the reported JDBC type is a user-visible change; the federation wedge is being handled Spark-side (the in-flight workaround / `CAST(NULL AS STRING)`). Can revisit if there's a stronger reason to change the mapping.

PR #1182, opened 2026-01-19
Normalize TIMESTAMP_NTZ to TIMESTAMP in Thrift path for metadata consistency

gopalldb:
## Summary
This PR adds TIMESTAMP_NTZ normalization in the Thrift path to ensure consistent metadata behavior across both SEA and Thrift API paths.

## Background
PR #1177 moved Arrow metadata extraction earlier in the processing pipeline, which exposed an inconsistency: the Thrift path started returning the correct "TIMESTAMP_NTZ" from server metadata, while the SEA path was already normalizing it to "TIMESTAMP" for backward compatibility.

## Changes
- Added TIMESTAMP_NTZ → TIMESTAMP normalization in `DatabricksResultSetMetaData.java` Thrift constructor (lines 205-208)
- This brings Thrift path behavior in line with existing SEA path normalization
- Fixes test failure in `PreparedStatementIntegrationTests.testGetMetaData_NoResultSet`

## Testing
- ✅ Local test run: `PreparedStatementIntegrationTests.testGetMetaData_NoResultSet` passes
- ✅ Metadata now consistent before and after `executeQuery()` for TIMESTAMP_NTZ columns
- ✅ Both SEA and Thrift paths return "TIMESTAMP" for TIMESTAMP_NTZ columns

## Related
- Builds on PR #1177 (Fix Arrow field metadata not available for queries with 0 rows)
- Fixes issue introduced by early metadata extraction in PR #1177
- Maintains backward compatibility with existing behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1518, opened 2026-06-30
Preserve TIMESTAMP_NTZ type name in ResultSetMetaData (#1495)

### LINKED ISSUE #1495: [BUG] MIN or MAX on a 'TIMESTAMP_NTZ' column returns a TIMESTAMP type instead of TIMESTAMP_NTZ
**Describe the bug**
The return of MIN and MAX (at least, did not test all functions) of a 'TIMESTAMP_NTZ' column returns a TIMESTAMP type instead of TIMESTAMP_NTZ

**To Reproduce**
Create a table like
CREATE TABLE  bug_min (
	`TIMESTAMP_NTZ` timestamp_ntz
)
and run 
SELECT 
    MIN(`TIMESTAMP_NTZ`) AS `MIN_TIMESTAMP_NTZ_0`,
    MAX(`TIMESTAMP_NTZ`) AS `MAX_TIMESTAMP_NTZ_0`
  FROM bug_min 

Check the type of returned column, they are of type TIMESTAMP

**Expected behavior**
Both column returned should be of type TIMESTAMP_NTZ, but they are of type TIMESTAMP

**Screenshots**

**Client side logs**

 - OS: [Mac
 - Java version 17
 - Java vendor OpenJDK
 - Driver Version 3.4.1
 - BI Tool 
 - BI Tool version 

**Additional context**

This was working with version 3.0.7

msrathore-db:
## Description

Fixes [#1495](https://github.com/databricks/databricks-jdbc/issues/1495): `ResultSetMetaData.getColumnTypeName()` returns `TIMESTAMP` instead of `TIMESTAMP_NTZ` for `TIMESTAMP_NTZ` columns (e.g. `SELECT MIN(ntz_col) ...`). This was a regression from 3.0.7.

**Root cause.** The driver rewrites the `TIMESTAMP_NTZ` type text to `TIMESTAMP` in three metadata-construction paths in `DatabricksResultSetMetaData`:
- The **SEA** path has done this since the OSS launch (`columnInfo.setTypeText(TIMESTAMP)`), to "maintain parity with thrift output".
- The **Thrift** path got the same normalization in #1182 ("for consistency with SEA path"), first released in v3.2.1.
- The **describe-query** path also normalized.

Because the `ColumnInfoTypeName` SDK enum has no `TIMESTAMP_NTZ` value, `getTypeName()` is `null` for NTZ; the code correctly maps it to the `TIMESTAMP` enum (so `getColumnType()` → `Types.TIMESTAMP`), but it also clobbered the human-readable `typeText` that `getColumnTypeName()` returns. Combined with the effective default protocol moving to SEA, every default connection lost the NTZ type name. In 3.0.7 the default was Thrift, which did not normalize, so it returned `TIMESTAMP_NTZ`.

**Fix.** The driver now **preserves `TIMESTAMP_NTZ`** across all three paths by default. `getColumnType()` still reports `java.sql.Types.TIMESTAMP`, since `TIMESTAMP_NTZ` is a timestamp without timezone.

The legacy (Simba) driver reports `TIMESTAMP` for NTZ columns, so the previous behavior is preserved behind a new connection property:

| `EnableTimestampNtzTypeName` | `getColumnTypeName()` for NTZ | `getColumnType()` |
|---|---|---|
| `1` (default) | `TIMESTAMP_NTZ` | `Types.TIMESTAMP` |
| `0` (legacy / Simba parity) | `TIMESTAMP` | `Types.TIMESTAMP` |

## Testing

Verified live against a SQL warehouse with `SELECT MIN(ntz_col), MAX(ntz_col), ntz_col ... GROUP BY ntz_col`:

| Path | Default (`=1`) | `EnableTimestampNtzTypeName=0` |
|---|---|---|
| SEA (default) | `TIMESTAMP_NTZ` | `TIMESTAMP` |
| Thrift (`usethriftclient=1`) | `TIMESTAMP_NTZ` | `TIMESTAMP` |

For reference, the legacy Simba 2.8.1 driver returns `TIMESTAMP` for the same query — which the `=0` mode matches.

Unit/integration tests (`mvn -pl jdbc-core test`):
- `DatabricksResultSetMetaDataTest` — updated NTZ assertions for the new default; added `testColumnsWithTimestampNTZ_legacyTypeNameDisabled` covering `=0`.
- `PreparedStatementIntegrationTests#testGetMetaData_NoResultSet` (the fakeservice test #1182 was tied to), `DatabricksConnectionContextTest`, `DatabricksTypeUtilTest` all pass. Spotless clean.

This intentionally re-scopes the normalization introduced in #1182 behind the new `EnableTimestampNtzTypeName` flag rather than removing it — the legacy behavior is still reachable via `EnableTimestampNtzTypeName=0`. The metadata stays internally consistent (pre-execute vs post-execute) in either mode, which is what `testGetMetaData_NoResultSet` guards. cc @gopalldb (author of #1182).


src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
```diff
@@ -90,12 +90,18 @@ public DatabricksResultSetMetaData(
...
-          // For TIMESTAMP_NTZ columns, getTypeName() returns null.
-          // use typeText (initially "TIMESTAMP_NTZ") to identify the type,
-          // overwrite it to "TIMESTAMP" to maintain parity with thrift output.
+          // For TIMESTAMP_NTZ columns, getTypeName() returns null because the SDK
+          // ColumnInfoTypeName enum has no TIMESTAMP_NTZ value. Use typeText to
+          // identify the type and map it to the TIMESTAMP enum so the java.sql type
+          // resolves to Types.TIMESTAMP. By default the "TIMESTAMP_NTZ" typeText is
+          // preserved so getColumnTypeName() reports the actual server type
+          // (see GitHub issue #1495); when EnableTimestampNtzTypeName=0 it is
+          // normalized to "TIMESTAMP" to match the legacy (Simba) driver.
```
gopalldb: remove vendor name, can say (v2.x.x)

gopalldb (approved):

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
msrathore-db: Done — removed the vendor name; the comments, the `EnableTimestampNtzTypeName` description, and the changelog now refer to the legacy driver as "(v2.x.x)". Pushed in 5ef3e206.

msrathore-db: Done

msrathore-db:
Superseded by #1519, which raises the same change from a branch on the upstream repo (databricks/databricks-jdbc) instead of a fork. This avoids the fork-PR Security Scan SBOM failure. Closing in favor of #1519.

PR #1531, opened 2026-07-03
Fix telemetry misattribution across connections on the same thread

sreekanth-db:
## Summary

Telemetry export previously read the connection context from a static `ThreadLocal` (`DatabricksThreadContextHolder`) **at export time**. When multiple connections are used on the same thread (e.g. a Thrift connection and a SEA connection), the `ThreadLocal` holds whichever connection was created last, so per-statement telemetry events were tagged with the **wrong** connection's context — including the transport mode (`driverMode`), connection UUID, and other connection parameters.

This caused a measurable data-quality issue: ~22% of distinct SEA statement IDs in JDBC telemetry had no corresponding server-side record, because Thrift statements executed on the same thread were mislabeled as SEA (ES-1961329).

## Root cause

`TelemetryHelper.exportTelemetryLog(...)` resolved the connection via `DatabricksThreadContextHolder.getConnectionContext()`. That `ThreadLocal` is set once per `DatabricksConnection` construction and never refreshed per statement, so the last-created connection on a thread "wins" and all subsequent telemetry on that thread inherits its context.

## Fix

Each `TelemetryCollector` is already per-connection (keyed by `connectionUuid` in `TelemetryCollectorManager`). It now:

- Stores its own `IDatabricksConnectionContext` at construction time.
- Passes that context explicitly to `TelemetryHelper.exportTelemetryLog(...)` instead of relying on the shared `ThreadLocal`.

A new `exportTelemetryLog(connectionContext, details, logLevel)` overload takes the context explicitly; the old 2-arg overload is retained (deprecated) for backward compatibility.

Real connections always have a unique `connectionUuid`, so each gets its own collector holding its own context. Only context-less calls share the `DEFAULT_CONNECTION` collector; those events carry a null context and are skipped at export time, so no cross-connection misattribution can occur (documented inline).

## Files changed

- `TelemetryCollector.java` — stores `connectionContext`; passes it to all three `exportTelemetryLog` call sites; adds `getConnectionContext()` accessor.
- `TelemetryCollectorManager.java` — forwards the context into the collector constructor; documents the `DEFAULT_CONNECTION` trade-off.
- `TelemetryHelper.java` — new explicit-context `exportTelemetryLog` overload; old overload deprecated.
- Tests + `NEXT_CHANGELOG.md`.

### Automated unit tests (added)
- `TelemetryCollectorTest`: collector stores its own context (`testCollectorStoresConnectionContext`); `recordOperationLatency` calls the 3-arg `exportTelemetryLog`.
- `TelemetryCollectorManagerTest`: each collector stores the correct per-connection context (`testCollectorStoresCorrectConnectionContext`); a collector's context is not affected by subsequently creating another connection's collector (`testCollectorContextNotAffectedBySubsequentConnectionCreation`).
- `TelemetryHelperTest`: `exportTelemetryLog` uses the provided context and not the ThreadLocal (`testExportTelemetryLogWithExplicitContext_UsesProvidedContext`); null context short-circuits without export (`testExportTelemetryLogWithExplicitContext_NullContextSkipsExport`).

### Test runs
- Telemetry suites — `TelemetryCollectorTest`, `TelemetryCollectorManagerTest`, `TelemetryHelperTest`: **128 tests, 0 failures**.
- Broader related suites — connection / statement / session / http / retry / sdk / thrift / streaming / pooled connection: **519 tests, 0 failures**.

### Manual end-to-end reproduction (SEA + Thrift on one thread)
Created a SEA connection and a Thrift connection on the same thread, executed a query on the SEA connection, and dumped the actual `DriverConnectionParameters` telemetry payload.

Before the fix — payload for the SEA statement was built from the stale ThreadLocal (which held the Thrift connection):
```
stmt1 actual mode  = SEA
ThreadLocal mode   = THRIFT
=== PAYLOAD for stmt1 (sent to backend) ===
{ "mode" : "THRIFT", ... }   <-- WRONG
```

After the fix — the collector uses its own stored context:
```
=== FIX VERIFICATION ===
Collector for conn1: mode=SEA
Collector for conn2: mode=THRIFT
FIX WORKS: true
```

Fixes ES-1961329.

### DIFF
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
@@ -84,21 +84,6 @@ public class TelemetryHelper {
     exportTelemetryEvent(connectionContext, telemetryDetails, null, null, logLevel);
   }
 
-  /**
-   * @deprecated Use {@link #exportTelemetryLog(IDatabricksConnectionContext,
-   *     StatementTelemetryDetails, TelemetryLogLevel)} instead to avoid stale ThreadLocal context.
-   */
-  @Deprecated
-  public static void exportTelemetryLog(
-      StatementTelemetryDetails telemetryDetails, TelemetryLogLevel logLevel) {
-    exportTelemetryEvent(
-        DatabricksThreadContextHolder.getConnectionContext(),
-        telemetryDetails,
-        null,
-        null,
-        logLevel);
-  }
-
   private static void exportTelemetryEvent(
       IDatabricksConnectionContext connectionContext,
       StatementTelemetryDetails telemetryDetails,
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
@@ -104,6 +104,9 @@ public class TelemetryHelper {
         new TelemetryEvent()
             .setDriverSystemConfiguration(DRIVER_SYSTEM_CONFIGURATION)
             .setDriverConnectionParameters(getDriverConnectionParameter(connectionContext))
+            // TODO(ES-1961329 follow-up): sessionId is still read from the shared thread-local and
+            // can be misattributed across connections on the same thread (unlike connectionContext,
+            // which is now passed explicitly). Thread sessionId through per-connection state too.
             .setSessionId(DatabricksThreadContextHolder.getSessionId())
             .setDriverErrorInfo(errorInfo) // This is only set for failure logs
             .setSqlStatementId(telemetryDetails.getStatementId())
--- a/src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
+++ b/src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
@@ -78,6 +78,18 @@ public class TelemetryHelper {
   }
 
   public static void exportTelemetryLog(
+      IDatabricksConnectionContext connectionContext,
+      StatementTelemetryDetails telemetryDetails,
+      TelemetryLogLevel logLevel) {
+    exportTelemetryEvent(connectionContext, telemetryDetails, null, null, logLevel);
+  }
+
+  /**
+   * @deprecated Use {@link #exportTelemetryLog(IDatabricksConnectionContext,
+   *     StatementTelemetryDetails, TelemetryLogLevel)} instead to avoid stale ThreadLocal context.
+   */
+  @Deprecated
+  public static void exportTelemetryLog(
       StatementTelemetryDetails telemetryDetails, TelemetryLogLevel logLevel) {
     exportTelemetryEvent(
         DatabricksThreadContextHolder.getConnectionContext(),
--- a/src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorTest.java
+++ b/src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorTest.java
@@ -84,7 +84,7 @@ public class TelemetryCollectorTest {
       mockedStatic.verify(
           () ->
               TelemetryHelper.exportTelemetryLog(
-                  any(IDatabricksConnectionContext.class),
+                  eq(mockContext),
                   any(StatementTelemetryDetails.class),
                   any(TelemetryLogLevel.class)));
     }
--- a/src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorTest.java
+++ b/src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorTest.java
@@ -3,6 +3,7 @@ package com.databricks.jdbc.telemetry.latency;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.Mockito.*;
 
+import com.databricks.jdbc.api.internal.IDatabricksConnectionContext;
 import com.databricks.jdbc.common.TelemetryLogLevel;
 import com.databricks.jdbc.common.util.DatabricksThreadContextHolder;
 import com.databricks.jdbc.model.telemetry.StatementTelemetryDetails;
@@ -18,12 +19,14 @@ import org.junit.jupiter.params.provider.EnumSource;
 import org.mockito.MockedStatic;
 
 public class TelemetryCollectorTest {
-  private final TelemetryCollector handler = new TelemetryCollector();
   private static final String TEST_STATEMENT_ID = "test-statement-id";
+  private final IDatabricksConnectionContext mockContext = mock(IDatabricksConnectionContext.class);
+  private final TelemetryCollector handler = new TelemetryCollector(mockContext);
 
   @BeforeEach
   void setUp() {
     DatabricksThreadContextHolder.setStatementId(TEST_STATEMENT_ID);
+    when(mockContext.getTelemetryLogLevel()).thenReturn(TelemetryLogLevel.OFF);
   }
 
   @AfterEach
@@ -81,8 +84,15 @@ public class TelemetryCollectorTest {
       mockedStatic.verify(
           () ->
               TelemetryHelper.exportTelemetryLog(
-                  any(StatementTelemetryDetails.class), any(TelemetryLogLevel.class)));
+                  any(IDatabricksConnectionContext.class),
+                  any(StatementTelemetryDetails.class),
+                  any(TelemetryLogLevel.class)));
     }
+  }
+
+  @Test
+  void testCollectorStoresConnectionContext() {
+    assertSame(mockContext, handler.getConnectionContext());
   }
 
   @ParameterizedTest
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,7 @@
 - Bumped the Databricks SDK for Java dependency from `0.106.0` to `0.118.0`.
 
 ### Fixed
+- Fixed telemetry misattribution when multiple connections (e.g. Thrift and SEA) are used on the same thread. Per-statement telemetry events could be tagged with another connection's context (e.g. transport mode); each connection's telemetry now uses its own context instead of a shared thread-local value.
 - Hardened the OAuth U2M token cache at rest (encryption key derivation and file permissions).
 - Fixed `DatabaseMetaData.getURL()` exposing credentials embedded in the connection URL; secret parameters are now masked (the URL is otherwise unchanged).
 - Fixed presigned URL credentials not being fully redacted in logs.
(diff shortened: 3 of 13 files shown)


src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorTest.java
```diff
@@ -81,10 +84,17 @@ void testRecordOperationLatency_WithCloseOperation() {
       mockedStatic.verify(
           () ->
               TelemetryHelper.exportTelemetryLog(
-                  any(StatementTelemetryDetails.class), any(TelemetryLogLevel.class)));
+                  any(IDatabricksConnectionContext.class),
```
vikrantpuppala: The one collector test exercising the export path verifies `exportTelemetryLog(any(IDatabricksConnectionContext.class), any(), any())`. But `mockContext` is available (used at line 95 with `assertSame`). With `any()`, the test passes for *any* context — including a regression where the collector reverts to reading the ThreadLocal. The central behavior of the fix (collector passes *its own* context) isn't asserted.

**Suggested fix:** change the matcher to `eq(mockContext)`.

_[MODERATE CONFIDENCE — test + architecture; verified against head SHA]_

---

src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
```diff
@@ -77,6 +77,18 @@ public static boolean isTelemetryAllowedForConnection(IDatabricksConnectionConte
...
+      TelemetryLogLevel logLevel) {
+    exportTelemetryEvent(connectionContext, telemetryDetails, null, null, logLevel);
+  }
+
+  /**
+   * @deprecated Use {@link #exportTelemetryLog(IDatabricksConnectionContext,
+   *     StatementTelemetryDetails, TelemetryLogLevel)} instead to avoid stale ThreadLocal context.
+   */
+  @Deprecated
   public static void exportTelemetryLog(
```
vikrantpuppala: All three production call sites migrated to the 3-arg form, so the deprecated 2-arg (ThreadLocal-based) overload has no production callers — only tests exercising the old path. `@Deprecated` is a soft signal; a future contributor could reach for the shorter overload and silently reintroduce this exact bug.

**Suggested fix:** consider removing it and rewriting the `TelemetryHelperTest` cases (which already build a context mock) to the 3-arg form; if it must stay one release for compatibility, note in the Javadoc that it has no internal callers.

_[MODERATE CONFIDENCE — maintainability + architecture; verified against head SHA]_

---

vikrantpuppala:

**Score**: 63/100 — HIGH RISK

This PR correctly fixes the *stamping* half of the telemetry misattribution bug (each `TelemetryCollector` now carries its own connection context), but the review found strong multi-reviewer-corroborated evidence the fix is incomplete: the highest-volume path (`DatabricksMetricsTimedProcessor.recordOperationLatency`) still *selects* the collector via the same stale ThreadLocal, so cross-connection misattribution persists there. Separately, a leak-regression test (in a file not touched by this PR) is silently disarmed by the overload migration. The change is safe as written, but does not fully deliver its stated goal.

---

Could not post inline comments for: F1, F2, F3, F5 — see body below.

## Critical & High Findings

### F1 (High) — `src/main/java/com/databricks/jdbc/telemetry/latency/DatabricksMetricsTimedProcessor.java`:L80

The PR fixes how a collector *stamps* context, but not *which* collector is chosen. This selection path still resolves via the shared ThreadLocal:

```java
TelemetryCollectorManager.getInstance()
    .getCollectorSafely(DatabricksThreadContextHolder::getConnectionContext);
```

`setConnectionContext` is only called in the `DatabricksConnection` constructor and async-task setup — not per statement. So with SEA conn A then Thrift conn B on the same thread, the ThreadLocal holds B; a timed method on A resolves **B's** collector, which now stamps B's (Thrift) context onto A's (SEA) statement — the same misattribution class as ES-1961329, for the highest-volume path (`recordOperationLatency` on every `@DatabricksMetricsTimed` RPC). `statementId` is also read from the ThreadLocal at `TelemetryCollector.java:76`.

- **Suggested fix:** resolve the collector from the timed target's own context — `IDatabricksClient` already exposes `getConnectionContext()` — so selection and stamping derive from the same per-connection source.

_[HIGH CONFIDENCE — devils-advocate + architecture; verified against head SHA]_

---

### F2 (High) — `src/test/java/com/databricks/jdbc/telemetry/TelemetryHttpClientLeakTest.java`:L116

This test `mockStatic`s `TelemetryHelper` and stubs the **2-arg** `exportTelemetryLog(any(), any())` with a `thenAnswer` that simulates the client-recreation leak (incrementing `reCreationCount`). This PR migrates all three `TelemetryCollector` export calls to the **3-arg** overload (`TelemetryCollector.java:84-85, 126-127, 176-177`). Because the class is fully mocked and the 3-arg overload is unstubbed, it resolves to Mockito's default no-op — the `thenAnswer` body **never runs**, so `assertEquals(0, reCreationCount.get(), ...)` passes vacuously.

This file is **not in the PR's changed set**, so the coverage evaporates silently. The test guards issue #1325 (CRaC socket leak via telemetry-client re-creation on close). A future regression would ship green.

- **Suggested fix:** update this stub and the `stubExportTelemetryLog` helper at line 236 to the 3-arg matcher `exportTelemetryLog(any(), any(), any())`, and confirm the test fails when the leak is reintroduced.

_[HIGH CONFIDENCE — test + language + maintainability; verified against head SHA]_

---

### F3 (High) — `src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java`:L122

The new overload threads `connectionContext` through, but the shared `exportTelemetryEvent` body still does `.setSessionId(DatabricksThreadContextHolder.getSessionId())`. On the close-time flush (`exportAllPendingTelemetryDetails`), connection A's events are tagged with A's context but carry whatever `sessionId` the flushing thread's ThreadLocal holds (possibly B's, or null after `clearAllContext`) — the same anti-pattern the PR eliminates for `connectionContext`, left in place for `sessionId`.

**Reviewer split:** rated Low by language (pre-existing, arguably out of scope) and High by devil's advocate (same bug class). Taking the conservative position. **Minimum action:** if out of scope, add a comment marking the remaining `sessionId` ThreadLocal dependency so it isn't mistaken as fully fixed.

_[MODERATE CONFIDENCE — devils-advocate + language; verified against head SHA]_

---

## Medium Findings

### F5 (Medium) — `src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorManagerTest.java`:L208

The new tests only assert the stored-field getter returns the constructor-supplied context. None wire two collectors + a thread-local pointing at the wrong one through `exportAllPendingTelemetryDetails`/`recordOperationLatency` — the layer where ES-1961329 manifested. A refactor re-deriving export context from the ThreadLocal inside the collector would pass all the new tests. (Compounds F1 — an end-to-end test would likely expose it.)

- **Suggested fix:** add a test creating collectors for context1/SEA and context2/THRIFT, set the thread-local to context2, export a close-op on collector1, and verify `exportTelemetryLog(eq(context1), …)` fired and `eq(context2)` did not.

_[LOW CONFIDENCE — test]_

---

src/test/java/com/databricks/jdbc/telemetry/latency/TelemetryCollectorTest.java
sreekanth-db: Good catch — done in 6b53dad3. Changed the matcher to `eq(mockContext)` so the test now asserts the collector passes its *own* stored context; a regression back to reading the ThreadLocal would fail this test.

src/main/java/com/databricks/jdbc/telemetry/TelemetryHelper.java
sreekanth-db: Agreed — done in 6b53dad3. Removed the deprecated 2-arg overload entirely rather than keeping it around as a footgun. Confirmed there were no production callers (all three sites already used the explicit-context form); migrated the remaining test callers in TelemetryHelperTest and TelemetryHttpClientLeakTest to the 3-arg form.

sreekanth-db:
Thanks for the thorough review — addressed in `123e52c2` (and `6b53dad3`):

- **F1 (collector selection via ThreadLocal)** — Fixed. `DatabricksMetricsTimedProcessor` now resolves the collector from the timed target's own connection context instead of the shared thread-local, so selection and stamping derive from the same per-connection source. This also covers the metadata path: `getConnectionContext()` was added to `IDatabricksMetadataClient` (implemented by `DatabricksMetadataQueryClient` and `DatabricksEmptyMetadataClient`; `DatabricksThriftServiceClient` already satisfies it via `IDatabricksClient`), since those clients implement only `IDatabricksMetadataClient` and every method is `@DatabricksMetricsTimed`.
- **F2 (leak test disarmed)** — Fixed. Both stubs in `TelemetryHttpClientLeakTest` (incl. `stubExportTelemetryLog`) were migrated to the 3-arg `exportTelemetryLog(any(), any(), any())`.
- **F5 (missing end-to-end coverage)** — Added regression tests in `DatabricksMetricsTimedProcessorTest` for both the statement-client and metadata-client paths: with the thread-local pointing at connection B, a timed method on connection A's client resolves A's context (not B's).
- **F3 (sessionId still from ThreadLocal)** — Marked with a TODO at the call site and **deferred**. sessionId, along with a few other telemetry-attribution items (error-log attribution, JVM-global user-agent/app-name, metadata statement-id) require broader, cross-cutting changes and will be taken up in follow-up PRs. This PR is intentionally scoped to the primary transport-mode misattribution (the ~22% SEA/Thrift mislabeling in ES-1961329).

vikrantpuppala:
**Re-review of `123e52c2` — F1/F2/F4/F5 confirmed resolved; F3 deferral acknowledged. One residual item for the same follow-up bucket.**

Thanks — the F1 fix (`resolveConnectionContext()` reading the target client's own context, plus threading `getConnectionContext()` through `IDatabricksMetadataClient`) is exactly the right shape, and the new `DatabricksMetricsTimedProcessorTest` cases reproduce the actual SEA-vs-Thrift-on-one-thread scenario rather than just asserting the stored field. The F3 deferral rationale (scoped to the primary transport-mode mislabeling; sessionId + other attribution items batched into follow-ups) is reasonable — no objection to deferring it here.

**Residual observation, same class as the deferred items:** `TelemetryCollector.recordOperationLatency` still reads the statement id from the shared thread-local:

```java
// TelemetryCollector.java:76
String statementId = DatabricksThreadContextHolder.getStatementId();
```

Post-fix the collector is now *selected* correctly (F1) and the `connectionContext` stamped on the event is correct, but the `statementId` used to key the per-statement tracker and label the operation-latency event is still thread-local-sourced. On a thread executing one statement at a time this is fine; but it's the same misattribution class as the `sessionId` `TODO(ES-1961329 follow-up)` you added — if a statement from another connection last populated the thread-local, the operation-latency event could carry the wrong statement id.

You already listed "metadata statement-id" among the deferred follow-up items, so this may well be covered — just flagging that this specific call site (`TelemetryCollector.java:76`) would benefit from a matching `TODO(ES-1961329 follow-up)` so it isn't mistaken as fully resolved when someone revisits this file. Non-blocking.

_[Code Review Squad — residual observation from re-review of the F1 fix]_

vikrantpuppala (approved):

PR #1241, opened 2026-03-02
Add EnableOAuthSecretFromPwd connection parameter

gopalldb:
## Summary

Resolves #1132 — OAuth secrets exposed in JDBC URL.

When using OAuth Client Credentials (M2M), BI tools like DBeaver treat the `password` field as sensitive (masked, not logged), but display the full JDBC URL in clear text. This exposes the `OAuth2Secret` parameter. This PR introduces an opt-in `EnableOAuthSecretFromPwd` connection parameter that lets the driver read the OAuth client secret from `PWD`/`password` instead.

### New Connection Parameter

| Parameter | Default | Description |
|-----------|---------|-------------|
| `EnableOAuthSecretFromPwd` | `0` | When `1`, read OAuth client secret from `PWD`/`password` instead of `OAuth2Secret` |

### Behavior

**When `EnableOAuthSecretFromPwd=1`:**
- `getClientSecret()` **always** reads from `PWD`/`password` (`pwd` takes priority over `password`, matching `getToken()` behavior)
- `OAuth2Secret` is **ignored** even if explicitly set — `PWD` always wins when the feature is enabled
- If neither `PWD` nor `password` is provided, throws `DatabricksDriverException` with a clear error message:
  > *"EnableOAuthSecretFromPwd is enabled but no PWD/password property was provided. Set the OAuth client secret via the PWD or password connection property."*
- Covers **all flows** that call `getClientSecret()`: M2M Standard, M2M Azure, Refresh Token, and Browser-Based (U2M)

**When `EnableOAuthSecretFromPwd=0` (default):**
- No change — `getClientSecret()` reads from `OAuth2Secret` as before. Fully backward compatible.

### Usage Example

```java
// M2M with secret in password field — BI tools will mask this
String url = "jdbc:databricks://host:443/default;AuthMech=11;Auth_Flow=1;"
    + "OAuth2ClientId=my-client-id;EnableOAuthSecretFromPwd=1;httpPath=/sql/...";
Connection conn = DriverManager.getConnection(url, "token", "my-oauth-secret");
```

### Files Changed

| File | Change |
|------|--------|
| `DatabricksJdbcUrlParams.java` | Add `ENABLE_OAUTH_SECRET_FROM_PWD` enum constant |
| `IDatabricksConnectionContext.java` | Add `isOAuthSecretFromPwdEnabled()` interface method |
[...]

PR #1533, opened 2026-07-06
Accept OAuth M2M client id/secret via JDBC user/password (#1132)

### LINKED ISSUE #1132: [FEATURE] Accept oauth secret as password
**Is your feature request related to a problem? Please describe.**
When using OAuth Client Credentials (Machine-to-Machine), the JDBC driver requires putting the oauth secret into the JDBC URL:

```
AuthMech=11;Auth_Flow=1;OAuth2ClientId=<client_id>;OAuth2Secret=<client_secret>
```

Depending on the application, the JDBC URL (and the secret) might be revealed e.g. in log files.

**Describe the solution you'd like**
Update the JDBC driver to accept the OAuth secret as password (and maybe the client ID as username). This way, the OAuth secret will not be revealed in log files.

**Describe alternatives or workarounds you've considered**
Using [`DriverManager.getConnection(String url, Properties info)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/DriverManager.html#getConnection(java.lang.String,java.util.Properties)) and specifying the secret in the `info` properties.

But this is not possible when the application only uses `DriverManager.getConnection(String url, String user, String password)`.

**Additional Context**
When using tokens, you can use `getConnection(url, "token", "<your_token>")`. Maybe a similar mechanism could be implemented also for M2M Oauth.

vikrantpuppala:
## Summary

Closes #1132.

When using OAuth Client Credentials (M2M), BI tools like DBeaver mask the `password` field (not logged, hidden in the UI) but display the full JDBC URL in clear text — which exposes the `OAuth2Secret` embedded in the URL.

This PR lets the driver read the OAuth **client secret** from the JDBC `password`/`PWD` property and the OAuth **client id** from the JDBC `user`/`UID` property, so users can pass them through `DriverManager.getConnection(url, user, password)` and let the BI tool mask the secret.

```java
String url = "jdbc:databricks://host:443/default;AuthMech=11;Auth_Flow=1;httpPath=/sql/...";
// client id via user, secret via password — masked by the BI tool
Connection conn = DriverManager.getConnection(url, "my-client-id", "my-oauth-secret");
```

### Behavior

- Applies only in OAuth mode (`AuthMech=11`), across all OAuth flows that consume the client secret (M2M standard/Azure, refresh token, browser).
- **Explicit params win:** if `OAuth2ClientId`/`OAuth2Secret` are set in the URL, they take precedence. The `user`/`password` values are used only as a fallback, so existing URLs behave identically.
- `user`/`UID` value of `"token"` (the PAT sentinel) is ignored as a client id.
- UID validation is relaxed in OAuth mode so a non-`"token"` UID may carry the client id. In PAT mode the `UID` must still be `"token"` or omitted (unchanged).
- The missing-properties advisor (`getPropertyInfo`) no longer reports `OAuth2ClientId`/`OAuth2Secret` as missing when `user`/`password` supply them.

### Addresses prior review feedback

This supersedes the stale PR #1241 and resolves both of its unaddressed review comments:
- *"why is this gated?"* — the behavior is **ungated** (no feature flag), instead of the opt-in `EnableOAuthSecretFromPwd` flag.
- *"why not also take the client ID as the user?"* — the client id is now read from the JDBC username.

## Changes

| File | Change |
|------|--------|
| `DatabricksConnectionContext.java` | `getClientSecret()`/`getNullableClientId()` fall back to `password`/`user` in OAuth mode |
| `DatabricksJdbcUrlParams.java` | Add `USER("user", ...)` param (the `user` arg from the 3-arg `getConnection` was previously unread) |
| `ValidationUtil.java` | Allow non-`token` UID in OAuth mode |
| `DatabricksDriverPropertyUtil.java` | Don't flag `OAuth2ClientId`/`OAuth2Secret` missing when `user`/`password` present |

## Test plan

- [x] `mvn test -Plocal` (full unit suite): **3425 tests, 0 failures**
- [x] 10 new unit tests in `DatabricksConnectionContextTest` (secret/id fallback, precedence, `token` sentinel, PAT-mode isolation)
- [x] 4 new unit tests in `ValidationUtilTest` (OAuth vs PAT UID validation)
- [x] 2 new unit tests in `DatabricksDriverPropertyUtilTest` (missing-property behavior)
- [x] `mvn spotless:apply`
- [ ] 2 new WireMock M2M integration tests in `M2MAuthIntegrationTests` (validate in CI — the OAuth token exchange requires the real M2M endpoint/creds available only in CI; recorded mappings copied from the existing `testSuccessfulM2MConnection` scenario)

Co-authored-by: Isaac


src/main/java/com/databricks/jdbc/api/internal/IDatabricksConnectionContext.java
```diff
@@ -479,4 +479,16 @@ default int getHeartbeatIntervalSeconds() {
...
+
+  /**
+   * Returns whether the OAuth client id or secret is being sourced from the JDBC {@code
+   * user}/{@code password} connection properties (the fallback for {@code OAuth2ClientId}/{@code
+   * OAuth2Secret}) rather than the explicit OAuth params. For telemetry only — no credential values
+   * are exposed.
+   *
+   * @return true if the user/password OAuth-credential fallback is in effect
+   */
+  default boolean usesOAuthCredentialsFromUserPassword() {
```
gopalldb: is this only for telemetry?

vikrantpuppala: yes it was removed it on second thought

gopalldb:
📄 src/main/java/com/databricks/jdbc/common/util/DatabricksDriverPropertyUtil.java

💡 [MINOR] Missing-property advisor treats UID/user='token' as a supplied OAuth client id
Line 233 | bug | logical_claude_agent

Advisor falsely reports OAuth2ClientId as supplied when UID/user is the PAT sentinel 'token'. addMissingClientId (line 233-236) returns early — suppressing the OAuth2ClientId missing-property hint — whenever isPropertyPresent(USER) or isPropertyPresent(UID) is true. But isPropertyPresent only checks that the key exists, not its value, whereas getNullableClientId() (line 347) explicitly ignores the value "token". So in OAuth CLIENT_CREDENTIALS (M2M) mode with uid=token or user=token (a value users commonly carry over from PAT connections), the advisor claims the client id is provided while getClientId() actually receives null from getNullableClientId() and silently substitutes a cloud-default client id. The getPropertyInfo guidance is misleading for exactly the M2M users this PR is meant to help, hiding a real misconfiguration.

---
📄 src/main/java/com/databricks/jdbc/dbclient/impl/common/ClientConfigurator.java

⚠️ [MAJOR] OAuth user/password fallback leaks into U2M browser flow, altering public-client PKCE behavior
Line 214 | bug | logical_claude_agent

JDBC user/password fallback bleeds into the U2M browser flow. getClientSecret() (DatabricksConnectionContext.java:381-392) and getNullableClientId() (340-352) fall back to the JDBC password/user for any OAuth mode (isOAuthMode(), AuthMech=11), not just the M2M client-credentials flow. setupU2MConfig() (BROWSER_BASED_AUTHENTICATION) reads both at ClientConfigurator.java:206 and :214. Consequently, an existing browser-auth user who opens a connection via getConnection(url, user, password) without explicit OAuth2ClientId/OAuth2Secret will now have those values injected as the OAuth client id and client secret. Previously the secret was empty and the flow ran as a public client using PKCE; the change silently turns it into a confidential-client request (and overrides the client id), a behavior change that can break or alter existing U2M connections.

---
📄 src/test/java/com/databricks/jdbc/integration/fakeservice/tests/M2MAuthIntegrationTests.java

⚠️ [MAJOR] testExplicitOAuthSecretWinsOverPassword does not assert the transmitted client_secret, so it cannot detect a precedence regression
Line 71 | bug | logical_claude_agent

Precedence test is non-discriminating — nothing asserts which secret was sent. testExplicitOAuthSecretWinsOverPassword (line 71-83) supplies an explicit OAuth2Secret plus a bogus password and asserts only that the connection opens. But the recorded WireMock fixtures contain no stub matching the OAuth token endpoint with bodyPatterns, and the sole OIDC stub matches on URL/method only with no constraint on client_secret. Under default REPLAY mode, the token exchange is never matched against the secret value, so the test passes irrespective of whether the explicit secret or the bogus password is actually transmitted. A regression that let the password override the explicit OAuth2Secret would still make this test green.

vikrantpuppala:
Thanks — addressed all three in `342fd457`:

**[MAJOR] U2M browser-flow leak** — Good catch. The fallback was gated on `isOAuthMode()` (all `AuthMech=11`), so it bled into the browser and token-passthrough flows. Now scoped to the M2M client-credentials flow only via `isM2MClientCredentialsMode()` (`AuthMech=11` **and** `Auth_Flow=1`). Added negative unit tests `testBrowserAuthDoesNotReadCredsFromUserPassword` (Auth_Flow=2) and `testRefreshTokenFlowDoesNotReadCredsFromUserPassword` (Auth_Flow=0) asserting `getClientSecret()`/`getNullableClientId()` stay null there.

**[MINOR] Advisor treats `uid/user='token'` as a supplied client id** — Fixed. `addMissingClientId`/`addMissingClientSecret` now decide based on the resolved `getNullableClientId()`/`getClientSecret()` values (which already ignore the `"token"` sentinel and are M2M-scoped) instead of raw `isPropertyPresent()`. So an M2M connection with `uid=token` is now correctly told `OAuth2ClientId` is missing.

**[MAJOR] Non-discriminating precedence integration test** — Correct, the WireMock stubs match on URL/method only, so it couldn't detect a precedence regression. Removed that integration test (and its copied mappings). Precedence is covered by the discriminating unit test `testExplicitOAuthSecretWinsOverPassword` in `DatabricksConnectionContextTest`, which asserts `getClientSecret()` returns the explicit secret over the password directly.

Full unit suite green (3426 tests).

gopalldb (approved):

PR #1546, opened 2026-07-08
Fix MAP with complex value type rendering as empty via Arrow (#1505)

### LINKED ISSUE #1505: [BUG]map<int,array<bigint>> column not selected properly in 3.3.1 with EnableArrow=1

vikrantpuppala:
## Summary

Fixes #1505. Selecting a `MAP` column whose values are themselves a complex type (e.g. `MAP<INT,ARRAY<BIGINT>>`) via Arrow (`EnableArrow=1`) rendered the values as empty:

- Query: `SELECT MAP(0, ARRAY(34277,0))`
- Expected: `{0:[34277,0]}`
- Actual (before fix): `{0:}`

The non-Arrow path was unaffected.

## Root Cause

When `EnableArrow=1` and `EnableComplexDatatypeSupport` is disabled (the default), `ArrowStreamResult.getObjectWithComplexTypeHandling` fetches the MAP as a raw JSON string (`[{"key":0,"value":[34277,0]}]`) and reformats it via `ComplexDataTypeParser.formatMapString`.

That method rendered each value with `JsonNode.asText()`, which Jackson returns as `""` for a container (array/object) node. Scalar keys/values rendered fine, but any nested ARRAY/STRUCT/MAP value collapsed to empty — producing `{0:}`.

## Fix

In `ComplexDataTypeParser.formatMapString`, detect container value nodes (`JsonNode.isContainerNode()`) and emit their JSON directly via `toString()` instead of `asText()`. This produces `{0:[34277,0]}`, consistent with how `DatabricksMap`/`DatabricksArray` render nested values.

## Test Plan

Added unit tests to `ComplexDataTypeParserTest` (all fail before the fix, pass after):
- `testFormatMapString_withArrayValue` — the exact issue scenario, `MAP<INT,ARRAY<BIGINT>>` → `{0:[34277,0]}`
- `testFormatMapString_withStructValue` — `MAP<INT,STRUCT<a:INT,b:INT>>`
- `testFormatMapString_withMapValue` — nested `MAP<INT,MAP<INT,INT>>`
- `testFormatMapString_withStringArrayValue` — `MAP<STRING,ARRAY<STRING>>` (value quoting preserved for the array elements, not the container)

Verified no regressions across `ComplexDataTypeParserTest`, `ArrowToJavaObjectConverterTest`, `ArrowStreamResultTest`, `DatabricksMapTest`, `DatabricksArrayTest`, `DatabricksStructTest` (185 tests pass).

Closes #1505

## E2E verification

Confirmed end-to-end against a live Databricks endpoint (dogfood), not just at the unit level. Added `ComplexTypeQueryTests#testMapWithArrayValue` running the reporter's exact query `SELECT MAP(0, ARRAY(34277,0))` across all four `{usethriftclient, EnableComplexDatatypeSupport}` combinations:

- With the fix **removed**, both `EnableComplexDatatypeSupport=0` cases (SEA/Arrow and Thrift) fail at runtime with `expected: <{0:[34277,0]}> but was: <{0:}>` — reproducing the reported bug exactly.
- With the fix in place, all four cases pass and the driver returns `{0:[34277,0]}`.

Note: `EnableArrow` is deprecated and Arrow is always enabled on non-AIX/PowerPC platforms, so the reporter's `EnableArrow=1` is the default path; the bug is gated purely on `EnableComplexDatatypeSupport` being disabled (the default).


sreekanth-db (approved):
LGTM

PR #1548, opened 2026-07-08
Fix NPE reading STRING COLLATE columns via Arrow

### LINKED ISSUE #1547: [BUG] NullPointerException reading STRING COLLATE UTF8_LCASE columns via Arrow (ColumnInfoTypeName.ordinal() - requiredType is null)
**Describe the bug**

Reading a `STRING` column that carries a non-default collation (e.g. `COLLATE UTF8_LCASE`) throws a `NullPointerException` inside the Arrow result converter. The driver cannot map the collated string type to a `ColumnInfoTypeName`, so `requiredType` is null and `ColumnInfoTypeName.ordinal()` is invoked on it.

The failure is on the row-read path (`ResultSet.getObject`), so any query returning a collated string column fails — including a single-row result. Non-collated columns in the same row read fine, and casting the column to plain `STRING` (dropping the collation) also reads fine.

This is a regression: `3.3.3` and `3.3.1` read the same column/query without error; `3.4.1` throws.

**To Reproduce**

1. Create a view/table exposing a `UTF8_LCASE`-collated string column:
```sql
CREATE OR REPLACE VIEW demo_collation AS
SELECT 'abc' COLLATE UTF8_LCASE AS name;
```
2. Query it over JDBC (driver 3.4.1, `AuthMech=3` PAT, SQL Warehouse, default Arrow result format) and read the column:
```java
try (Connection c = DriverManager.getConnection(url);
     Statement st = c.createStatement();
     ResultSet rs = st.executeQuery("SELECT name FROM demo_collation")) {
    while (rs.next()) {
        rs.getObject("name"); // NullPointerException
    }
}
```
3. Observe the NPE (stack trace below).

Note: reproduces when the SQL warehouse serializes the collation type tag over Arrow. We saw two warehouses on the same DBSQL version label (`2026.15`) but different build hashes behave differently — one triggered the NPE, one did not. `EnableQueryResultDownload=0` (JSON inline results, no Arrow) avoids the failure on `3.4.1`.

**Expected behavior**

`rs.getObject("name")` returns the string value, as it does for a non-collated `STRING` column and as `3.3.3`/`3.3.1` do for the collated column.

**Screenshots**

N/A

**Client side logs**

```
java.lang.NullPointerException: Cannot invoke
"com.databricks.jdbc.model.core.ColumnInfoTypeName.ordinal()" because "requiredType" is null
    at com.databricks.jdbc.api.impl.converters.ArrowToJavaObjectConverter.convert(ArrowToJavaObjectConverter.java:97)
    at com.databricks.jdbc.api.impl.arrow.ArrowResultChunkIterator.getColumnObjectAtCurrentRow(ArrowResultChunkIterator.java:84)
    at com.databricks.jdbc.api.impl.arrow.ArrowStreamResult.getObjectWithComplexTypeHandling(ArrowStreamResult.java:391)
    at com.databricks.jdbc.api.impl.arrow.ArrowStreamResult.getObject(ArrowStreamResult.java:242)
    at com.databricks.jdbc.api.impl.DatabricksResultSet.getObjectInternal(DatabricksResultSet.java:2298)
    at com.databricks.jdbc.api.impl.DatabricksResultSet.getObject(DatabricksResultSet.java:802)
```

 - OS: macOS (also reproduced on Linux containers)
 - Java version: Java 21
 - Java vendor: OpenJDK
 - Driver Version: 3.4.1 (`databricks-jdbc-thin`)
 - BI Tool (if used): N/A (direct JDBC)
 - BI Tool version (if applicable): N/A

**Additional context**

- The Arrow result manifest reports the collated columns with `type_name = UNKNOWN_COLUMN_TYPE_NAME` and `type_text = STRING COLLATE UTF8_LCASE` — likely where `requiredType` becomes null.
- Version matrix (same warehouse, same query, only the driver jar changed): `3.4.1` → NPE; `3.3.3` → OK; `3.3.1` → OK. Regression introduced between `3.3.3` and `3.4.1`.
- Client-side mitigations found: cast the column to plain `STRING`, or set `EnableQueryResultDownload=0`, or downgrade to `3.3.3`.

msprunck:
## Description

Fixes #1547.

A `STRING` column carrying a non-default collation (e.g. `STRING COLLATE UTF8_LCASE`) cannot be read when the warehouse reports its type via the Statement Execution API (SEA). Such a column arrives with a `type_name` that does not map to any `ColumnInfoTypeName`, so `ColumnInfo.getTypeName()` is `null`, and two things break:

1. **Value read** — `ArrowToJavaObjectConverter.convert` reaches `switch (requiredType)` with `requiredType == null` and throws a `NullPointerException`. This aborts every result set containing a collated string column, including a single-row result.
2. **Metadata** — `DatabricksResultSetMetaData` maps the null type name to `Types.OTHER`, so `getColumnType()` returns `OTHER` and the column is not usable as a string, even once the value read is fixed.

### Trigger

The defect is latent in the driver's type handling — the converter and metadata builder both rely on the `ColumnInfoTypeName` enum resolving, and neither tolerates a null. What surfaces it is which metadata path the warehouse answers with: the Thrift path maps the collated column to `STRING`, while the SEA path returns the collated type text verbatim (`STRING COLLATE UTF8_LCASE`, over Arrow `type_name = UNKNOWN_COLUMN_TYPE_NAME`), which yields a null `ColumnInfoTypeName`. The converter source is unchanged across recent versions; the trigger is the metadata routing, gated on the advertised driver version. (Verified: same driver + query, one warehouse answered Thrift and read fine, another answered SEA and NPE'd; the SDK 0.69→0.106 bump was ruled out.)

## Changes

- **`DatabricksTypeUtil.recoverStringType(typeText)`** (new): returns `ColumnInfoTypeName.STRING` when the given type text denotes a (possibly collated) string, else `null`. Case-insensitive (`Locale.ROOT`) and matches on a word boundary, so `STRING`, `STRING COLLATE ...` and `STRING(...)` are recovered but a hypothetical future `STRINGVIEW` is not.
- **`ArrowToJavaObjectConverter.convert`** (value path): when `requiredType` is null, recover it via `recoverStringType(arrowMetadata)`. The `switch` is also guarded with an explicit `DatabricksValidationException` (logging only the metadata, not the raw cell value) so any genuinely unmappable type produces a clear error instead of a raw NPE.
- **`DatabricksResultSetMetaData`** (SEA result manifest constructor, metadata path): when the column type name is null, recover it via `recoverStringType(columnInfo.getTypeText())` so `getColumnType()` resolves to `VARCHAR` instead of `OTHER`. The collated type text is preserved so `getColumnTypeName()` still reports the server type (mirroring the existing `TIMESTAMP_NTZ` null-recovery).

Both call sites share the same recovery rule but pass their own source (arrow field metadata for the value path, manifest `type_text` for the metadata path), since those channels legitimately differ.

## Testing

- `DatabricksTypeUtilTest#testRecoverStringType` — plain/collated string in any case → `STRING`; word-boundary rejection (`STRINGVIEW`/`STRINGSET`); non-string and null → `null`.
- `ArrowToJavaObjectConverterTest`: `testCollatedStringWithNullRequiredType` (upper), `testCollatedStringLowerCaseMetadata` (lower, covers case handling), `testTypeStartingWithStringWordIsNotCoercedToString` (word boundary → `DatabricksValidationException`), `testUnresolvableTypeThrowsValidationException` (null-guard).
- `DatabricksResultSetMetaDataTest`: `testColumnsWithCollatedString` and `testColumnsWithCollatedStringLowerCase` — `getColumnType()` is `VARCHAR`, `getColumnTypeName()` preserves the collated text.
- Affected suites pass (151 tests across the three classes).
- Validated end-to-end against a live SQL warehouse whose view exposes `rule_name COLLATE UTF8_LCASE` / `rule_description COLLATE UTF8_LCASE` over the SEA path: with this change, all columns read successfully and `getColumnType()` reports `VARCHAR`, where the unpatched driver throws the NPE.

- This revision consolidates the earlier two inline prefix checks into the single `recoverStringType` helper, addressing the review: the two sites now share one case-insensitive, word-boundary rule (no case drift between value and metadata paths), and a future unmappable type fails with a clear exception rather than an NPE.
- Longer term, the more robust fix is to drive value conversion from the Arrow physical type and treat the server type name as a non-fatal string (as the databricks-sql-python and databricks-sql-go connectors do, which do not have this bug); that is a larger, hot-path change better owned by maintainers. This PR is the contained fix. See #1547 for the design discussion.

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
@@ -105,13 +105,15 @@ public class DatabricksResultSetMetaData implements ResultSetMetaData {
           }
 
           // For collated string columns (e.g. "STRING COLLATE UTF8_LCASE") getTypeName() returns
-          // null because the collated type name does not map to a ColumnInfoTypeName. Recover STRING
-          // from the typeText prefix so the java.sql type resolves to VARCHAR instead of OTHER; the
+          // null because the collated type name does not map to a ColumnInfoTypeName. Recover
+          // STRING from the typeText so the java.sql type resolves to VARCHAR instead of OTHER; the
           // original typeText is preserved so getColumnTypeName() still reports the collated type.
-          if (columnTypeName == null
-              && columnInfo.getTypeText() != null
-              && columnInfo.getTypeText().toUpperCase().startsWith(STRING)) {
-            columnTypeName = ColumnInfoTypeName.STRING;
+          if (columnTypeName == null) {
+            ColumnInfoTypeName recovered =
+                DatabricksTypeUtil.recoverStringType(columnInfo.getTypeText());
+            if (recovered != null) {
+              columnTypeName = recovered;
+            }
           }
 
           // Check if we need to convert geospatial types to string when geospatial support is
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
@@ -109,11 +109,7 @@ public class DatabricksResultSetMetaData implements ResultSetMetaData {
           // STRING from the typeText so the java.sql type resolves to VARCHAR instead of OTHER; the
           // original typeText is preserved so getColumnTypeName() still reports the collated type.
           if (columnTypeName == null) {
-            ColumnInfoTypeName recovered =
-                DatabricksTypeUtil.recoverStringType(columnInfo.getTypeText());
-            if (recovered != null) {
-              columnTypeName = recovered;
-            }
+            columnTypeName = DatabricksTypeUtil.recoverStringType(columnInfo.getTypeText());
           }
 
           // Check if we need to convert geospatial types to string when geospatial support is
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
@@ -104,6 +104,16 @@ public class DatabricksResultSetMetaData implements ResultSetMetaData {
             }
           }
 
+          // For collated string columns (e.g. "STRING COLLATE UTF8_LCASE") getTypeName() returns
+          // null because the collated type name does not map to a ColumnInfoTypeName. Recover STRING
+          // from the typeText prefix so the java.sql type resolves to VARCHAR instead of OTHER; the
+          // original typeText is preserved so getColumnTypeName() still reports the collated type.
+          if (columnTypeName == null
+              && columnInfo.getTypeText() != null
+              && columnInfo.getTypeText().toUpperCase().startsWith(STRING)) {
+            columnTypeName = ColumnInfoTypeName.STRING;
+          }
+
           // Check if we need to convert geospatial types to string when geospatial support is
           // disabled
           String typeText = columnInfo.getTypeText();
--- a/src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
@@ -92,9 +92,9 @@ public class ArrowToJavaObjectConverter {
       }
       // A collated string column (e.g. "STRING COLLATE UTF8_LCASE") is reported with a type_name
       // that does not map to any ColumnInfoTypeName, leaving requiredType null. Recover the type
-      // from the metadata prefix so the value is read as a string instead of throwing an NPE below.
-      if (requiredType == null && arrowMetadata.startsWith(STRING)) {
-        requiredType = ColumnInfoTypeName.STRING;
+      // from the metadata so the value is read as a string instead of throwing an NPE below.
+      if (requiredType == null) {
+        requiredType = recoverStringType(arrowMetadata);
       }
     }
     if (object == null) {
--- a/src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
@@ -90,9 +90,23 @@ public class ArrowToJavaObjectConverter {
       if (arrowMetadata.startsWith(GEOGRAPHY)) {
         requiredType = ColumnInfoTypeName.GEOGRAPHY;
       }
+      // A collated string column (e.g. "STRING COLLATE UTF8_LCASE") is reported with a type_name
+      // that does not map to any ColumnInfoTypeName, leaving requiredType null. Recover the type
+      // from the metadata prefix so the value is read as a string instead of throwing an NPE below.
+      if (requiredType == null && arrowMetadata.startsWith(STRING)) {
+        requiredType = ColumnInfoTypeName.STRING;
+      }
     }
     if (object == null) {
       return null;
+    }
+    if (requiredType == null) {
+      String errorMessage =
+          String.format(
+              "Unable to determine column type for value %s with metadata %s",
+              object, arrowMetadata);
+      LOGGER.error(errorMessage);
+      throw new DatabricksValidationException(errorMessage);
     }
     switch (requiredType) {
       case BYTE:
--- a/src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
@@ -101,10 +101,9 @@ public class ArrowToJavaObjectConverter {
       return null;
     }
     if (requiredType == null) {
+      // Do not log the raw cell value, which may contain sensitive data; log only the metadata.
       String errorMessage =
-          String.format(
-              "Unable to determine column type for value %s with metadata %s",
-              object, arrowMetadata);
+          String.format("Unable to determine column type from metadata %s", arrowMetadata);
       LOGGER.error(errorMessage);
       throw new DatabricksValidationException(errorMessage);
     }
--- a/src/test/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverterTest.java
+++ b/src/test/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverterTest.java
@@ -137,6 +137,22 @@ public class ArrowToJavaObjectConverterTest {
     enableArrowNullChecking();
   }
 
+  @Test
+  public void testCollatedStringWithNullRequiredType() throws Exception {
+    // A collated string column (e.g. STRING COLLATE UTF8_LCASE) is reported with a type_name that
+    // does not map to any ColumnInfoTypeName, so requiredType arrives as null. The type must be
+    // recovered from the metadata prefix and read as a string rather than throwing an NPE.
+    VarCharVector vector = new VarCharVector("collatedVector", this.bufferAllocator);
+    vector.allocateNew(1);
+    vector.set(0, "Hello".getBytes());
+    vector.setValueCount(1);
+
+    assertEquals(
+        "Hello", convert(vector, 0, null, "STRING COLLATE UTF8_LCASE", new ColumnInfo()));
+
+    vector.close();
+  }
+
   private void disableArrowNullChecking() {
     System.setProperty("arrow.enable_null_check_for_get", "false");
   }
--- a/src/test/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverterTest.java
+++ b/src/test/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverterTest.java
@@ -141,7 +141,7 @@ public class ArrowToJavaObjectConverterTest {
   public void testCollatedStringWithNullRequiredType() throws Exception {
     // A collated string column (e.g. STRING COLLATE UTF8_LCASE) is reported with a type_name that
     // does not map to any ColumnInfoTypeName, so requiredType arrives as null. The type must be
-    // recovered from the metadata prefix and read as a string rather than throwing an NPE.
+    // recovered from the metadata and read as a string rather than throwing an NPE.
     VarCharVector vector = new VarCharVector("collatedVector", this.bufferAllocator);
     vector.allocateNew(1);
     vector.set(0, "Hello".getBytes());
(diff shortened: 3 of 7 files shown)


src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
```diff
@@ -90,10 +90,23 @@ public static Object convert(
       if (arrowMetadata.startsWith(GEOGRAPHY)) {
         requiredType = ColumnInfoTypeName.GEOGRAPHY;
       }
+      // A collated string column (e.g. "STRING COLLATE UTF8_LCASE") is reported with a type_name
+      // that does not map to any ColumnInfoTypeName, leaving requiredType null. Recover the type
+      // from the metadata prefix so the value is read as a string instead of throwing an NPE below.
+      if (requiredType == null && arrowMetadata.startsWith(STRING)) {
```
vikrantpuppala: **Medium** · flagged by devil's-advocate + test reviewers, confirmed against head SHA.

The two recovery sites added in this PR normalize case **differently**:
- **Value path** (this line, `ArrowToJavaObjectConverter.java:96`): `arrowMetadata.startsWith(STRING)` — case-sensitive.
- **Metadata path** (`DatabricksResultSetMetaData.java:113`): `columnInfo.getTypeText().toUpperCase().startsWith(STRING)` — case-insensitive.

If any server build emits the collated type text in lower/mixed case (e.g. `"string collate utf8_lcase"`), the metadata path resolves the column to `VARCHAR`, but this value path's `startsWith` fails, `requiredType` stays null, and `getObject` throws `DatabricksValidationException`. The driver would then advertise a readable VARCHAR column that cannot actually be read — worse than a clean failure.

**Suggestion**: align both sites on the same case handling. Since the value-path siblings (ARRAY/STRUCT/MAP/…) are all case-sensitive by the Spark `SqlName` uppercase convention, either accept that in both places or normalize both with `toUpperCase(Locale.ROOT)`.

---

vikrantpuppala: **Medium (open question)** · flagged by devil's-advocate; consequence not verified at the branching point.

The read path (`ArrowStreamResult.java:251-253`) falls back to `columnInfo.getTypeText()` **only when `chunkIterator.getType()` returns null** — not when it returns a non-null value that maps to no known type. So this recovery keys off `arrowMetadata` (the Arrow field `Spark:DataType:SqlName`), which may differ from the manifest `type_text`.

Issue #1547 notes two warehouse builds on the same version label behaved differently (one NPE'd, one didn't). If some build's `Spark:DataType:SqlName` for a collated column is a value that doesn't start with `"STRING"`, this recovery won't fire and the read now throws `DatabricksValidationException` instead of the old NPE — still unreadable for the user.

I could **not** confirm what `Spark:DataType:SqlName` actually holds for the failing build (`UNKNOWN_COLUMN_TYPE_NAME` is a protobuf/SDK enum sentinel → null after deserialization, not an Arrow-metadata string), so this is a question rather than a confirmed defect.

**Question**: for the exact warehouse build in #1547 that NPE'd, what does the Arrow field metadata `Spark:DataType:SqlName` contain for the collated column? If it can be non-`STRING`, consider recovering when *either* `arrowMetadata` or `columnInfo.getTypeText()` starts with `"STRING"`.

---

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
```diff
@@ -104,6 +104,16 @@ public DatabricksResultSetMetaData(
...
           }
 
+          // For collated string columns (e.g. "STRING COLLATE UTF8_LCASE") getTypeName() returns
+          // null because the collated type name does not map to a ColumnInfoTypeName. Recover STRING
+          // from the typeText prefix so the java.sql type resolves to VARCHAR instead of OTHER; the
+          // original typeText is preserved so getColumnTypeName() still reports the collated type.
+          if (columnTypeName == null
+              && columnInfo.getTypeText() != null
+              && columnInfo.getTypeText().toUpperCase().startsWith(STRING)) {
+            columnTypeName = ColumnInfoTypeName.STRING;
```
vikrantpuppala: **Medium** · flagged by architecture + maintainability reviewers.

This recovery now lives in both `DatabricksResultSetMetaData` and `ArrowToJavaObjectConverter` because the SEA path has no single type-name resolution point — the metadata constructor recovers into a local variable and never writes back to the shared `ColumnInfo`, so the read path independently re-derives the null `getTypeName()`. Every future null-`type_name` server type must be patched in lockstep, and the two copies **have already drifted** on case handling (see F1).

This mirrors a pre-existing convention, so it's a maintainability cost to note rather than a blocker. **Suggested direction** (follow-up, not required here): a single `resolveTypeName(ColumnInfo)` helper applied right after SEA manifest deserialization that calls `setTypeName(...)` on the shared object, letting both consumers drop their local recovery.

---

src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
vikrantpuppala: **Low** · flagged by devil's-advocate.

A hypothetical future type whose name begins with the letters `STRING` (e.g. `"STRINGVIEW"`, `"STRINGSET"`) would be silently coerced to scalar `STRING` and read via `toString()` — producing wrong values rather than an error, precisely on the null-typeName path where a brand-new type is most likely to land. Consistent with the existing sibling prefix checks, so low priority.

**Optional**: match a word boundary — `typeText.equals("STRING") || typeText.startsWith("STRING ") || typeText.startsWith("STRING(")` — so `"STRING COLLATE …"` matches but `"STRINGVIEW"` does not.

---

src/test/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverterTest.java
```diff
@@ -137,6 +137,38 @@ public void testNullHandlingInVarCharVector() throws Exception {
     enableArrowNullChecking();
   }
 
+  @Test
+  public void testCollatedStringWithNullRequiredType() throws Exception {
```
vikrantpuppala, line 141: **Low** · flagged by test reviewer.

Both new tests feed **already-uppercase** literals, so:
- The `.toUpperCase()` on the metadata side is a no-op in the test — delete it from production and the test still passes (covered by zero assertions).
- Plain `"STRING"` via the *null*-recovery path is uncovered (the existing `testNullHandlingInVarCharVector` uses a non-null `requiredType`).

A converter case with lowercase metadata — `convert(vector, 0, null, "string collate utf8_lcase", new ColumnInfo())` expecting the value — would currently **fail** and directly surface F1. Adding a lowercase metadata assertion covers the `.toUpperCase()` call. The two added tests are otherwise sound (real Arrow vectors, independent literal expectations, would fail if the fix were reverted).

---

vikrantpuppala:

**Score**: 89/100 — MODERATE RISK

Solid, well-scoped bug fix that faithfully follows the existing in-file prefix-recovery convention. No merge blockers — every finding below is a robustness or test-coverage gap around the recovery logic, verified against the PR head SHA. The highest-value item is the case-asymmetry between the two recovery sites (F1); the lowercase test in F5 would have caught it for free. Security and language reviewers found nothing at threshold.

---

msprunck:
@vikrantpuppala thanks for the thorough review. Addressed the findings in the latest commit — replied inline under each one. Summary: consolidated the recovery into a single `DatabricksTypeUtil.recoverStringType` helper used by both the value and metadata paths (one case-insensitive, word-boundary rule, no drift between sites), verified end-to-end against the affected warehouse.

Left as a follow-up (not this PR): the larger refactor to drive value conversion from the Arrow physical type — see the design note on #1547.

src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
msprunck: Fixed. Both sites now call the shared `DatabricksTypeUtil.recoverStringType`, which normalizes with `Locale.ROOT`, so the value path and metadata path resolve a lower/mixed-case collated type text identically. Added a lowercase test on each path (which would fail under the old case-sensitive value-path `startsWith`).

msprunck: Confirmed against the affected warehouse: instrumented capture shows the converter receives `arrowMetadata = "STRING COLLATE UTF8_LCASE"` for the collated columns, so `recoverStringType(arrowMetadata)` fires and the value reads. End-to-end validation on that warehouse: all columns read and `getColumnType()` is `VARCHAR`. If a build ever reports the collated type only in the manifest `type_text` (not the Arrow field), the value path still falls back to `columnInfo.getTypeText()` via `ArrowStreamResult` when the Arrow field is null.

src/main/java/com/databricks/jdbc/api/impl/DatabricksResultSetMetaData.java
msprunck: The decision rule is now single-sourced in `recoverStringType`. I kept two call sites rather than a single write-back on `ColumnInfo` on purpose: the value path keys off the Arrow field metadata (`Spark:DataType:SqlName`) and the metadata path off the manifest `type_text` — those channels legitimately differ, so a single write-back would change which source drives recovery. Sharing the rule (not the source) removes the case drift that caused F1 without that behavior change.

src/main/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverter.java
msprunck: Fixed. `recoverStringType` matches on a word boundary (`equals("STRING")`, or `startsWith("STRING ")` / `startsWith("STRING(")`), so `STRING COLLATE ...` and `STRING(...)` are recovered but `STRINGVIEW`/`STRINGSET` are not. Added a rejection test asserting `STRINGVIEW` throws `DatabricksValidationException` rather than being coerced to STRING.

src/test/java/com/databricks/jdbc/api/impl/converters/ArrowToJavaObjectConverterTest.java
msprunck: Added: a lowercase-metadata test on both the value path (`testCollatedStringLowerCaseMetadata`) and the metadata path (`testColumnsWithCollatedStringLowerCase`) — these exercise the case-insensitive recovery and would fail under the old case-sensitive value path. Also a direct `DatabricksTypeUtilTest#testRecoverStringType` (plain/collated/any-case, word-boundary rejection, null) and the `STRINGVIEW` word-boundary test.

vikrantpuppala (approved):

vikrantpuppala:
@sreekanth-db can you do an additional review and merge please

msrathore-db (approved):
Thanks a lot for the fix and detailed RCA!

PR #1574, opened 2026-07-23
Fix PreparedStatement.getMetaData() returning null for CTE (WITH ...) queries

### LINKED ISSUE #1562: [BUG] PreparedStatement#getMetaData() returns null for CTE (WITH ...) queries
**Describe the bug**
For a query using a common table expression (`WITH ... SELECT ...`), calling
`PreparedStatement#getMetaData()` before executing the statement returns `null`
instead of the expected `ResultSetMetaData`. Equivalent plain `SELECT` queries
(without a leading CTE) return metadata correctly in the same conditions.

This is a regression: the previous driver version (`databricks-jdbc` 2.7.6,
Simba-based) returns correct metadata for the exact same CTE query in the same
scenario. The current driver (`databricks-jdbc-thin` 3.4.1) does not.

**To Reproduce**

```java
String cteQuery = "WITH CTE AS (SELECT * FROM catalog.schema.sales) SELECT * FROM CTE";

try (Connection connection = DriverManager.getConnection(url, "token", authToken);
     PreparedStatement statement = connection.prepareStatement(cteQuery)) {

  ResultSetMetaData metadata = statement.getMetaData();
  // metadata is null here

  ResultSetMetaData workingMetadata =
      connection.prepareStatement("SELECT * FROM catalog.schema.sales").getMetaData();
  // workingMetadata is correctly populated for the equivalent non-CTE query
}
```

**Expected behavior**
`getMetaData()` should return the column metadata for the CTE query without
requiring the statement to be executed first, consistent with:
- plain `SELECT` queries against the same driver version, and
- the same CTE query against `databricks-jdbc` 2.7.6 (previous driver).

 - OS: Linux
 - Java version: Java 21
 - Java vendor: OpenJDK
 - Driver Version: 3.4.1

msrathore-db:
## Summary

`PreparedStatement#getMetaData()` returns `null` for CTE (`WITH ...`) queries when called before execution, even though plain `SELECT` queries return metadata correctly. This is a regression from the legacy Simba driver (2.7.6), reported in #1562.

### Root cause

`getMetaData()` resolves result-set metadata before execution by running `DESCRIBE QUERY <sql>`, but it only does so when `DatabricksStatement.isSelectQuery(sql)` returns true. `isSelectQuery()` matches only queries that (after stripping comments/whitespace) start with `SELECT`. A CTE query starts with `WITH`, so it fails the check and `getMetaData()` falls through to `return null`.

```java
if (DatabricksStatement.isSelectQuery(sql)) {   // false for "WITH cte AS (...) SELECT ..."
  return getMetaDataFromDescribeQuery();
} else {
  return null;                                   // <-- CTE queries hit this
}
```

### Fix

Introduce `isDescribableQuery()` that recognizes exactly the query forms `DESCRIBE QUERY` accepts as a top-level query expression — `SELECT`, `WITH` (CTEs), `VALUES`, `FROM (...)`, and parenthesized selects — and use it as the guard in `getMetaData()`. Query forms that `DESCRIBE QUERY` rejects (`SHOW`, `EXPLAIN`, `DESCRIBE`, DML/DDL) continue to return `null`, so no invalid `DESCRIBE QUERY` is ever issued.

The set of accepted forms was verified empirically against a real warehouse: `DESCRIBE QUERY` succeeds for `SELECT`/`WITH`/`VALUES`/`FROM (...)`/`(SELECT ...)` and fails with a parse error (SQLSTATE 42601) for `SHOW`/`EXPLAIN`/`DESCRIBE`/`SET`/`TABLE`.

### Verification

End-to-end test against a real warehouse with the built driver:

| Query | Before | After |
|---|---|---|
| `SELECT 1 AS a, 'x' AS b` | `[a, b]` | `[a, b]` |
| `WITH cte AS (SELECT 1 AS a, 'x' AS b) SELECT * FROM cte` | **null** | `[a, b]` |
| `VALUES (1, 'a'), (2, 'b')` | **null** | `[col1, col2]` |
| `FROM (SELECT 1 AS a)` | **null** | `[a]` |
| `(SELECT 1 AS a)` | `[a]` | `[a]` |
| `SHOW TABLES` / `EXPLAIN ...` / `DESCRIBE QUERY ...` | null | null (unchanged) |

This matches the legacy Simba driver's behavior for CTE/VALUES/FROM/parenthesized-select queries.

## Tests

- Added `DatabricksStatementTest#testIsDescribableQuery` covering CTE (incl. leading whitespace/comments), VALUES, FROM, parenthesized selects, and the excluded statement forms.
- Added `DatabricksPreparedStatementTest#testGetMetaData_NoResultSet_CteQuery_UsesDescribeQuery` verifying a CTE query issues `DESCRIBE QUERY` and returns non-null metadata.
- Full `DatabricksStatementTest` and `DatabricksPreparedStatementTest` suites pass (176 tests).

Fixes #1562

NO_CHANGELOG=true


sreekanth-db (approved):

msrathore-db:
Superseded by #1581, which contains the same changes but is raised from a branch on the upstream repository instead of a fork. Continuing review there.

PR #1580, opened 2026-07-24
Use SHOW COLUMNS IN ALL CATALOGS for getColumns with null catalog (same-repo, integration-test verification)

msrathore-db:
## Summary

Same change as #1576, pushed as a **same-repo branch** (not a fork) so the integration-test gate runs with a write-scoped token. The fork PR (#1576) had `skip-integration-tests-pr` fail with `403 Resource not accessible by integration` purely because a fork's `GITHUB_TOKEN` is read-only — this PR verifies the integration checks pass on a same-repo branch.

`DatabaseMetaData.getColumns(catalog=null, ...)` now issues a single `SHOW COLUMNS IN ALL CATALOGS` statement (consistent with `getSchemas`/`getTables`) instead of enumerating every catalog and issuing a per-catalog `SHOW COLUMNS`. Older DBR versions that don't support the syntax fall back to the previous enumerate-and-fan-out behavior on a parse syntax error.

## Details

- `CommandConstants`: add `SHOW_COLUMNS_IN_ALL_CATALOGS_SQL`.
- `CommandBuilder.fetchColumnsSQL`: emit `SHOW COLUMNS IN ALL CATALOGS` when `catalogName == null` (mirrors `fetchTablesSQL` / `fetchSchemaSQL`).
- `DatabricksMetadataQueryClient.listColumns`: null-catalog request goes through the normal path; on a parse syntax error it falls back to `fetchColumnsAcrossCatalogs`, exactly like `listSchemas`.

The `SHOW COLUMNS IN ALL CATALOGS` result schema is identical to per-catalog `SHOW COLUMNS` (the `catalogName` column is populated per row), verified against a real SQL warehouse, so `COLUMN_COLUMNS` mapping needs no changes.

## Testing

- Verified `SHOW COLUMNS IN ALL CATALOGS SCHEMA LIKE '…' TABLE LIKE '…'` succeeds against a real SQL warehouse and returns the expected 13-column schema with `catalogName` populated per row.
- Added `CommandBuilderTest` null-catalog cases and `DatabricksMetadataQueryClientTest` tests for both the single-statement path and the parse-error fallback.
- Full metadata suites green locally.


msrathore-db:
This is the same commit as #1576 (approved by @sreekanth-db), re-opened as a **same-repo branch** so the integration-test gate could run with a write-scoped token. All required checks are green here — including `JDBC Integration Tests`, `Build and Run Integration Tests (PR)`, and `Security Scan` — which #1576 could not run because a fork PR gets a read-only token (it 403'd on the `skip-integration-tests-pr` stub).

Requesting an approving review so this can merge; #1576 will be closed as superseded. cc @sreekanth-db @gopalldb

sreekanth-db (approved):

PR #1638, opened 2026-08-14
Add FAST session version exchange to SEA client

aakash-saravanan-db:
## Description

Add the SEA client-token/session-version contract required for Fast Path explicit sessions:

- Create SEA sessions with `execution_mode=FAST` and seed session-local state from `CreateSessionResponse.session_version.version_id`.
- Send the highest observed session version on every execute and fold versions from execute, polling, result, and heartbeat responses.
- Guard updates by session identity and exclude reattached statement IDs so delayed or foreign responses cannot contaminate the current session.
- Validate FAST create responses and best-effort delete a created server session when its initial version is missing.

Thrift behavior is unchanged.

## Testing

- `mvn -q -DskipTests compile`
- `mvn -q spotless:check`
- `mvn -q -pl jdbc-core -Dtest=DatabricksSessionTest,DatabricksSdkClientTest,DatabricksStatementTest,DatabricksResultSetTest test`
- Full `databricks-jdbc-core` local-profile test suite

The implementation tracks the maximum version because statement completions may be observed out of order. Live endpoint validation remains pending availability of a Fast Path-capable SEA endpoint.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Added FAST explicit-session mode and session-version exchange for SQL Exec API connections.
 
 ### Updated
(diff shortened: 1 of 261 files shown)


aakash-saravanan-db:
Superseded by #1639, which uses the same commit from an upstream repository branch so same-repository CI credentials and check permissions apply.

PR #1641, opened 2026-08-14
Add Thrift-native metadata support

vuanhphung:
## Summary

Adds opt-in `EnableThriftNativeMetadata` support for metadata operations executed through SEA. The server can return Thrift-shaped metadata rows through the Statement Execution API, while the driver preserves the same filtering, normalization, error behavior, and JDBC metadata exposed by the Thrift client.

Supported operations are catalogs, schemas, tables, columns, functions, primary keys, and cross references. Procedures and procedure columns continue to use the existing SEA path.

## Request and result flow

```text
DatabaseMetaData request
  → send X-Databricks-Metadata-Operation-Type
  → when enabled and supported, request Thrift-native metadata
  → inspect ResultManifest.is_native_metadata_result
      false / absent → existing SEA SHOW-result processing
      true           → copy native rows and rebuild them with the existing
                       Thrift normalization and JDBC metadata builders
```

The manifest flag is authoritative: neither the request header nor a native-looking schema changes how a result is processed. This keeps the feature opt-in and preserves legacy behavior when the server does not return a native result.

## Requests requiring special handling

- `getTables`: the native path sends catalog and types to runtime, then reapplies the JDBC filters to returned rows. This is necessary because runtime treats catalog as a pattern, may return temporary views outside the requested catalog, and does not consistently handle empty or exact table-type filters. SEA native results reuse this processing to preserve existing JDBC-over-Thrift behavior.

  | Path | Catalog behavior | `types` behavior |
  | --- | --- | --- |
  | Native GetTables | Applies the direct-Thrift exact filter using the original JDBC catalog | `null` accepts every runtime-returned type, an empty array returns no rows, and a non-empty array is matched exactly |
  | SHOW TABLES | SQL is scoped to the resolved catalog | `null` uses the driver's supported default types because SHOW has no JDBC `types` argument |

- `getFunctions`: the runtime paths have different catalog semantics:

  | Path | Search behavior | Runtime `FUNCTION_CAT` | Driver handling |
  | --- | --- | --- | --- |
  | Native GetFunctions | Ignores the requested catalog and searches the session catalog | Always `""` | Replaces it with the original `DatabaseMetaData.getFunctions` catalog, including `null` |
  | SHOW FUNCTIONS | Searches the requested/resolved catalog | Returns the function identifier's catalog | Leaves it unchanged |

  The driver therefore saves the original JDBC catalog before resolving a catalog for SQL construction and passes that original value only when rebuilding a manifest-confirmed native result. This preserves Thrift compatibility, but it is a column-label correction rather than catalog filtering: native rows can come from catalog A and be labeled as catalog B.
- `getCrossReference`: the SQL query narrows only the foreign-key side, so the returned native rows are additionally filtered by the requested parent catalog, schema, and table.
- Native metadata failures use Thrift-compatible propagation and timeout codes instead of the legacy SHOW-query compatibility fallbacks.

## SQLState correction

Key-based metadata validation now reports the applicable SQLState (`42000` or `08000`) and keeps `EXECUTE_STATEMENT_FAILED` (`1003`) as the driver error code. Previously these errors were constructed with `DatabricksDriverErrorCode.INVALID_STATE`, which incorrectly exposed `INVALID_STATE` through `SQLException.getSQLState()`.

Tests: `mvn spotless:check`; focused core metadata suites (331 tests).

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Added `EnableThriftNativeMetadata` to request and consume supported Thrift-native SEA metadata results.
 
 ### Updated
 - `DatabaseMetaData.getColumns(...)` with a `null` catalog now issues a single `SHOW COLUMNS IN ALL CATALOGS` statement (consistent with `getSchemas`/`getTables`) instead of enumerating every catalog and issuing a per-catalog `SHOW COLUMNS`. Older DBR versions that do not support the syntax transparently fall back to the previous enumerate-and-fan-out behavior.
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
@@ -1512,6 +1512,11 @@ public class DatabricksConnectionContext implements IDatabricksConnectionContext
   }
 
   @Override
+  public boolean isThriftNativeMetadataEnabled() {
+    return getParameter(DatabricksJdbcUrlParams.ENABLE_THRIFT_NATIVE_METADATA).equals("1");
+  }
+
+  @Override
   public boolean getDisableOauthRefreshToken() {
     return getParameter(DatabricksJdbcUrlParams.DISABLE_OAUTH_REFRESH_TOKEN, "1").equals("1");
   }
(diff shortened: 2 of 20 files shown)


sreekanth-db (approved):
`getTables()` and `getColumns()` use SQL fallbacks for unsupported `SHOW ... IN ALL CATALOGS` syntax and return empty results for object-not-found or empty-pattern cases.

Can we test the new flow for these cases against older-version compute that does not support the native metadata header?

PR #1643, opened 2026-08-16
fix(thrift): retry transient transport failures on GetOperationStatus/CloseOperation polls

### LINKED ISSUE #1642: [BUG] GetOperationStatus polls not retried on IOException — orphaned server operations and "Query has been timed out due to inactivity"
## Summary

During sustained DBSQL concurrency stress tests we consistently observed queries failing with
`Query has been timed out due to inactivity` on the server side, paired with a `TTransportException`
on the client. Root cause analysis points to a single underlying gap, plus a secondary cleanup leak:

1. **Root cause** — transport-level `IOException` (stale pooled connection, TCP reset) on a
   `GetOperationStatus` poll is **never retried** on the Thrift path. `DatabricksHttpRetryHandler`
   only retries HTTP-status-code errors, not raw `IOException`, so a single transient failure
   abandons a still-running server operation.
2. **Secondary** — `CloseOperation` cleanup RPCs fail against the same broken connection during
   `Statement`/`ResultSet` close, leaving completed operations open until the server's inactivity
   timeout reaps them.

The SEA path (`UseThriftClient=0`) already retries this class of `IOException` via the SDK's
idempotent-request strategy; the Thrift path does not. The fix is to bring the Thrift path to
parity.

---

## Environment

- Driver: `databricks-jdbc` (also reproduced against Simba JDBC 3.3.1)
- Warehouse type: DBSQL Serverless
- Workload: 20+ concurrent long-running queries (TPC-DS benchmark)
- Transport: HTTP/Thrift (`DatabricksHttpTTransport`)

---

## Observed behaviour

**Server side** (`system.query.history`):

```
execution_status : FAILED
error_message    : Query has been timed out due to inactivity.
total_duration_ms: ~430 000  (query ran ~430 s after the client had already moved on)
```

**Client side** (exception propagated from `pollTillOperationFinished`):

```
com.databricks.jdbc.exception.DatabricksSQLException:
  ... TTransportException: <connection reset / No route to host / stale connection>
```

The operation continues running server-side for ~430 seconds after the client exception, then the
server records the inactivity failure. In a second variant (9 occurrences), a completed operation
whose `CloseOperation` RPC failed remained open for ~23 minutes before the server reaped it.

> Note: the ~430 s / ~23 min windows are observed empirically from `system.query.history`; they
> reflect server-side reaping behaviour, not something visible in the driver source.

---

### 1 — `retryRequest` does not retry plain `IOException` (stale connection, TCP reset)

**`DatabricksHttpRetryHandler.java`, `retryRequest()` (lines 130–134):**

```java
@Override
public boolean retryRequest(IOException exception, int executionCount, HttpContext context) {
    int statusCode = getErrorCodeFromException(exception);
    if (!isStatusCodeRetryable(statusCode)) {
        return false;   // ← returns false for every raw IOException
    }
    // ... only reaches here for DatabricksRetryHandlerException wrapping 429/503
}
```

`getErrorCodeFromException` (lines 251–256) only extracts a status code from
`DatabricksRetryHandlerException` (thrown by the `process()` interceptor for HTTP-level errors);
for any other `IOException` it returns `0`:

```java
static int getErrorCodeFromException(IOException exception) {
    if (exception instanceof DatabricksRetryHandlerException) {
        return ((DatabricksRetryHandlerException) exception).getErrCode();
    }
    return 0;
}
```

`isStatusCodeRetryable(0)` falls through to `default: return false` (line 298). The result: **any
transport-level failure on a `GetOperationStatus` poll — stale pooled connection, TCP RST,
load-balancer reset — is treated as unrecoverable and not retried.**

### 2 — The polling loop delegates retry to the HTTP layer, which then doesn't retry

**`DatabricksThriftAccessor.java`, `pollTillOperationFinished()` (lines 339–364):**

```java
try {
    statusResp = getOperationStatus(statusReq, statementId);
} catch (TTransportException e) {
    // "after retries were exhausted"
    throw buildTransportFailureException(statementId.toSQLExecStatementId(), e);
}
```

The poll loop has no retry of its own — the comment at the catch site assumes retries already
happened at the HTTP layer. But because of gap 1 they didn't. The raw `IOException` surfaces from
`DatabricksHttpTTransport.flush()`, which wraps it as a `TTransportException` (lines 138–150) with no
retry:

```java
try (CloseableHttpResponse response = httpClient.execute(request)) {
    ...
} catch (DatabricksHttpException | IOException e) {
    ...
    throw new TTransportException(TTransportException.UNKNOWN, errorMessage, e);
}
```

One failed status poll → `TTransportException` → caller abandons the operation → server runs until
the ~430 s inactivity window.

### 3 — `closeServerOperation` swallows the failure; the follow-up close also fails and the operation leaks

**`DatabricksStatement.java`, `closeServerOperation()` (lines 1026–1034):**

```java
} catch (SQLException | RuntimeException e) {
    // Best-effort — don't fail the user's close for a server cleanup failure.
    // serverOperationClosed stays false so Statement.close() will retry the RPC.
    LOGGER.warn("Failed to proactively close server operation for statement {}: {}", ...);
}
```

If the same class of connection error hits during `closeServerOperation()` (called from
`ResultSet.close()`), the `CloseOperation` RPC is dropped and `serverOperationClosed` stays `false`.
`Statement.close()` then does retry the RPC (`DatabricksStatement.java:165–166`), but against the
same broken connection it fails again. Unlike the proactive close, this second failure is **not**
swallowed — `close(boolean)` is a `try { … } finally { … }` with no `catch` (lines 142–194), so the
`SQLException` propagates to the caller. Either way the operation is left alive server-side and is
reaped by the inactivity timeout (~23 minutes in the observed cases).

### Why this only affects long-running concurrent workloads

HTTP connection pooling (`PoolingHttpClientConnectionManager`) reuses persistent connections
across the 200 ms poll-sleep intervals (`POLL_INTERVAL` default `"200"`). Under high concurrency,
pooled connections can become stale (server-side close of a keep-alive connection, AWS NAT gateway
timeout, load-balancer idle reset). Because polls reuse a connection every ~200 ms, Apache
HttpClient's idle-revalidation does not kick in between polls, so a stale connection is used and
throws `IOException` mid-flight. Per gap 1, there is no transparent reconnect-and-retry.

---

### Fix 1 (primary) — Retry transport-level `IOException` on safe polling/cleanup RPCs

`GetOperationStatus` and `CloseOperation` are read-only or idempotent from the server's perspective
(polling a running operation or closing an already-completed one causes no side effects on retry),
so a transport failure should transparently reconnect and retry. Two implementation options:

**Option A (preferred) — retry at the poll-loop level**, where the RPC identity and operation handle
are already known, so no request-tagging plumbing is needed:

```java
// DatabricksThriftAccessor.pollTillOperationFinished()
try {
    statusResp = getOperationStatus(statusReq, statementId);
    transportRetries = 0; // reset on success
} catch (TTransportException e) {
    if (++transportRetries <= MAX_TRANSPORT_RETRIES) {
        LOGGER.warn("Retrying GetOperationStatus after transport failure "
            + "(attempt {}): {}", transportRetries, e.getMessage());
        // brief backoff, then continue the loop with a fresh connection
        continue;
    }
    throw buildTransportFailureException(statementId.toSQLExecStatementId(), e);
}
```

**Option B — retry inside `DatabricksHttpRetryHandler.retryRequest`** for safe RPCs. Note this is
**not** a one-line change: `DatabricksHttpTTransport.flush()` currently calls
`httpClient.execute(request)` with no `HttpContext` and only sees opaque serialized Thrift bytes, so
it cannot know which RPC it is flushing. This option requires new plumbing to tag each request with
its RPC name (e.g. a transport field or thread-local set before `flush()`, plus passing an
`HttpContext` into `execute`) so the handler can allow-list `GetOperationStatus` / `CloseOperation`.

The same cleanup path should also be applied to `CloseOperation` (see Fix 2) so the leak in root
cause 3 is closed as well.

### Fix 2 — Make `CloseOperation` failures recoverable instead of best-effort-and-forget

Even with Fix 1, `closeServerOperation` should not depend on a single follow-up attempt against a
connection that is already broken. Options: retry the close on a fresh connection, or schedule a
deferred retry via the session's cleanup path rather than only relying on `Statement.close()`:

```java
// DatabricksStatement.java — closeServerOperation()
} catch (SQLException | RuntimeException e) {
    LOGGER.warn("Failed to close server operation for statement {}: {}", statementId, e);
    // Schedule a deferred retry via the session's cleanup queue rather than
    // relying on a single follow-up attempt in Statement.close().
    connection.getSession().scheduleCleanup(statementId);
}
```

### Fix 3 (optional, narrow) — Heartbeat for slow-consumer results only

`ENABLE_HEARTBEAT` defaults to `"0"` (`DatabricksJdbcUrlParams.java:252–255`). Enabling heartbeat
sends periodic `GetOperationStatus` pings while results are consumed slowly
(`ResultHeartbeatManager` / `DatabricksResultSet.startHeartbeatIfEnabled`).

This is **not** a fix for the main problem, and we do not recommend flipping the default on the basis
of this issue:

- It cannot affect the dominant failure mode (transport failure *during polling*): the heartbeat is
  only started once a `DatabricksResultSet` exists, which is *after* the operation has finished — so
  it is never running while a still-`RUNNING` operation is being polled.
- It is gated by `isHeartbeatEligible()` (excludes SEA-inline, direct/`CLOSED` results, and
  `PENDING`/`RUNNING`) and has a 60 s initial delay (`HEARTBEAT_INTERVAL_SECONDS` default `"60"`), so
  it only matters for a genuinely stalled consumer holding an eligible open result set for >60 s.
- Its benefit assumes a `GetOperationStatus` poll resets the server's inactivity timer, which is a
  server-side assumption not verifiable from the driver.

Keep it as an opt-in mitigation for slow-consumer scenarios, not as the remedy for this issue.

---

## Impact

- Operations are abandoned after a single transient HTTP failure during polling, even though the
  server-side query is still running and could be re-polled with a fresh connection.
- Completed operations are leaked to the server's inactivity timeout, holding resources for up to
  ~23 minutes per operation.
- The transport-failure mode surfaces only as a generic `TTransportException`, with no indication
  that the server is still carrying orphaned work.

---

## Files

| File | Location of gap |
|---|---|
| `DatabricksHttpRetryHandler.java` | `retryRequest()` (130–134) + `getErrorCodeFromException()` (251–256) — returns `false`/`0` for all plain `IOException` |
| `DatabricksThriftAccessor.java` | `pollTillOperationFinished()` (347–349) — no retry on `TTransportException`; `getOperationStatus()` (1042) — bare RPC |
| `DatabricksHttpTTransport.java` | `flush()` (128, 138–150) — executes with no `HttpContext`; wraps `IOException` as `TTransportException` |
| `DatabricksStatement.java` | `closeServerOperation()` (1026–1034) — swallows close failure; `close(boolean)` (142–194) retries once and then propagates |
| `DatabricksJdbcUrlParams.java` | `ENABLE_HEARTBEAT` (252) — disabled by default (relevant only to Fix 3) |

---

## Behaviour difference between `UseThriftClient=1` and `UseThriftClient=0`

Both paths share the same close-cleanup gap, but differ on the poll-retry and interrupt behaviour.

### Mode A — IOException on a status poll

On the **Thrift path** (`UseThriftClient=1`), `GetOperationStatus` is a **POST** routed through
`DatabricksHttpRetryHandler`. As above, `retryRequest()` returns `false` for all plain
`IOException`, so a single stale connection immediately abandons the operation.

On the **SEA path** (`UseThriftClient=0`), status polling is a **GET** to
`/api/2.0/sql/statements/{id}`, which goes through the Databricks SDK's `ApiClient`. The SDK wraps
any `IOException` into `DatabricksError("IO_ERROR", 523, e)` (`ApiClient.java:274–283`) and applies
`RequestBasedRetryStrategyPicker`: `GET /api/2.0/sql/statements/.*` is explicitly listed as an
**idempotent request**, so it receives `IdempotentRequestRetryStrategy`. That strategy's
`isRetriable()`:

```java
public boolean isRetriable(DatabricksError databricksError) {
    if (RetryUtils.isCausedByTransientError(databricksError)) { return true; }
    if (isNonRetriableException(databricksError))              { return false; }
    if (isNonRetriableHttpCode(databricksError))               { return false; }
    return true;  // ← default: retry everything else, including IOException (523)
}
// NON_RETRIABLE_HTTP_CODES = {400,401,403,404,405,409,410,411,412,413,414,415,416}
// 523 is not in the set → isRetriable returns true → IOException IS retried on SEA
```

**Result: the SEA path retries `IOException` on status polls; the Thrift path does not.** (Verified
in `databricks-sdk-java`: `RequestBasedRetryStrategyPicker`, `IdempotentRequestRetryStrategy`,
`ApiClient`.)

### Mode A — Thread interrupt

On the **Thrift path**, `InterruptedException` in the poll sleep triggers a `cancelOperation` RPC
(`DatabricksThriftAccessor.java:357–363`). That cancel is *attempted*; if the underlying connection
is broken it throws `DatabricksHttpException` (from `cancelOperation`, lines 164–174) which
propagates — so the client is not left unaware, though the server op may still be orphaned because
the cancel never landed.

On the **SEA path**, `InterruptedException` immediately throws `DatabricksTimeoutException`
**without sending any cancel** (`DatabricksSdkClient.java:278–285`) — the operation is orphaned even
faster.

### Mode B — closeOperation fails

Both paths share `DatabricksStatement.closeServerOperation()`. The proactive close is best-effort
(swallowed); `Statement.close()` retries once and then propagates on failure. **Mode B is identical
on both paths.**

### Summary

| Failure scenario | Thrift (`=1`) | SEA (`=0`) |
|---|---|---|
| IOException on status poll | ❌ Not retried, operation orphaned | ✅ Retried by SDK idempotent strategy |
| Thread interrupt | ⚠️ Cancel attempted; throws (not silent) if transport broken | ❌ No cancel sent — worse |
| closeOperation IOException | ❌ Leaked (proactive swallowed; retry propagates) | ❌ Leaked (same shared path) |

Switching to `UseThriftClient=0` mitigates Mode A poll failures but does not fix the close leak and
makes the interrupt case worse. The proper fix is in the shared cleanup path and in the Thrift poll
retry, regardless of client.

---

## Related

**#1571** appears to track a sibling gap in the same method — `DatabricksHttpRetryHandler.retryRequest`
aborts when `retryInterval == -1` and the status code isn't in `ApiRetriableHttpCodes` (line 142),
which would drop header-less 503/429 on the Thrift path. The mechanism is consistent with the code
here (same file/method), though we haven't re-verified the specifics of #1571 itself. A
`retryRequest`/poll-loop fix that also handles raw `IOException` for safe RPCs would be
complementary and could land in the same PR.

A couple of judgment calls I made — tell me if you'd rather go the other way:
- I made Fix 1 the primary and led with poll-loop-level retry (Option A) since it avoids the flush() plumbing problem, keeping Option B as an alternative.
- I kept #1571 but softened it to "appears to / haven't re-verified," since I validated the mechanism against the code but didn't fetch that issue. If you want, I can fetch #1571 and either firm this up or drop it.

diegofanesi:
## Summary

Fixes #1642.

Long-running queries on the Thrift client path (`UseThriftClient=1`) could fail with `Query has been timed out due to inactivity` under sustained concurrency. A single transient transport-level failure (stale pooled connection, TCP reset, load-balancer idle drop) on a `GetOperationStatus` poll abandoned the still-running server operation — the driver surfaced a `TTransportException` and stopped polling, so the server kept running the query until its inactivity window elapsed and reaped it. The same failure class on `CloseOperation` / `CancelOperation` could leak completed operations until the server reaped them too.

## Root cause

`DatabricksHttpRetryHandler.retryRequest` only retries HTTP status-code errors (429/503) that arrive wrapped in a `DatabricksRetryHandlerException`. A raw `IOException` (stale pooled connection, connection reset) maps to status code `0` via `getErrorCodeFromException`, which `isStatusCodeRetryable` rejects — so it is never retried. The Thrift polling loop in `DatabricksThriftAccessor` has no retry of its own either, so one transient poll failure permanently abandons an otherwise-healthy server operation.

## Change

Adds a bounded, jittered exponential-backoff retry (`withTransportRetry`) in `DatabricksThriftAccessor` and routes the idempotent Thrift RPCs through it:

- `GetOperationStatus` — both the main execution polling loop and the metadata polling loop (also covers the Thrift heartbeat, which shares the same accessor method)
- `CloseOperation` and `CancelOperation`

Each RPC invocation builds a fresh transport, so a retry naturally leases a new pooled connection while the broken one is discarded. This lets a still-running server operation be re-polled — or a completed one be re-closed / cancelled — instead of being abandoned after a single connection blip.

Details:
- Bounded to **5 retries** with **full-jitter backoff** (1s → 16s cap), interrupt-aware (a thread interrupt during a backoff sleep restores the interrupt flag and aborts the retry loop).
- The original `TTransportException` is rethrown once retries are exhausted, so existing caller-side failure handling (`buildTransportFailureException`, etc.) is fully preserved.
- Only **transport-level** failures are retried; non-transport `TException`s propagate on the first attempt.
- **Statement submission (`ExecuteStatement`) is deliberately excluded** from the retry path — re-sending it could double-execute the query. Only read-only / idempotent RPCs retry.

## Scope

- **Thrift client path only** (`UseThriftClient=1`). All changes live in `DatabricksThriftAccessor`.
- The **SEA client path is unchanged** — it already retries this class of `IOException` via the SDK's idempotent-request retry strategy (`GET /api/2.0/sql/statements/*` is treated as idempotent, and an `IOException` maps to a retriable synthetic `523`).

## Files changed

- `src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java` — retry helper + call-site wiring
- `NEXT_CHANGELOG.md` — `Fixed` entry

## Testing

- Verified the retry helper's generics, lambda capture, exception flow, and backoff math via a standalone compile + run.
- Note: the full Maven build was not run locally (no network access to Maven Central in the authoring environment); please rely on CI for the full build + `spotless` check. If `spotless` flags formatting, it is cosmetic and auto-fixable with `mvn spotless:apply`.


diegofanesi:
Please Note: **this is currently untested!** I am unable to test this from my laptop so I am looking for a way to test this somewhere. I will update this as soon as my tests are complete.

diegofanesi:
tests completed. ready to merge on my end. 

I run 3 parallel tests on three large warehouses for a total of almost 10000 queries launched in parallel on a 10 TB dataset. this lasted 24h and I had only 1 query failure related to a different issue. Normally I would have seen multiple queries failing under such conditions. 

I can conclude that the stability has been improved significantly with this patch.

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -962,6 +983,78 @@ private DatabricksSQLException buildTransportFailureException(
...
+   * retry loop.
+   */
+  private <T> T withTransportRetry(String rpcName, String statementId, TransportSafeRpc<T> rpc)
+      throws TException {
+    int attempt = 0;
+    long backoffMillis = TRANSPORT_RETRY_MIN_BACKOFF_MILLIS;
+    while (true) {
+      try {
+        return rpc.call();
+      } catch (TTransportException e) {
```
gopalldb, line 1103: Retry catches all TTransportExceptions, including permanent HTTP errors (DatabricksThriftAccessor.java:1009)
DatabricksHttpTTransport.flush() wraps every non-2xx response (via ValidationUtil.checkHTTPError → DatabricksHttpException) into a TTransportException. So a poll that hits a 401 (expired token), 403, or 404 (invalid handle) is now retried 5 times with up to ~31s cumulative backoff before failing — turning an instant permanent error into a 30s+ hang. There's no transient/permanent discrimination.

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -962,6 +983,78 @@ private DatabricksSQLException buildTransportFailureException(
...
+   * inactivity timeout. Only RPCs that are safe to repeat may use this path (status polling,
+   * operation close, cancel); statement submission must not.
+   *
+   * <p>Retries are bounded by {@link #MAX_POLL_TRANSPORT_RETRIES}. Once exhausted, the original
+   * {@link TTransportException} is rethrown so the existing caller-side failure handling still
+   * applies. Non-transport {@link TException}s are never retried — they propagate on the first
+   * attempt. A thread interrupt during a backoff sleep restores the interrupt flag and aborts the
+   * retry loop.
+   */
+  private <T> T withTransportRetry(String rpcName, String statementId, TransportSafeRpc<T> rpc)
```
gopalldb: Retry loop ignores the TimeoutHandler, so queries overshoot queryTimeout (DatabricksThriftAccessor.java:1009)
pollTillOperationFinished and fetchMetadataResults call timeoutHandler.checkTimeout() only at the top of the while loop, before withTransportRetry. A single transport-failure sequence sleeps up to 1+2+4+8+16 = 31s inside the retry with no timeout check, so a statement with queryTimeout=5s blocks ~31s past its deadline before the timeout is enforced

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -962,6 +983,78 @@ private DatabricksSQLException buildTransportFailureException(
...
+      throws TException {
+    int attempt = 0;
+    long backoffMillis = TRANSPORT_RETRY_MIN_BACKOFF_MILLIS;
+    while (true) {
```
gopalldb, line 1100: Retry amplification with the existing HTTP-layer 429/503 retry (DatabricksThriftAccessor.java:1009)
DatabricksHttpRetryHandler already retries 429/503. When it exhausts and surfaces the failure as DatabricksHttpException → TTransportException, withTransportRetry retries 5 more times, each triggering a fresh full HTTP-layer retry cycle. Against a rate-limited or unavailable endpoint this produces multiplicative retry storms (HTTP retries × 5 outer retries), worsening load on a recovering cluster.

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
diegofanesi: it makes perfect sense. however, there are other 4xx errors that should be retried like "429 too many requests". This means I have to be very specific on the errors I exclude from the retrial. will try to do so.

diegofanesi: fixed

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
diegofanesi: fixed

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
diegofanesi: fixed

gopalldb (approved):

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -962,6 +999,222 @@ private DatabricksSQLException buildTransportFailureException(
...
+   *   <li>{@link DatabricksHttpException} carrying a concrete HTTP status → retryable only if the
+   *       status is a transient gateway code ({@link #RETRYABLE_TRANSPORT_HTTP_CODES}); permanent
+   *       statuses (401/403/404 …) are not.
+   *   <li>{@link DatabricksHttpException} with no status (status {@code 0}) → a genuine
+   *       connection-level failure (stale pooled connection, reset, socket timeout) surfaced as an
+   *       {@link IOException} cause; retryable.
+   *   <li>A direct {@link IOException} cause (e.g. a response-body read error) → retryable.
+   *   <li>No cause at all → retryable. The transport raises bare {@link TTransportException}s (e.g.
+   *       {@code DatabricksHttpTTransport.read()} on a truncated / empty response body) that are
+   *       plausibly transient; treating them as retryable preserves the pre-existing behavior of
```
vuanhphung, line 1177: can you explain this comment? I thought pre-existing behavior is no retry?

src/main/java/com/databricks/jdbc/dbclient/impl/thrift/DatabricksThriftAccessor.java
```diff
@@ -962,6 +999,222 @@ private DatabricksSQLException buildTransportFailureException(
...
+   * storm on top of the HTTP layer's.
+   *
+   * <p>When {@code timeoutHandler} is non-null the operation's deadline is enforced before each
+   * backoff sleep and the sleep is capped to the remaining budget, so a failing RPC cannot overshoot
+   * the statement's {@code queryTimeout}. Retries are bounded by {@code maxRetries}; once exhausted
+   * the original {@link TTransportException} is rethrown so existing caller-side handling still
+   * applies. A thread interrupt during a backoff sleep restores the interrupt flag and aborts the
+   * retry loop.
+   */
+  private <T> T withTransportRetry(
```
vuanhphung, line 1091: the logic looks good to me, but I wonder if we can separate this retry logic out of here and do sth like
```
class RetryingTCLIService implements TCLIService.Iface {
    private final Supplier<TCLIService.Iface> factory;   // = this::newThriftClient
    RetryingTCLIService(Supplier<TCLIService.Iface> factory) { this.factory = factory; }

    @Override public TGetOperationStatusResp GetOperationStatus(TGetOperationStatusReq r) throws TException {
      return withTransportRetry("GetOperationStatus", POLL_BUDGET, () -> factory.get().GetOperationStatus(r));
    }
    @Override public TCloseOperationResp CloseOperation(TCloseOperationReq r) throws TException { ... cleanup budget ... }
    @Override public TCancelOperationResp CancelOperation(TCancelOperationReq r) throws TException { ... }

    // ExecuteStatement, OpenSession, FetchResults, ~35 others: straight pass-through
    @Override public TExecuteStatementResp ExecuteStatement(TExecuteStatementReq r) throws TException {
      return factory.get().ExecuteStatement(r);
    }
    ...
  }
  
Then getThriftClient() returns new RetryingTCLIService(this::newThriftClient).
```

We can implement the same retry logic for other idempotent requests like metadata

vuanhphung (approved):

vikrantpuppala:
@diegofanesi can we do a rebase and dco signoff so that we can get this merged? thanks for the contribution!

PR #1645, opened 2026-08-18
Fix logging when an OFF connection initializes first

cathleeny:
## Description

Fix process-wide JUL initialization when the first connection uses `LogLevel=OFF`.

- `OFF` suppresses the shared parent logger without creating a handler or permanently completing initialization.
- The first logging-enabled connection installs the single shared handler.
- Once enabled, later connections—including `OFF` connections—do not reconfigure the logger.
- Failed handler creation remains retryable.

## Testing

- Created unit tests that verifies that `OFF -> TRACE` enables logging, `TRACE -> OFF` continues to log at `TRACE` level, concurrent initializations do not create duplicate handlers.
- Packaged thin and uber jars: manually verified `OFF -> TRACE` creates one  handler, while `TRACE -> OFF` retains the original handler and level.

This is the short-term fix and intentionally keeps handler creation inside `JulLogger.initLogger()`.


cathleeny:
Continued in #1670. The original source fork was deleted, so the unchanged commit was recovered and pushed directly to this repository.

PR #1670, opened 2026-08-27
Fix logging when an OFF connection initializes first

cathleeny:
## Description

Fix process-wide JUL initialization when the first connection uses `LogLevel=OFF`.

- `OFF` suppresses the shared parent logger without creating a handler or permanently completing initialization.
- The first logging-enabled connection installs the single shared handler.
- Once enabled, later connections—including `OFF` connections—do not reconfigure the logger.
- Failed handler creation remains retryable.

## Testing

- Unit tests cover `OFF -> TRACE`, `TRACE -> OFF`, concurrent initialization without duplicate handlers, and retry after failed handler creation.
- `JulLoggerTest` and `LoggingUtilTest`: 30 tests passed.
- Thin and uber JARs were manually verified: `OFF -> TRACE` creates one handler, while `TRACE -> OFF` retains the original handler and level.

This is the short-term fix and intentionally keeps handler creation inside `JulLogger.initLogger()`.

### DIFF
--- a/src/main/java/com/databricks/jdbc/log/JulLogger.java
+++ b/src/main/java/com/databricks/jdbc/log/JulLogger.java
@@ -136,7 +136,7 @@ public class JulLogger implements JdbcLogger {
     Logger jdbcJulLogger = Logger.getLogger(PARENT_CLASS_PREFIX);
     jdbcJulLogger.setUseParentHandlers(false);
 
-    if (level == Level.OFF) {
+    if (level.intValue() == Level.OFF.intValue()) {
       jdbcJulLogger.setLevel(Level.OFF);
       return;
     }
--- a/src/main/java/com/databricks/jdbc/log/JulLogger.java
+++ b/src/main/java/com/databricks/jdbc/log/JulLogger.java
@@ -114,7 +114,9 @@ public class JulLogger implements JdbcLogger {
 
   /**
    * Initializes the logger with the specified configuration. This method is synchronized to prevent
-   * concurrent modifications to the logger configuration.
+   * concurrent modifications to the logger configuration. A Level.OFF request suppresses JUL output
+   * without installing a handler, allowing a later enabled request to initialize the shared logger.
+   * Once an enabled handler is installed, subsequent requests do not reconfigure it.
    *
    * @param level the log level
    * @param logDir the directory for log files or {@code STDOUT} for console output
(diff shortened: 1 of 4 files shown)


src/main/java/com/databricks/jdbc/log/JulLogger.java
```diff
@@ -124,35 +126,41 @@ public void error(Throwable throwable, String format, Object... arguments) {
...
-                flush();
-              }
-            };
-      } else {
-        handler = new FileHandler(logPattern, logFileSizeBytes, logFileCount, true);
-      }
-      handler.setLevel(level);
-      handler.setFormatter(new Slf4jFormatter());
-      jdbcJulLogger.addHandler(handler);
+    if (isLoggerInitialized) {
```
gopalldb, line 129: the issue will still remain, logger will remain disabled if first connection had it disabled

src/main/java/com/databricks/jdbc/log/JulLogger.java
```diff
@@ -124,35 +126,41 @@ public void error(Throwable throwable, String format, Object... arguments) {
...
+      return;
+    }
+
+    // java.util.logging uses hierarchical loggers, so we just need to set the log level on the
+    // parent package logger. Using "com.databricks" as the prefix captures all JDBC driver
+    // classes as well as shaded dependencies (SDK, Apache HTTP client, etc.)
+    Logger jdbcJulLogger = Logger.getLogger(PARENT_CLASS_PREFIX);
+    jdbcJulLogger.setUseParentHandlers(false);
+
+    if (level == Level.OFF) {
```
gopalldb: from isaac, better comparison would be:

level.intValue() == Level.OFF.intValue()

src/main/java/com/databricks/jdbc/log/JulLogger.java
```diff
@@ -124,35 +126,41 @@ public void error(Throwable throwable, String format, Object... arguments) {
...
-      jdbcJulLogger.addHandler(handler);
+    if (isLoggerInitialized) {
+      return;
+    }
+
+    // java.util.logging uses hierarchical loggers, so we just need to set the log level on the
+    // parent package logger. Using "com.databricks" as the prefix captures all JDBC driver
+    // classes as well as shaded dependencies (SDK, Apache HTTP client, etc.)
+    Logger jdbcJulLogger = Logger.getLogger(PARENT_CLASS_PREFIX);
+    jdbcJulLogger.setUseParentHandlers(false);
```
gopalldb, line 137: do we want to set this on OFF only path? This can be done after checking OFF level

src/main/java/com/databricks/jdbc/log/JulLogger.java
cathleeny: I verified the OFF to enabled change. The OFF path returns without setting `isLoggerInitialized`, so the later enabled call installs a handler, updates the logger level, and marks initialization complete. `testInitLoggerPromotesFromOffToEnabled` covers this, and I also tested locally

src/main/java/com/databricks/jdbc/log/JulLogger.java
cathleeny: The OFF path needs it to prevent logs from propagating to root handlers, so both the OFF path and enabled path need this

gopalldb (approved):

PR #1672, opened 2026-09-02
Enable native SEA metadata defaults behind server flag

vuanhphung:
## Description

Sets `UseBoundedSeaApi` and `EnableThriftNativeMetadata` defaults to `1`. When unset, both features require the existing server-side `databricks.partnerplatform.clientConfigsFeatureFlags.enableSqlExecForJdbc` flag on SQL warehouses; explicit JDBC settings continue to take precedence.

## Testing

- `mvn test -pl jdbc-core -Dtest=DatabricksConnectionContextTest#testNativeMetadataViaSea*` (4 passed)
- `mvn spotless:check`
- Full `DatabricksConnectionContextTest` run: 157 passed; one existing Mockito test could not initialize Byte Buddy attachment on the local JDK.

## Telemetry Errors

- [x] Not applicable — this PR does not add or change a telemetry-visible error.
- [ ] Applicable — the error uses `DatabricksDriverErrorCode` where appropriate, and any new code is uniquely numbered and tested.
- [ ] Applicable — its driver/server/user classification is linked, or maintainer help is requested because the author cannot access the classification.

The existing SQL Exec server flag controls the rollout only when the JDBC properties are not explicitly configured.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added `EnableThriftNativeMetadata` to request and consume supported Thrift-native SEA metadata results.
 
 ### Updated
+- `UseBoundedSeaApi` and `EnableThriftNativeMetadata` now default to `1`; when unset, activation is controlled by the server-side `enableNativeMetadataViaSEA` rollout flag.
 - `DatabaseMetaData.getColumns(...)` with a `null` catalog now issues a single `SHOW COLUMNS IN ALL CATALOGS` statement (consistent with `getSchemas`/`getTables`) instead of enumerating every catalog and issuing a per-catalog `SHOW COLUMNS`. Older DBR versions that do not support the syntax transparently fall back to the previous enumerate-and-fan-out behavior.
 - Updated bundled Jackson, lz4-java, Netty, and Apache HttpComponents Client and Core dependencies to patched versions to address security findings.
 
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,7 +6,7 @@
 - Added `EnableThriftNativeMetadata` to request and consume supported Thrift-native SEA metadata results.
 
 ### Updated
-- `UseBoundedSeaApi` and `EnableThriftNativeMetadata` now default to `1`; when unset, activation is controlled by the server-side `enableNativeMetadataViaSEA` rollout flag.
+- `UseBoundedSeaApi` and `EnableThriftNativeMetadata` now default to `1`; when unset, activation is controlled by the server-side `enableSqlExecForJdbc` rollout flag.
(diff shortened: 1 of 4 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksConnectionContext.java
```diff
@@ -49,6 +49,9 @@ public class DatabricksConnectionContext implements IDatabricksConnectionContext
   private static final String USE_QUERY_FOR_THRIFT_FLAG_NAME =
       "databricks.partnerplatform.clientConfigsFeatureFlags.enableUseQueryForThriftJdbc";
 
+  private static final String NATIVE_METADATA_VIA_SEA_FLAG_NAME =
+      "databricks.partnerplatform.clientConfigsFeatureFlags.enableNativeMetadataViaSEA";
+
```
rahuls-db: discussed offline. This flag is not needed. We can piggy back on existing flag SQL_EXEC_FLAG_NAME in driver. Connector service can map its value to different set of workspaces depending on driver version

src/main/java/com/databricks/jdbc/common/DatabricksJdbcUrlParams.java
```diff
@@ -215,11 +215,11 @@ public enum DatabricksJdbcUrlParams {
   USE_BOUNDED_SEA_API(
       "UseBoundedSeaApi",
       "Use bounded SEA API for CloudFetch: send row_offset on GetResultData, force StreamingChunkProvider, stop relying on total_chunk_count. Requires server support.",
-      "0"),
+      "1"),
   ENABLE_THRIFT_NATIVE_METADATA(
       "EnableThriftNativeMetadata",
       "Request Thrift-native SEA results for catalogs, schemas, tables, columns, functions, primary keys, imported keys, and cross references",
-      "0"),
+      "1"),
```
rahuls-db, line 222: is this is set to 1 by default, won't this make server side flag redundant?

vuanhphung: @rahuls-db The logic of `resolveFeatureFlag` is as follow:
1. If caller explicitly pass this connection param in, use it.
2. Otherwise, if the default value is 1, use feature flag value
3. Else, return false

NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Added `EnableThriftNativeMetadata` to request and consume supported Thrift-native SEA metadata results.
 
 ### Updated
+- `UseBoundedSeaApi` and `EnableThriftNativeMetadata` now default to `1`; when unset, activation is controlled by the server-side `enableSqlExecForJdbc` rollout flag.
```
rahuls-db, line 9: This is public facing release note. Do we need to talk about `enableSqlExecForJdbc` or should we remove it or leave it as - `UseBoundedSeaApi and EnableThriftNativeMetadata now default to 1.`

Also, EnableThriftNativeMetadata is a new param introduced for the first time in this release. 

@gopalldb any guidance?

sreekanth-db: we usually include what the new parameter does and how to opt out if if user prefers to not have this change.

sreekanth-db (approved):

PR #1678, opened 2026-09-05
Add session version exchange to SQL Exec API client

aakash-saravanan-db:
## Description

- Create SQL Exec API sessions with `execution_mode=FAST`.
- Track the maximum server-provided `session_version` per session and include it in subsequent execute requests.
- Update the tracker from create-session, the initial execute response, and synchronous polling within the originating execute call.
- Keep detached async polling non-authoritative because statement results can be retrieved through another connection.
- Document the Lakehouse Real-Time async session-state limitation.

Create-session responses without a version remain supported. Execute requests omit `session_version` until the server supplies one.

## Testing

- `mvn -q spotless:check`
- `mvn -q -pl jdbc-core -Dtest=DatabricksSessionTest,DatabricksSdkClientTest,SessionVersionJsonTest test` (87 tests, with the Byte Buddy agent required by the local FIPS host)

## Telemetry Errors

- [x] Not applicable — this PR does not add or change a telemetry-visible error.
- [ ] Applicable — the error uses `DatabricksDriverErrorCode` where appropriate, and any new code is uniquely numbered and tested.
- [ ] Applicable — its driver/server/user classification is linked, or maintainer help is requested because the author cannot access the classification.

The public async API supports polling a statement from another connection, while statement-result responses do not identify the originating session. The client therefore updates session state only from responses attributable to the originating session.


src/main/java/com/databricks/jdbc/api/impl/DatabricksSession.java
```diff
@@ -43,6 +45,7 @@ public class DatabricksSession implements IDatabricksSession {
   private final IDatabricksComputeResource computeResource;
   private boolean isSessionOpen;
   private ImmutableSessionInfo sessionInfo;
+  private final AtomicReference<Long> sessionVersion = new AtomicReference<>();
```
vuanhphung: should we use Optional to be null-safe?

vuanhphung: nit: sessionVersionId seems more appropriate

src/main/java/com/databricks/jdbc/api/impl/DatabricksSession.java
```diff
@@ -111,6 +114,32 @@ public ImmutableSessionInfo getSessionInfo() {
...
+  public SessionVersion getSessionVersion() {
+    Long versionId = sessionVersion.get();
+    return versionId == null ? null : new SessionVersion().setVersionId(versionId);
+  }
+
+  @Override
+  public void updateSessionVersion(@Nullable SessionVersion newSessionVersion) {
+    if (newSessionVersion == null || newSessionVersion.getVersionId() == null) {
+      return;
+    }
```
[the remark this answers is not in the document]
vuanhphung: agree with this, AtomicReference seems redundant if you are already using synchronized. We can get away with volatile nullable field instead of AtomicReference

src/main/java/com/databricks/jdbc/dbclient/impl/sqlexec/DatabricksSdkClient.java
```diff
@@ -292,9 +302,11 @@ public DatabricksResultSet executeStatement(
       }
       String getStatusPath = String.format(STATEMENT_PATH_WITH_ID, statementId);
       try {
-        Request req = new Request(Request.GET, getStatusPath, apiClient.serialize(request));
```
vuanhphung, line 295: what's going on here before? is it sending ExecuteStatement to poll? seems weird

src/main/java/com/databricks/jdbc/model/core/SessionExecutionMode.java
```diff
@@ -0,0 +1,7 @@
+package com.databricks.jdbc.model.core;
+
+public enum SessionExecutionMode {
+  SESSION_EXECUTION_MODE_UNSPECIFIED,
```
vuanhphung: also do we need all of this if we only need FAST?

vuanhphung (approved):

PR #1683, opened 2026-09-10
Auto-recover Reyden Thrift rejection to SEA (KP001)

rahuls-db:
## What

Transparently auto-recover the Reyden onboarding case for JDBC: when a **default-Thrift**
connection to a Real-Time (Reyden) SQL warehouse is rejected by the gateway with SQLSTATE
`KP001` ("Lakehouse/RT is not supported for Thrift protocol"), the driver re-opens the session
on its SEA path (`DatabricksSdkClient`) and remembers the warehouse so later connections skip
Thrift — no customer configuration change. This mirrors the same feature already in the Python
(#948), Go (#479), and Node (#523) drivers; design follows ADBC databricks#670.

## Changes

- **`ReydenWarehouseCache`** (new): process-wide, thread-safe cache keyed by
  `host_lowercased|warehouse_id`, ~6h TTL, lazy eviction on read plus an opportunistic sweep on
  write.
- **`DatabricksSession.open`**:
  - **Pre-check** — on the default Thrift path, if the warehouse is already known-Reyden, open
    SEA directly and skip the Thrift round-trip.
  - **Reactive catch** — a `KP001` on the Thrift `OpenSession` marks the cache and (default path
    only) retries once on SEA. On a double failure both error chains are preserved (SEA error as
    cause, original Thrift `KP001` as a suppressed exception).
- Detection (`getSQLState() == "KP001"`) is scoped to the `OpenSession` `createSession` call
  only — there is no shared status checker that maps `KP001` elsewhere.

## Guardrail

Recovery engages only on the **default** Thrift path (`isDefaultThriftPath()` = `UseThriftClient`
unset). An explicit `UseThriftClient=1` is honored: `KP001` is surfaced to the caller, never
auto-switched to SEA.

### Known edge case (documented, not fixed here)

`isDefaultThriftPath()` gates on `UseThriftClient` alone. A user who sets a Thrift-forcing
metadata param (`UseQueryForMetadata=0` or `TreatMetadataCatalogNameAsPattern=1`) on a Reyden
warehouse will therefore be auto-recovered to SEA, silently dropping the native-Thrift-metadata
behavior. This is judged acceptable — the connection would otherwise hard-fail with `KP001`, and
a Reyden warehouse cannot serve Thrift metadata at all — but flagging it for reviewer input.

## Testing

- **Unit** (`ReydenThriftAutoRecoveryTest`, 8 tests): reactive fallback success, cache pre-check,
  explicit-Thrift guardrail (2 cases incl. cached-Reyden), double-failure error chaining,
  non-`KP001` propagation, and cache host/warehouse isolation. `DatabricksSessionTest` (21)
  passes unchanged.
- **End-to-end** against a prod Reyden warehouse:
  - Forced Thrift (`UseThriftClient=1`) returns the real `TStatus(ERROR_STATUS, sqlState:KP001)`
    and the guardrail surfaces it (no fallback).
  - Reactive path: connect #1 recovers `KP001` → SEA (10 rows, cache marked); connect #2
    pre-checks (skips Thrift), opens SEA directly (10 rows).

## Notes

- No new `DatabricksDriverErrorCode` is introduced; the double-failure exception reuses the
  existing `CONNECTION_ERROR` code, so no telemetry-taxonomy change is required.
- `NEXT_CHANGELOG.md` updated under `### Added`.

### DIFF
--- a/src/main/java/com/databricks/jdbc/api/impl/DatabricksSession.java
+++ b/src/main/java/com/databricks/jdbc/api/impl/DatabricksSession.java
@@ -143,10 +143,11 @@ public class DatabricksSession implements IDatabricksSession {
       String warehouseId = getWarehouseId();
       if (warehouseId != null
           && ReydenWarehouseCache.getInstance()
-              .isReydenWarehouse(connectionContext.getHost().toLowerCase(), warehouseId)) {
+              .isReydenWarehouse(connectionContext.getHost(), warehouseId)) {
         LOGGER.debug(
             "Warehouse {} is known to be Reyden. Switching to SEA to skip Thrift rejection.",
             warehouseId);
+        warnIfOverridingThriftMetadataPreference(warehouseId);
         connectionContext.setClientType(DatabricksClientType.SEA);
       }
     }
@@ -198,7 +199,7 @@ public class DatabricksSession implements IDatabricksSession {
             String warehouseId = getWarehouseId();
             if (warehouseId != null) {
               ReydenWarehouseCache.getInstance()
-                  .markReydenWarehouse(connectionContext.getHost().toLowerCase(), warehouseId);
+                  .markReydenWarehouse(connectionContext.getHost(), warehouseId);
             }
             // Only auto-recover on the default Thrift path; an explicit use_thrift_client=1 is
             // honored. Recovery is a single SEA attempt (no loop).
@@ -207,6 +208,7 @@ public class DatabricksSession implements IDatabricksSession {
                   "Thrift OpenSession rejected with SQLSTATE KP001 (not supported for Thrift protocol). "
                       + "Attempting transparent fallback to SEA. warehouse_id={}",
                   warehouseId);
+              warnIfOverridingThriftMetadataPreference(warehouseId);
               try {
                 this.connectionContext.setClientType(DatabricksClientType.SEA);
                 this.databricksClient =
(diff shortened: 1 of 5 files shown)


src/main/java/com/databricks/jdbc/api/impl/DatabricksSession.java
```diff
@@ -175,45 +193,94 @@ public void open() throws SQLException {
...
-              throw new DatabricksSQLException(
-                  String.format(
-                      "SEA session creation failed with HTTP %d rate limit, "
-                          + "and Thrift fallback also failed: %s",
-                      SeaCircuitBreakerManager.HTTP_TOO_MANY_REQUESTS,
-                      fallbackException.getMessage()),
-                  fallbackException,
-                  DatabricksDriverErrorCode.CONNECTION_ERROR);
+          } else if (e instanceof DatabricksRateLimitException) {
+            // Handle 429 rate limit error during SEA session creation
```
vuanhphung, line 242: side question: do we intend to keep this code around after SEA limit has been raised?

vuanhphung (approved):

src/main/java/com/databricks/jdbc/api/impl/DatabricksSession.java
rahuls-db: i guess we can get rid of it at a later point once server side is stable

PR #1691, opened 2026-09-23
fix: return HYT00 for query timeouts

cathleeny:
## Description

Return SQLSTATE `HYT00` when the client-side query timeout expires, matching the Databricks reference JDBC driver.

The SQLSTATE is set only at the client deadline enforcement points: the shared polling timeout handler used by SEA and Thrift, and the outer statement timeout. The default `DatabricksTimeoutException` mapping is left unchanged so this fix does not globally relabel server-originated or other timeout errors.

## Testing

- Focused statement, SEA, Thrift, and shared timeout tests: 247 passed, 0 failed
- Full `jdbc-core` unit suite before the scoping refinement: 3,571 passed, 0 failed
- Spotless passed under JDK 11


vikrantpuppala:
Can you add some context on why this change is needed?

cathleeny:
Closing since this is not needed

